Issue #1, opened 2018-06-20
Website content


jnewbery:
Thanks to @harding we now have a basic website that tells the world who we are and what we're doing.

We'll want to put more content on there as we start delivering our objective. This list is taken from @moneyball's comms draft. It can be updates and items added/removed.

- [x] Branding
    - [x] Name
    - [x] Logo
    - [x] Color scheme
    - [x] Tagline/brief description
- [ ] About
    - [ ] About us
        - [x] Mission
        - [ ] No business model, profit motive, etc.
    - [ ] Team
        - [ ] Give each of us a voice as to why we’re involved
        - [x] Sponsors/supporters such as Wences, JP, AB, AM
    - [ ] Member companies
    - [x] Contact us
- [ ] Content
    - [x] Quotes / support from Wences, JP, AB, AM
    - [ ] Blog
    - [ ] Cookbooks
    - [ ] Dashboard
    - [ ] Research
    - [x] Newsletters
    - [ ] Workshops

jnewbery:
> "No business model, profit motive, etc." What language can we use to speak to this? It is a common question about us so would be good to proactively address it.

Addressed in #4

jnewbery:
> "Sponsors/supporters such as Wences, JP, AB, AM" should we ask them if they'd like to be listed? maybe excluded AB as he's not providing financial support?

I think we should include WC, JP and Chaincode since they all provided funding/resource (assuming of course that they want to be on the website). We should also include all member companies who have signed up and contributed.

jnewbery:
> Workshops. can add details of the SF workshop. any reason not to? if so i can do this.

Yes, please go ahead!

jnewbery:
> "Quotes / support from Wences, JP, AB, AM" i think this would give our web site a lot more credibility. should i reach out to these 4 asking for a quote?

Yes please! I'd prefer if they came up with the quotes themselves, but up to you how you handle this!

jnewbery:
Emailed our sponsors for quotes and permission to put their logos on the website.

moneyball:
Can probably close this out with future work stated as separate issues? @jnewbery

jnewbery:
Agree

PR #4, opened 2018-06-26
Clarify business model

jnewbery:
Addresses the following point from #1:

> "No business model, profit motive, etc." What language can we use to speak to this? It is a common question about us so would be good to proactively address it.


jnewbery:
ping @moneyball

moneyball:
ACK. Looks ready to merge to me!

PR #10, opened 2018-06-28
Use a two-column list for sponsors (when screen space is wide enough)

harding:
Adds a general two-column-list I've used before (e.g. Bitcoin.org) and then customizes it to this particular application.  Also puts Pfeffer's quote in quotations for reasons described in the YAML file (feel free to drop that commit if you disagree).

Note the unused `br.clear` entry in the CSS can be useful if a list with an odd number of entries is interacting with other floating content.  That's not the case now, so it's not used, but it's implemented against that future situation.

I would suggest adding some content for Casares, as his entry looks lonely without it.


jnewbery:
This is great. Thanks @harding !

PR #18, opened 2018-07-11
Minor layout and graphics tweaks

jamesob:
- Add a horizontal logo
- Reposition a few elements on the home page

# Before 
selection_061

# After
selection_060


assets/css/main.scss
```diff
@@ -54,3 +54,23 @@ p.sponsor-quote {
...
+
+a.site-title {
+    width: 30%;
+}
+
+.home .content {
+    margin-top: 2em;
+
+    > .logo {
+        width: 30%;
```
harding: If you put `min-width: 10em;` below the existing `width` line, the graphic will be shown a bit larger on narrow mobiles but will be the same size on the widest displays.

_layouts/page.html
```diff
@@ -0,0 +1,18 @@
+---
+# Forked from the default minima layout to include our main.css
```
harding: It already did that on my system, were you having problems?  I don't see any layout-affecting differences when diffing a copy of the `_site` dir generate with this file to one generated without it, nor do I on a visual inspection of the rendered site.

OTOH, even if creating this file now is unnecessary, it's likely something that we'll end up customizing later anyway, so I'm not opposed to this.

index.md
```diff
@@ -4,8 +4,10 @@
 
 layout: home
 ---
-{:style="text-align: center"}
-![OpTech Logo](/img/logos/optech.png)
+
+<div class="logo">
```
harding: I don't have any problem with using HTML here (or anywhere, really), but FYI you can apply a CSS class to a block element in Kramdown like this:

    {:.logo}
    OpTech Logo

That will create:

    <p class="logo">OpTech Logo</p>

Other things you can do:
- `{#custom-id}`
- `{:custom-tag: "custom-value}`
- Plus other fancier stuff, see the [manual](https://kramdown.gettalong.org/syntax.html#block-ials)

assets/css/main.scss
```diff
@@ -54,3 +54,23 @@ p.sponsor-quote {
...
+        float: right;
+        display: inline-block;
+        margin-left: 2em;
+        margin-bottom: 2em;
+    }
+
+    div#sponsors {
```
harding: If this vertical whitespace looks good over the top of the Sponsors `<h3>`, you might want to consider applying it to every `<h3>` on the website, both for consistency and for the avoidance of having too many ad hoc rules that may making figuring out style things later harder.

_layouts/page.html
harding: Oh, I see, the CSS file was moved from its default location in d091a44f7b12b9ab19668112333c193d84af5a4c and only the homepage layout was updated to reflect that, so without this file, the newsletters don't get style changes.

This is fine.  An alternative is to move the css file back to its original location in /assets/ and drop the special code from the home layout.  I don't think it matters long term if this site is continued to be customized.

jnewbery:
Requires rebase (and addressing harding's comments).

assets/css/main.scss
```diff
@@ -54,3 +54,23 @@ p.sponsor-quote {
         max-width: 100%;
         box-shadow: 0 0 0 1px rgba(0,0,0,0.2);
 }
+
+a.site-title {
+    width: 30%;
```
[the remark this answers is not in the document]
jamesob: Good catch, thanks!

assets/css/main.scss
jamesob: Done

_layouts/page.html
jamesob: Yup; I prefer to keep `_layout/page.html` bundled as it's convenient to have the code ready at hand for modification, as well as for easier discovery - had to google around a bit to figure out where the page template lived beforehand.

index.md
jamesob: Oh, neat! Fixed.

jamesob:
I've incorporated @harding's good feedback (thanks, Dave) and I've changed the inline logo to remove the text, as it felt like we were repeating "Bitcoin Optech" a bit too much graphically. I can revert if you guys don't like this.

selection_064

selection_065

harding:
0c258965f7d3795e1771f4aa52006afc0b336147 LGTM.

jnewbery:
> The new PNG also has a transparent background

Thanks! That was bugging me :slightly_smiling_face:

jnewbery:
0c258965f7d3795e1771f4aa52006afc0b336147 LGTM :rocket:

PR #19, opened 2018-07-11
Add email subscription form to / and /newsletters

jamesob:
Note that this is based on #18 which should be tested and merged first.

selection_062

### DIFF
--- a/newsletters.md
+++ b/newsletters.md
@@ -9,6 +9,8 @@ share: false
 version: 1
 ---
 
+{% include newsletter-signup.html %}
+
 <ul class="post-list">
   {%- for post in site.posts -%}
   {%- if post.type == 'newsletter' -%}
--- a/_includes/newsletter-signup.html
+++ b/_includes/newsletter-signup.html
@@ -1,0 +1,20 @@
+<!-- Begin MailChimp Signup Form -->
+<link href="//cdn-images.mailchimp.com/embedcode/slim-10_7.css" rel="stylesheet" type="text/css">
+<style type="text/css">
+	#mc_embed_signup{clear:left; font:14px Helvetica,Arial,sans-serif; }
+	/* Add your own MailChimp form style overrides in your site stylesheet or in this style block.
+	   We recommend moving this block and the preceding CSS link to the HEAD of your HTML file. */
+</style>
+<div id="mc_embed_signup">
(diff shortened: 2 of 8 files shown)


newsletters.md
```diff
@@ -27,3 +27,5 @@ version: 1
   {%- endif -%}
   {%- endfor -%}
 </ul>
+
+{% include newsletter-signup.html %}
```
harding: Suggest moving this to the top of the page, probably right after the end of the YAML header (line 11) so it displays  below the Newsletters page heading but above the list.  The reason is that the list of newsletters is going to get longer over time, so this box will move further down the page few people ever see it.

You may also wish to add it to each newsletter (in the same position right below the end of the YAML header), as those pages may receive a lot of incoming link traffic.  There are a few ways to do this, such as creating a layout special to newsletters, putting it in the page layout triggered by a `news_signup: true` option in the YAML header, or just putting the `{% include %}` in each newsletter---the later is the easiest for now, so I'd suggest that. 

Otherwise this LGTM.  I tested and it took me to the mailchimp page and from there back to the live site, but they didn't ask me for double opt-in, so maybe it's in some sort of test mode?

jnewbery:
requires rebase on rebased #18.

jnewbery:
rebase plz :pray:

jamesob:
Rebased and addressed @harding's good feedback. The form is now above the list of newsletters, and I've added a `newsletter` layout that contains a signup form at the bottom of the page.

harding:
Tested ACK 41dead2c4d1214d1fbf19a39e656deaa9d940c70, a rather unfortunate commit id, with the caveat that a test of signup doesn't do double opt-in, so I think the list is in test mode or Mailchimp doesn't follow best practices.

jnewbery:
> Tested ACK 41dead2, a rather unfortunate commit id

I think Satoshi is trying to communicate with us.

jnewbery:
@jamesob is out of office today, so I've gone ahead and pushed a new branch. Changes:

- squashed commits down to two
- changed the CSS slightly so the subscribe area doesn't have a white box surrounding it.
- changed the mailchimp list that the subscribe form adds the subscriber to. Dave - this should fix your comment about double opt-in. Feel free to retest.
- changed the wording from **Get newsletter** to **Subscribe to our newsletter**

harding:
Fully tested ACK 258bdbfe665e2f8ee97a26c2d107e88510c4e20a  Thanks!

jnewbery:
Thanks @jamesob for code and @harding for review 🌠

PR #20, opened 2018-07-12
Remove unused _includes/_newsletters.html

jnewbery:
`_includes/_newsletters.html` was not being used. It was vestigial from a different approach that I was attempting for the newsletters page. Remove it.


harding:
Untested ACK.  It's a fatal build error to Jekyll if any included files are missing, so as long as this builds, it's safe to remove.  (On that note, now that this repository is public, you may want to setup Travis CI.  The `.travis.yml` from the BitcoinCore.org repository should work, sans the IRC notifications section.)

PR #21, opened 2018-07-12
Founding sponsors on one row

jamesob:
selection_066

### DIFF
--- a/_data/associates.yml
+++ b/_data/associates.yml
@@ -17,7 +17,7 @@ associates:
     github: jamesob
     avatar: https://avatars1.githubusercontent.com/u/73197?s=400&v=4
     quote: >-
-      "If Optech succeeds, the Bitcoin ecosystem will have higher rates of technical coordination and understanding; businesses will gain technical insight from the opensource community, and opensource engineers will understand the challenges that industrial users of Bitcoin face. This may benefit everyone working with and relying upon Bitcoin."
+      "If Optech succeeds, the Bitcoin ecosystem will have higher rates of technical coordination and understanding; businesses will gain insight from the opensource community, and opensource engineers will understand the challenges that industrial users of Bitcoin face."
   -
     name: Steve Lee
     role: Project Manager, Optech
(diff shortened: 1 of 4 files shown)


jamesob:
Ah oops - this broke /about. Fixing.

jamesob:
Fixed the about page:

selection_068

Also
- trimmed @harding's title down to "Technical writer" (can revert if preferred)
- shortened my quote a bit.

harding:
Doesn't render well in Firefox and will probably break in Chrom(e|inium) if you play around with the responsive preview mode at various widths.  At a super-quick look at the diff, I'd guess the problem is the removal of the nth-child clear rule, without which the layout engine tries to fit the content wherever it can rather than ensuring that all the previous floating content has been filled before filling new content.  I can look into it more tomorrow if you'd like.

(As for changing my title, that's fine.  I have no strong preferences.  If we ever meet in person sometime, you should ask me about my amusing story related to job titles.)

jamesob:
Hm, odd - the screenshot was from Chromium. I'll be sure to test in Firefox, and may special-case `ul.associates` with an n'th clear. 

> If we ever meet in person

@jnewbery, @marcinja, and I would love to drive south sometime and grab a beer! Serious offer!

jamesob:
Strange; works for me in Firefox:

about bitcoin optech - mozilla firefox_069

harding:
Tested ACK 0f0c674896bae1709176c9b75ccca5ec4f7795e0

I rechecked after some sleep and discovered that somehow I hadn't checked out the latest branch head.  Pulling the latest, it works for me.  Forcing a min-height on the content seems slightly dangerous, as it'll be easy to make a change that later results in breaking something, e.g.:



But fixing that problem correctly probably means redoing this with CSS flexbox or something.  So this solution seems satisfactory for now since it does improve the ascetics IMO. 

> @jnewbery, @marcinja, and I would love to drive south sometime and grab a beer! Serious offer!

You guys are nuts, then. :-)  It's a four hour round trip from NYC to here even with your own car.  The sane thing to do is to wait until my list of things to do---and people to meet---in NYC grows to a length where *I* get unlazy enough to actually hop the bus to go there.

jnewbery:
Tested ACK 0f0c674896bae1709176c9b75ccca5ec4f7795e0. New layout is :kissing: :ok_hand:

PR #26, opened 2018-07-19
Minor edits

jnewbery:
Minor edits.


jnewbery:
Should be non-controversial, so I'm just going to push.

PR #31, opened 2018-07-20
Add member company logos to home page

jnewbery:


jnewbery:
ok, I've hacked together something that works. The html/css is ugly so I'd like to go back and fix it later, but it's good enough for our launch tomorrow.

PR #32, opened 2018-07-20
reduce whitespace with company logos

moneyball:
I think this looks cleaner. Thoughts?


harding:
Tested ACK 7214a37fe33d5dfd0839588c7f1ed8427570f3fd .  Also tested that it's compatible with #34 

Technically, I'm pretty sure that when `max-height:` and `min-height:` have the same value, we could just use `height:`, but that suboptimal definition isn't introduced here, so I suggest merging as-is and leaving fixing it until #35

Also, it may be worth moving the newsletter subscribe form to the bottom of the page to move the logos up further.  There's already a big "Newsletter" link in the top-right corner of the page and, if we want to be even more brazen, we can link "weekly newsletter" in opening paragraph 3 to that page too.

PR #30, opened 2018-07-20
Add member company logos to home page

jnewbery:
Adds member logos to the home page.

PR #34, opened 2018-07-20
Always start a new row of content for <h2>s

harding:
On Firefox, I was having the issues mentioned in https://github.com/bitcoinops/bitcoinops.github.io/pull/30#issuecomment-406467996 despite the fix introduced in #31, so here's a solution that works for me (tested in FF and Chromium with various screen widths).

### DIFF
--- a/_includes/sponsors.html
+++ b/_includes/sponsors.html
@@ -2,7 +2,7 @@
 
 <div id="sponsors">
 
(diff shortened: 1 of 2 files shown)


jnewbery:
Looks great. ACK 217007cba9c0c4d90b0a95ead016cf47f212c92e. Thanks @harding !

PR #38, opened 2018-07-20
Add date to launch announcement

### LINKED ISSUE #37: Announcement page doesn't mention date
Announcement page https://bitcoinops.org/en/announcing-bitcoin-optech starts with "Today we’re announcing our new project.." but the date of announcement (July 20) doesn't appear anywhere on the page. Ideally the page would have a byline with date so the text is sensible in the future.

jnewbery:
Fixes #37 

~I added the date in parens after _today_. That meant the first sentence had two parens, which felt a bit unwieldy to me. I've moved the _(Optech)_ down to the second paragraph.~

Added a `post` layout and changed the launch post to use it.

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -1,0 +1,21 @@
+---
+layout: default
+---
+<link rel="stylesheet" href="/assets/css/main.css">
+
+<article class="post">
+
+  <header class="post-header">
+    <h1 class="post-title">{{ page.title | escape }}</h1>
+    <p class="post-meta">
+      <time class="dt-published" datetime="{{ page.date | date_to_xmlschema }}">
+        {%- assign date_format = site.minima.date_format | default: "%b %-d, %Y" -%}
+        {{ page.date | date: date_format }}
+      </time>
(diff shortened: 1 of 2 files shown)


harding:
I think it's fine, but it might be better to remove the relative date rather than add a secondary absolute date.  E.g., start with: "We're pleased to announce our new..."

Long-term, we can put the date in the footer [BitcoinCore.org-style](https://bitcoincore.org/en/2018/06/15/release-0.16.1/) or just under the title like in the [default](https://jekyll.github.io/minima/2016/05/20/welcome-to-jekyll.html).

jamesob:
Tested ACK

Side-note: looks like the John Pfeffer img is broken.

jnewbery:
> Side-note: looks like the John Pfeffer img is broken.

Fixed in master (#42)

jamesob:
Ah nevermind; my master was out of date. Sponsor images work.

jamesob:
Thanks for fixing this. Always like to see dates next to titles. :)

PR #44, opened 2018-07-25
Style changes

jnewbery:
- Centers titles and date
- adds date to newsletter posts
- adds space between the horizontal bar and text

blog post before:

image

blog post after:

image

newsletter before:

image

newsletter after:

image

### DIFF
--- a/_layouts/newsletter.html
+++ b/_layouts/newsletter.html
@@ -1,5 +1,21 @@
 ---
-layout: page
+layout: default
 --- 
-{{ content }}
+<link rel="stylesheet" href="/assets/css/main.css">
+<link rel="stylesheet" href="/assets/css/post.css">
+<article class="newsletter">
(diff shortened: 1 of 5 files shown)


jnewbery:
@jamesob has already visually reviewed these changes

harding:
utACK.

This doesn't add the bar and it does make it look better, so technically this paragraph is out of scope for this PR, but I'll note that I don't understand why the bar is there and think the page would look better without it (but it's not a big deal for me).

jnewbery:
> I don't understand why the bar is there and think the page would look better without it 

Agree - I've removed the bar (I think that might have been added by a copy-paste error from the front-matter :flushed: )

I also noticed that this branch centered the `h1`s in the about, newsletter and workshop pages, which I don't think we want. I've updated the commit to add a new post.css file.

PR #50, opened 2018-08-06
Update .gitignore

jnewbery:
Mostly taken from https://github.com/achow101/btcinformation.org/blob/master/.gitignore


harding:
I added a number of those things to the original Bitcoin site file and I know they don't apply to this site, but I don't really see the harm in including them here, so utACK.

Issue #33, opened 2018-07-20
reduce vertical whitespace between subscribe button and "Member Companies"


moneyball:
I'm hoping to have some of the company logos be above the fold on a 13" display in order to make the landing page a bit more attention-getting and catchy. There seems to be quite a bit of unnecessary vertical whitespace between the subscribe button and the Member Companies header. I took a quick look at the HTML/CSS and didn't see an obvious way to reduce it.

@harding do you know how?

jnewbery:
Fixed by #56

PR #56, opened 2018-08-08
style: reduce margin above h2 elements

jnewbery:


jnewbery:
@moneyball - do you think this is an improvement and resolves #33 for you?

moneyball:
It definitely improves the spacing so for that I tACK.

Separately though, I note that "Posts" font is larger than "Member Companies" and "Founding Sponsors" font, even though I checked and all are H2. I even modified the text "Posts" to "Member Companies" just to make sure my eyes weren't playing tricks on me but it is definitely bigger. Did you want to fix this in this PR or a separate one? I'd fix it but don't see how to.

jnewbery:
The *Posts* header is using the standard minima style:

```
.post-list-heading {
  @include relative-font-size(1.75);
}
```

I can remove that class from the Posts header to get it the same size.

jnewbery:
@moneyball - please test with the latest commit and let me know what you think.

PR #60, opened 2018-08-27
Add bitrefill as a member

jnewbery:
bitrefill is a member company. Add their logo to the website.

I have permission from Sergej to do this.

PR #61, opened 2018-08-28
Add bitrefill as a member

jnewbery:
Add bitrefill as a member company


jnewbery:
I accidentally opened #60 as a PR against the wrong branch.

PR #69, opened 2018-10-01
Newsletters: add #15 (2018-10-02)

harding:
Super slow news week, so a short newsletter.  Some notes:

- I was able to find a way to connect C-Lightning commits to PR numbers, so we now include those (see first two commits on this PR)
- I deliberately didn't link to the 0.17 release notes to avoid over-promoting the new release before it's been officially announced.  If the binaries are uploaded before the newsletter is released, it might be worth revising that action item accordingly.
- The LND RPCs are actually written in their code in UpperCamelCase, so I'm just following their convention here (I mention this because I've tried, and failed, in the past to use upper camel case for bitcoind things, but this time I have upstream on myside.  :-)  From the examples I've read, it seems they can be called all-lowercase or case insensitively.

### DIFF
--- a/_contrib/clc2b
+++ b/_contrib/clc2b
@@ -1,0 +1,11 @@
+#!/bin/bash -eu
+
+## C-Lightning Commit to Browser: opens a browser on a commit so that
+## you can find out its PR number
+
+if [ $# -lt 1 ]
+then
+  echo "Usage: $0 <commitid>"
+fi
+
+x-www-browser https://github.com/ElementsProject/lightning/commit/"$1"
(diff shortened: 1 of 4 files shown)


moneyball:
I gave the newsletter a quick review now and it looks good. Tomorrow morning I will test it locally, push it to publish to the web site, and send out the newsletter.

moneyball:
tACK. I plan to merge this PR now.

jnewbery:
Post-merge ACK 4ae097b573f61b3176799e2fc230964d3ab52465. Thanks @moneyball and @harding for handling this while I was away!

PR #79, opened 2018-10-23
Newsletter #18: fix broken link

harding:
Oops, as part of the review I fixed a PR number in one place and not the other, causing one of the links not to render.  This fixes it.

### DIFF
--- a/_posts/en/newsletters/2018-10-23-newsletter.md
+++ b/_posts/en/newsletters/2018-10-23-newsletter.md
@@ -187,7 +187,7 @@ commits].*
   default and which permits passing arbitrary data to that function.
 
 {% include references.md %}
-{% include linkers/issues.md issues="14291,14424,1978,2062,2063,2051,2033,354,14415,14531" %}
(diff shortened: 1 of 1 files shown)


markben93:
hello my name is mark not a coder yet . 
Id like to know if there is talks in the BTC crew too Fix the scaling problem in a similar way of DASH and Digibyte . With intensives that promotes the the crypto in question with multiple algorithms DGB  ? A promotion and Usability Focus with Dash, is That in the works and last thoughts on liquad and what does it do ?   
Last Wallets might need a structure that is visually appealing . is there a way too store multiple wallets on a small logo that when putting your mouse near it you will see a small repusentation of that wallet or wallets .once clicked it will auto copy it too the place you was too get coins from? 
p.s sorry if the idea doesn't make sense .

jnewbery:
Oops. Should have spotted this in my final review. Thanks for fixing.

ACK 783d26ebe9c3cf4feb263149e13233b81302ab04

PR #84, opened 2018-11-08
Makefile: test for broken reference-style links

harding:
This adds a test that should catch the problem that lead to #79.  This is the same test I've [used](https://github.com/bitcoin-core/bitcoincore.org/blob/master/Makefile#L21) in the BitcoinCore.org repository (and Bitcoin.org before it).

This PR also fixes an incidence of a broken ref-style link, although it only occurred in the HTML metadata for an article.  The excerpt used to fix that bug is the same as appears on the site [here](https://bitcoinops.org/en/newsletters/2018/07/24/#coming-attractions).  This should not result in any user-visible changes to the site (and I don't see any in my quick testing).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,4 +9,7 @@ build:
 	bundle exec jekyll build --future --drafts --unpublished
 
 test:
+	## Check for broken Markdown reference-style links that are displayed in text unchanged, e.g. [broken][broken link]
+	! find _site/ -name '*.html' | xargs grep ']\[' | grep -v skip-test | grep .
+	## Check for broken links
 	bundle exec htmlproofer --check-html --disable-external --url-ignore '/^\/bin/.*/' ./_site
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -9,4 +9,7 @@ build:
 	bundle exec jekyll build --future --drafts --unpublished
 
 test:
+	## Check for broken Markdown reference-style links that are displayed in text unchanged, e.g. [broken][broken link]
+	! find _site/ -name '*.html' | xargs grep ']\[' | grep -v skip-test | grep .
```
jnewbery: Is the `grep -v skip-test` necessary?

harding: Not yet here, but it's been used to solve issues in other repositories:

```text
user@devbco:~/bitcoincore.org$ git grep skip-test
Makefile:       ! find _site/ -name '*.html' | xargs grep ']\[' | grep -v skip-test | grep .
_releases/0.14.0.md:`"proxy"`      | `getnetworkinfo()["networks"][0]["proxy"]` <!--skip-test-->
_releases/0.14.0.md:array index or a non-quoted string (i.e. `listunspent()[0][txid]`). <!--skip-test--> Both
```

jnewbery: Makes sense!

PR #87, opened 2018-11-19
CSS: use locally-hosted copy of mailchimp signup CSS

harding:
I lost connectivity for a bit yesterday while proofreading the newsletter via a local preview, where I noticed we were pulling this CSS over the Internet.  I don't think that's necessary, so this just re-hosts the exact same CSS locally.  This not only makes working offline nicer but it also reduces the amount of information about casual visitors to the site that gets disclosed to Mailchimp.


practicalswift:
Concept ACK

jnewbery:
Tested ACK bdd8075a38b667d5e86ad69559eee62402b0b1c3. Confirmed that the added CSS file is the same as http://cdn-images.mailchimp.com/embedcode/slim-10_7.css.

Issue #25, opened 2018-07-19
Add member quotes to web site


moneyball:
“We're excited to work with Optech on the effort to scale and improve Bitcoin. By collaborating with leading engineers in this space, we'll be able to achieve more than we could have by tackling these problems alone." - Brock Miller, lead Bitcoin Engineer at Coinbase

“We’re proud to be participating in Optech to help further contribute to the crypto community. At Square we continue to explore ways cryptocurrency can expand financial access and we’re excited to help foster a collaborative ecosystem for the benefit of all.” -- Mike Brock, Strategic Development Lead at Square.

moneyball:
"It's great too have a forum to trade ideas about tackling engineering challenges Bitcoin businesses face."

"We appreciate the efforts of Bitcoin Optech to foster communication between Bitcoin companies and the open source community."

- Mark Erhardt, Engineer at BitGo

harding:
@Xekyo is it ok to use that two part quote as a single paragraph in the opposite order?  Also I'd like to add a couple stop words:

> We appreciate the efforts of Bitcoin Optech to foster communication between Bitcoin companies and the open source community.  It's great to have a forum to trade ideas about tackling **the** engineering challenges **that** Bitcoin businesses face.

Xekyo:
@harding: Thanks for asking. Yes, that's fine.

jnewbery:
@moneyball - do you plan to open a PR for this? Where do you think the member quotes should go?

moneyball:
At this point I think the cost outweighs the benefit. The home page is
already a bit noisy/messy and this would just add to that.

PR #93, opened 2018-12-10
Newsletters: add #25 (2018-12-11)

harding:
Our shortest newsletter since October, but still a reasonable length IMO.

Note, for the sighash news item, I sent an email to the bitcoin-dev list skeptically inquiring about whether that feature would actually be useful.  (The email appears to be in moderation as of this PR open.)  It's possible, although probably unlikely, that there will be agreement that it's not useful, at which point this is probably not noteworthy news and should be removed (even though it'll make the newsletter even shorter).

### DIFF
--- a/_posts/en/newsletters/2018-12-11-newsletter.md
+++ b/_posts/en/newsletters/2018-12-11-newsletter.md
@@ -1,0 +1,130 @@
+---
+title: "Bitcoin Optech Newsletter #25"
+permalink: /en/newsletters/2018/12/11/
+name: 2018-12-11-newsletter
+type: newsletter
+layout: newsletter
+lang: en
+---
+This week's newsletter suggests helping test a Bitcoin Core maintenance
+release candidate, provides a link to a modern block explorer whose code
+has been open sourced, and briefly describes a suggestion for signature
+hashes to optionally cover transaction size.  Notable code changes made
+in the past week to popular infrastructure projects are also described.
+
+## Action items
+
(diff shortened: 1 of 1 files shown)


_posts/en/newsletters/2018-12-11-newsletter.md
```diff
@@ -0,0 +1,130 @@
...
+This week's newsletter suggests helping test a Bitcoin Core maintenance
+release candidate, provides a link to a modern block explorer whose code
+has been open sourced, and briefly describes a suggestion for signature
+hashes to optionally cover transaction size.  Notable code changes made
+in the past week to popular infrastructure projects are also described.
+
+## Action items
+
+- **Help test Bitcoin Core 0.17.1RC1:** the first release candidate for
+  this [maintenance release][] has been tagged.  Testing by businesses
```
jnewbery: Include link to tagged release? https://github.com/bitcoin/bitcoin/releases/tag/v0.17.1rc1

moneyball:
utACK

harding:
ACK 0d182e9.  Thanks!  If binaries are uploaded to bitcoincore.org/bin by publication time, I suggest linking to that instead.

harding:
Updated to link to binaries directory as the RC has been uploaded.  Squashed.

jnewbery:
ACK e9ec9ff57a610a7a37462ad770c0daa14650cf04. Only change is linking to the binary instead of the tagged release on github.

PR #96, opened 2018-12-19
blog post: Add review of 2018/plan for 2019

jnewbery:
Incorporated feedback from @harding @moneyball and @bitschmidty . Thanks!

Could do with an image and a snappier title. It looks a bit :yawning_face: right now.

I suggest we release at the same time as our year-in-review newsletter (and add a link from the newsletter to the blog post).

PR #97, opened 2018-12-27
Newsletters: add #27

harding:
Based on top of #96 (squashed).

I've left in the commits based on emailed feedback in case anyone wants to check the edit trail, but these can be squashed at commit time.


moneyball:
utACK. Amazing.

jnewbery:
squashed commits down to two and merged. Thanks all!

PR #98, opened 2018-12-29
CSS: shrink tables and handle x overflows

harding:
This decreases the text size in the tables slightly and adds horizontal scrollbars to tables that are still too wide.  You can preview the changes using an actual mobile devices at http://dg3.dtrt.org/en/newsletters/2018/12/28/

I guess in the future, though, we'll just have to avoid tables.  :-(

### DIFF
--- a/_posts/en/newsletters/2018-12-28-newsletter.md
+++ b/_posts/en/newsletters/2018-12-28-newsletter.md
@@ -33,11 +33,15 @@ and signature.  This not only reduces overhead and costs, it also
 increases privacy by making basic multisig transactions look identical
 to single-sig transactions.
 
+<div markdown="1" class="xoverflow shrink80">
+
 | | Receive script | Spend data |
 |-|-|-|
 | **Single-sig, current Script (P2PK)** | `<pubkey> OP_CHECKSIG` | `<signature>` |
 | **Bare multisig, current Script** | `2 <pubkey> <pubkey> <pubkey> 3 OP_CHECKMULTISIG` | `OP_0 <signature> <signature>` |
 | **Multisig, muSig[^fn-opcodes]** | `<pubkey> OP_CHECKSIG` | `<signature>` |
+
+</div>
 
 Building on the idea that muSig, or something like it, could become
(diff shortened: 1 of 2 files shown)


moneyball:
The tables added a lot of value. In which situations will users have
problems after this PR, and what proportion of users will have problems vs.
those benefitting from the visual benefit of the tables?

jnewbery:
> I guess in the future, though, we'll just have to avoid tables. :-(

Why do you say this? As far as I'm aware, the newsletter on our site works fine on all devices. (I've tested using Chrome's developer tools and with mobile and tablet devices).

Mailchimp has issues with making images embedded in text blocks responsive to different screen widths, which caused problems with the year-in-review newsletter.

The mailchimp emails also had a 'View this email in your browser' link which linked to mailchimp's campaign archive page (eg https://us18.campaign-archive.com/?u=70c3f85e5d13ffec674f30af8&id=b6aeae0273). I've now changed that link on the template so that it always links to https://bitcoinops.org/en/newsletters.

So I think we're in an ok place now:

- The website, which is the definitive edition and archive of the newsletter works fine on all devices.
- I can work around most issues in mailchimp, but there are occasional problems (eg the graph images in the text box for the year-in-review issue).
- the mailchimp version links to our website now.

Are we aware of any other issues?

harding:
Oh, I was under the impression that the complaint we received was someone looking at the website version of the article on their iPhone and having problems with the table.  However...

> eg https://us18.campaign-archive.com/?u=70c3f85e5d13ffec674f30af8&id=b6aeae0273

That does indeed look terrible on a small mobile screen.  The paragraph width is greater than the screen width, so you need to scroll horizontally to read each sentence.  I can easily believe that this is what the reader was complaining about, and it wasn't a fundamental problem with tables on our website.

I still think that the changes made in this PR make the text a bit more readable.  However, if you want, I can reduce the changes to just the `xoverflow` class, which I think is a strict improvement with no downside except so extra formatting complexity.

jnewbery:
Looks good. Thanks Dave!

PR #28, opened 2018-07-19
Xapo consolidation

jnewbery:
This is a contribution from @ajtowns at Xapo.

PR #58, opened 2018-08-19
Newsletters: add #9 (2018-08-21)

harding:
- [x] Dashboard items @moneyball 

Note: I'll be AFK most of the day tomorrow (Monday), so please push changes without waiting to hear back from me.  I'll review later in the evening (EDT) when I get home.

PR #29, opened 2018-07-19
added call to action

moneyball:

PR #99, opened 2019-01-05
Newsletters: add #28 (2019-01-08)

harding:
Happy 2019.  :-)

### DIFF
--- a/_posts/en/newsletters/2019-01-08-newsletter.md
+++ b/_posts/en/newsletters/2019-01-08-newsletter.md
@@ -87,22 +87,6 @@ commits].*
   their fee income.  See [Newsletter #20][] for a recent instance where
   this may have happened to a large mining pool.
 
-- [Bitcoin Core #14336][] changes the system call (syscall) Bitcoin Core
-  uses to handle network connections and other resources on platforms
-  that support an alternative syscall (only Linux for now).  On a
-  Debian GNU/Linux x86_64 system tested by Optech, this increased the
-  maximum allowed number of connections from 865 to 9,999.  Please note,
-  developers strongly advise against increasing the maximum number of
-  connections unless you have a special need for more connections.  For
-  each peer you have after your first peer, you increase the amount of
-  bandwidth overhead used by your node and by all of your peers.  The
-  default minimum of 8 and maximum of 125 peers were chosen to provide
-  enough connections to ensure robust relaying but also to not create so
-  much bandwidth overhead that the network would be excessively
-  wasteful.  As relay improvements are implemented (see the News section
-  of [Newsletter #26][]), Bitcoin Core developers may adjust these
-  defaults.
-
 - [C-Lightning #2172][] allows `lightningd` to be shutdown normally even
   if it's operating as the primary process (PID 1), which can be useful
   in Docker containers.  This is, for example, how the open source
@@ -132,7 +116,7 @@ commits].*
   future states are planned.
 
 {% include references.md %}
-{% include linkers/issues.md issues="14565,14811,14336,2172,2188,2374,2354" %}
+{% include linkers/issues.md issues="14565,14811,2172,2188,2374,2354" %}
 {% include linkers/github-log.md
   refname="core commits"
   repo="bitcoin/bitcoin"
--- a/_posts/en/newsletters/2019-01-08-newsletter.md
+++ b/_posts/en/newsletters/2019-01-08-newsletter.md
@@ -1,0 +1,168 @@
+---
+title: 'Bitcoin Optech Newsletter #28'
+permalink: /en/newsletters/2019/01/08/
+name: 2019-01-08-newsletter
+type: newsletter
+layout: newsletter
+lang: en
+---
+This week's newsletter announces a new maintenance release of Bitcoin
+Core, describes continued discussion about new signature hashes, and
+links to a post about possible economic barriers to LN payments crossing
+different currencies.  Descriptions of notable code changes in popular
+Bitcoin infrastructure projects are also provided.
+
+## Action items
+
+- **Upgrade to Bitcoin Core 0.17.1:** this new [minor][maintenance]
+  version released December 25th restores some previously-deprecated
+  functionality to the `listtransactions` RPC and includes bug fixes and
+  other improvements.  Consider reading the [release notes][0.17.1
+  notes] and [upgrading][0.17.1 bin].
+
+## News
+
+- **Continued sighash discussion:** as mentioned in the News section of
+  [Newsletter #25][], developers on the Bitcoin-Dev mailing list
+  discussed how signature hashes could be modified to give transactions
+  access to new capabilities.  Sighashes give spenders the ability to
+  allow their transactions to be modified in specified ways after they
+  are signed---for example, two people who open a payment channel
+  together can use a particular type of sighash to allow either one of
+  them to unilaterally attach additional transaction fees to a channel
+  close transaction.
+
+    The most recent discussion in this thread of almost 70 posts has
+    mostly involved edge cases related to new sighash flags,
+    particularly a BIP118-like `SIGHASH_NOINPUT_UNSAFE`.  As part of the
+    discussion, protocol developer Johnson Lau described an
+    [optimization for Eltoo-based payment channels][lau bip68].  Also
+    [discussed][rm codesep] is whether the `OP_CODESEPARATOR` opcode
+    should be disabled in a script update that supports MAST (e.g. via
+    Taproot).  That opcode is not in common use, but if you plan to use
+    it in future Script versions, you should comment on the thread.
+
+- **Cross-chain LN as an options contract:** pseudonymous LN
+  contributor ZmnSCPxj started a thread on the Lightning-Dev mailing
+  list pointing out that users could abuse payments that cross
+  currencies to create almost free [short-term options contracts][] by
+  delaying payment settlement.  A [previous thread][cjp risk] by Corné
+  Plooy in May 2018 described the same thing.
+  
+    For example, Mallory learns that Bob is willing to route payments
+    from Bitcoin to Litecoin, so she sends a payment from one of her
+    Bitcoin nodes through Bob to one of her Litecoin nodes.  If this
+    were a normal payment, she'd settle it immediately by releasing the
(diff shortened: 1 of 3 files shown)


_posts/en/newsletters/2019-01-08-newsletter.md
```diff
@@ -0,0 +1,168 @@
...
+    Taproot).  That opcode is not in common use, but if you plan to use
+    it in future Script versions, you should comment on the thread.
+
+- **Cross-chain LN as an options contract:** pseudonymous LN
+  contributor ZmnSCPxj started a thread on the Lightning-Dev mailing
+  list pointing out that users could abuse payments that cross
+  currencies to create almost free [short-term options contracts][] by
+  delaying payment settlement.  A [previous thread][cjp risk] by Corné
+  Plooy in May 2018 described the same thing.
+
```
practicalswift: Nit: Trailing spaces here :-)

_posts/en/newsletters/2019-01-08-newsletter.md
```diff
@@ -0,0 +1,168 @@
...
+This week's newsletter announces a new maintenance release of Bitcoin
+Core, describes continued discussion about new signature hashes, and
+links to a post about possible economic barriers to LN payments crossing
+different currencies.  Descriptions of notable code changes in popular
+Bitcoin infrastructure projects are also provided.
+
+## Action items
+
+- **Upgrade to Bitcoin Core 0.17.1:** this new [minor][maintenance]
+  version released December 25th restores some previously-deprecated
```
practicalswift: Nit: Drop "-" in "previously-deprecated"?

_posts/en/newsletters/2019-01-08-newsletter.md
```diff
@@ -0,0 +1,168 @@
...
+layout: newsletter
+lang: en
+---
+This week's newsletter announces a new maintenance release of Bitcoin
+Core, describes continued discussion about new signature hashes, and
+links to a post about possible economic barriers to LN payments crossing
+different currencies.  Descriptions of notable code changes in popular
+Bitcoin infrastructure projects are also provided.
+
+## Action items
```
practicalswift: Nit: I don't know if we're strict we are with the Markdown headers, but in the case we are: should we remove one `#` in the header level (from `##` to `#`) to make it so that the first header is always a top level header?

harding:
Some upstream developer decided to break the way we used Travis, so this PR now includes a commit with a temporary hack fixing it [based on these instructions](https://docs.travis-ci.com/user/languages/ruby/#bundler-20).  Ultimately, everyone who generates local previews will have to upgrade our ruby versions, which is something we probably want to coordinate.

_posts/en/newsletters/2019-01-08-newsletter.md
```diff
@@ -0,0 +1,168 @@
...
+  in Docker containers.
+
+- [C-Lightning #2188][] adds notification subscription handlers that can
+  be used by plugins, with initial support for notifications that the
+  node has connected to a new peer or disconnected from an existing
+  peer.  The [plugin documentation][cl plugin event] and [sample
+  plugin][cl helloworld.py] have been updated for these handlers.
+
+- [LND #2374][] increases the maximum size of the gRPC messages the
+  `lncli` tool will accept from 4 MB to 50 MB.  This fixes a problem
```
jnewbery: minor nit: I found the "will accept from..." tripped me up on my first parse - the *from* attached itself to *accept* instead of *increases*. You could consider parenthesizing *from 4 MB to 50 MB* (although I'll admit I'm often too liberal with parentheses)

_posts/en/newsletters/2019-01-08-newsletter.md
```diff
@@ -0,0 +1,168 @@
...
+  some nodes were encountering where the `describegraph` RPC was failing
+  due to the network having grown so large that messages exceeded this
+  limit.  Developers using gRPC directly will need to increase their
+  client-side maximum message size setting---descriptions of how to do
+  this have already been added as comments to the PR for python and nodejs.
+  Ultimately it's expected that the network will grow large enough to
+  exceed even this new maximum, so developers are planning to revamp the
+  relevant RPCs to handle this situation.
+
+- [LND #2354][] adds a new `invoicestate` field and deprecates the former
```
jnewbery: I think 'removes' is better than 'deprecates'. *Deprecates* suggests to me that the field is still there but will be removed in future, but in fact PR 2354 removes the field.

_posts/en/newsletters/2019-01-08-newsletter.md
```diff
@@ -0,0 +1,168 @@
...
+
+    The most recent discussion in this thread of almost 70 posts has
+    mostly involved edge cases related to new sighash flags,
+    particularly a BIP118-like `SIGHASH_NOINPUT_UNSAFE`.  As part of the
+    discussion, protocol developer Johnson Lau described an
+    [optimization for Eltoo-based payment channels][lau bip68].  Also
+    [discussed][rm codesep] is whether the `OP_CODESEPARATOR` opcode
+    should be disabled in a script update that supports MAST (e.g. via
+    Taproot).  That opcode is not in common use, but if you plan to use
+    it in future Script versions, you should comment on the thread.
```
moneyball: Should this be in the Action Items section as well?

moneyball:
aside from my one comment/question, utACK

_posts/en/newsletters/2019-01-08-newsletter.md
harding: I think it's ok to leave this a bit buried.  We do want people to comment sooner than later, but I think it's ok if nobody comments until there's been an actual formal proposal of a soft fork with a new Script version that doesn't include this opcode.  Besides, my guess is that nobody is using this opcode.

_posts/en/newsletters/2019-01-08-newsletter.md
```diff
@@ -0,0 +1,168 @@
...
+  the that import (but only if there were any problems).
+
+- [Bitcoin Core #14811][] updates the [getblocktemplate][rpc
+  getblocktemplate] RPC to require that the segwit flag be passed.  This
+  helps prevent miners from accidentally not using segwit, which reduces
+  their fee income.  See [Newsletter #20][] for a recent instance where
+  this may have happened to a large mining pool.
+
+- [Bitcoin Core #14336][] changes the system call (syscall) Bitcoin Core
+  uses to handle network connections and other resources on platforms
```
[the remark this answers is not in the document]
harding: That user only encountered those problems because they were running with more than the default maximum number of connections:

```
229 2018-07-31T19:58:47  <sipa> how many connwctions do you have?
230 2018-07-31T20:01:26  <ossifrage> sipa, 160
```

IIUC, the lack of file descriptors never caused problems with the database---the user-visible problem was that sometimes the user couldn't previously get to their configured maximum number of connections, as we described in Newsletter 8:

> Bitcoin Core #13925: increases the maximum number of file descriptors Bitcoin Core’s internal database can use, which can allow more file descriptors to be used for network connections. If you’ve modified Bitcoin Core to accept more than 117 incoming connections, you may see an additional increase in the number of connections after upgrading past this merge. (Note: we don’t recommend increasing the default unless you have a special need.)

We also previously discussed some technical aspects of this issue: https://github.com/bitcoinops/bitcoinops.github.io/pull/58#issuecomment-414384299

In short, I think the description in the newsletter now describes the user-visible changes, which are what I think is most important to our readers, rather than going into the technical details (which are still available to anyone who clicks the link to the PR).

I don't think I'll be around to see your reply to this issue before publication time, so feel free to (1) modify the text as you think is appropriate or (2) drop the bullet point from this newsletter and we can discuss it and optionally put it in next week's newsletter.

_posts/en/newsletters/2019-01-08-newsletter.md
harding: Are you sure?  I don't run lnd, so I can't quickly check, but I wrote my description based on how I guessed it worked from the patch:

```diff
diff --git a/lnrpc/rpc.proto b/lnrpc/rpc.proto
index 8794c1b4..10e53179 100644
--- a/lnrpc/rpc.proto
+++ b/lnrpc/rpc.proto
@@ -1699,7 +1699,7 @@ message Invoice {
     int64 value = 5 [json_name = "value"];
 
     /// Whether this invoice has been fulfilled
-    bool settled = 6 [json_name = "settled"];
+    bool settled = 6 [json_name = "settled", deprecated = true];
```

I don't know enough about gRPC in general or lnd in particular to know whether this removed the field or just tags it as deprecated.  However, I would consider it poor form on their part to remove a field without deprecating it first.

I rewrote the opening of this bullet so that it's consistent with either of our interpretations, but feel free to edit as you think appropriate.

_posts/en/newsletters/2019-01-08-newsletter.md
harding: When using Jekyll, the `title` field in the YAML sets the the content of the rendered page's `<title>` and `<h1>` tags, so our top-level section are semantically correct for using `##` to generate `<h2>`s.

harding:
For the Travis workaround, I suggest we use it for this newsletter and the next one (`#29`).  During this week, I'll submit a PR with the upgrade to a new Ruby, instructions for upgrading, and add to the Makefile one of the lint checks @practicalswift is doing that I think is particularly important.  If the PR review goes well, we can all upgrade after we publish `#29`, giving us a whole week to work out any problems.

_posts/en/newsletters/2019-01-08-newsletter.md
jnewbery: You're right and I'm wrong. This is deprecated and not removed. The `settled` field is set from the `ContractTerm.State` field here: https://github.com/lightningnetwork/lnd/pull/2354/files#diff-01ada5680aaa18fc79cbf78b2d55a37bR3306

jnewbery: I've reverted your change back to '...and deprecates the former'

_posts/en/newsletters/2019-01-08-newsletter.md
practicalswift: Makes perfect sense! Will not report this going forward :-)

PR #100, opened 2019-01-09
Update Ruby version, remove Travis workaround, and add Markdown linter

harding:
This removes the need for the workaround in 71e9fad42cecd44cfc418a5802490cd809703a8d by updating Ruby.  I chose Ruby 2.5.1 because [that's the version](https://packages.ubuntu.com/bionic/ruby2.5) used for Ubuntu 18.04 LTS, so users of that system should, in theory, not even need to use rvm to get a Ruby to build site previews.  I use Debian, not Ubuntu, so I haven't tested that.

This also adds a lint checker suggested by @practicalswift with a single check for extraneous spaces at the end of a line.  I think this is an important rule because some Markdown parsers turn two spaces at the end of a line into a `<br>` tag, complicating review because whitespace at the end of a line is often invisible.  There are plenty of [other rules](https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md) we can add if we want, although I'd prefer to focus on preventing problems rather than enforcing a consistent style.

The short steps to upgrade are:

1. `cd` to your checkout
2. (Optional) Checkout current master, run `make all`, then run `cp -a _site _old_site`
2. Checkout this PR
3. Then,

```bash
rvm install 2.5.1
cd ; cd -
ruby --version
# Should say something like: ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
bundle install
make all
```

Optionally, you can finish by running `diff -ru _old_site _site` to see how this change affects the build.  Besides changing in-file datestamps, the only changes I see are rearragements in the order associative arrays are printed and the README.md file is no longer rendered on the site (as expected).

Note that if you checkout out a branch that uses the previous version of Ruby, you need to exit and re-enter the top-level repository directory to get RVM to select the current Ruby version.  You can go to your home directory and back to your last-used directory using: `cd ; cd -`

On my system, going back to an older version of Ruby causes one of HTMLproofer's dependencies to segfault.  This can be resolved by deleting your cached bundler gems (on Linux, try `mv ~/.gems /tmp`) and then re-running `bundle install`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-  - "2.2.2"
+  - "2.5.1"
 
 ## Allows use of docker-based containers, which are more available than
 ## the xen-based (or whatever) VMs, so builds tend to happen sooner
@@ -8,10 +8,6 @@ sudo: false
 
 ## Save bundler deps.  Requires sudo: false
 cache: bundler
-
-## Required if our Ruby version is 2.2 or below
-before_install:
-   - gem install bundler -v '< 2'
 
 ## Build all possible jekyll pages for use with link checking.
 script:
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 ## If you update the version here, also update it in .travis.yml and
 ## README.md. Then push your branch and make sure Travis supports that
 ## version.
-ruby '2.2.2'
+ruby '2.5.1'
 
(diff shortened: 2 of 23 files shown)


jnewbery:
Looks great. Tested ACK f3103ce20698a743dbe92b0fc56c4409511b633e.

As suggested here: https://github.com/bitcoinops/bitcoinops.github.io/pull/99#issuecomment-452112506, I'll merge this after next week's newsletter is merged.

jnewbery:
#101 is merged, so merging this too.

PR #101, opened 2019-01-14
Newsletters: add #29 (2019-01-15)

harding:
Slow news week, so a short newsletter.  Some notes:

- This week includes notable merges from Eclair, as suggested previously by Steve and John.  (Just one notable merge this week.)  I'm also monitoring the BIPs and lightning-rfc repositories, but neither has had any merges this year, so I don't mention them yet in the introduction.

- I've dropped links to the GitHub diffs for the different projects this week.  Maintaining that was mildly annoying for me and I suspect nobody was looking at them.

- The versions of Bitcore and Copay mentioned as being affected are my own research---Bitcore by looking at which git tags included the commits Maxwell identified as introducing the vulnerability and the fix; Copay by looking at which versions of its bower.json file depended on which versions of Bitcore.

### DIFF
--- a/_posts/en/newsletters/2019-01-15-newsletter.md
+++ b/_posts/en/newsletters/2019-01-15-newsletter.md
@@ -1,0 +1,109 @@
+---
+title: 'Bitcoin Optech Newsletter #29'
+permalink: /en/newsletters/2019/01/15/
+name: 2019-01-15-newsletter
+type: newsletter
+layout: newsletter
+lang: en
+---
+This week's newsletter announces a security upgrade for C-Lightning,
+describes a paper and additional research into wallets that accidentally
+revealed their private keys, and lists some notable code changes in
+popular Bitcoin infrastructure projects.
+
+## Action items
+
+- **Upgrade to C-Lightning 0.6.3:** this [release][cl 0.6.3] fixes a remote DoS
+  vulnerability that could be used to crash C-Lightning nodes and
+  potentially steal money.  See the *notable code changes* section below
+  for details.  This release also includes other less critical bug fixes
+  and new features.
+
+## News
+
+- **Weak signature nonces discovered:** a preprint [paper][weak nonces]
+  by researchers Joachim Breitner and Nadia Heninger describes how they
+  discovered hundreds of Bitcoin private keys by looking for signatures
+  generated using nonces with less than the expected entropy of 256
+  bits.  Independent [code archaeology][gmaxwell bitcore] by Gregory
+  Maxwell indicates that the main culprit was probably the BitPay
+  Bitcore software which introduced a bug around July 2014 and released
+  a fix about a month later.  (Note: BitPay Bitcore is unrelated to
+  Bitcoin Core.)  From there, the bug propagated to software such as
+  BitPay Copay that depended upon Bitcore.  About 97% of the faulty
+  signatures found in the paper are compatible with Maxwell's Copay
+  hypothesis, and the paper provides plausible explanations for most of
+  the remaining 3% of signatures, indicating that users of modern
+  wallets are probably safe provided they do not continue to use
+  addresses whose bitcoins they spent using earlier vulnerable programs.
+
+    If you ever used an affected version of Bitcore (0.1.28 to 0.1.35),
+    Copay (0.4.1 to 0.4.3), or other vulnerable software, you should
+    create a new wallet file, send all of your funds from the old wallet
+    file to an address in the new wallet, and discontinue use of the
+    previous wallet file.  When designing software that signs Bitcoin
+    transactions, you should prefer to use peer-reviewed implementations
+    that generate signature nonces deterministically, such as
+    [libsecp256k1][] which implements [RFC6979][].
+
+    The fast analysis method employed by the authors of the paper took
+    advantage of users who engaged in address reuse, but even keys for
+    addresses that have not been reused are vulnerable to attack if the
+    nonce generation is biased or too small.  This can be either through
+    using the same method for keys that were used multiple times (e.g.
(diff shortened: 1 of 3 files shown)


_posts/en/newsletters/2019-01-15-newsletter.md
```diff
@@ -0,0 +1,94 @@
...
+lang: en
+---
+This week's newsletter announces a security upgrade for C-Lightning,
+describes a paper and additional research into wallets that accidentally
+revealed their private keys, and lists some notable code changes in
+popular Bitcoin infrastructure projects.
+
+## Action items
+
+- **Upgrade to C-Lightning 0.6.3:** this release fixes a remote DoS
```
jnewbery: Consider adding a link to the github release page: https://github.com/ElementsProject/lightning/releases and the changelog: https://github.com/ElementsProject/lightning/blob/v0.6.3/CHANGELOG.md

_posts/en/newsletters/2019-01-15-newsletter.md
```diff
@@ -0,0 +1,94 @@
...
+    create a new wallet file, send all of your funds from the old wallet
+    file to an address in the new wallet, and discontinue use of the
+    previous wallet file.  When designing software that signs Bitcoin
+    transactions, you should prefer to use peer-reviewed implementations
+    that generate signature nonces determinstically, such as
+    [libsecp256k1][] which implements [RFC6979][].
+
+## Notable code changes
+
+*Notable code changes this week in Bitcoin Core, LND, C-lightning,
```
jnewbery: Consider making these links to the repos:

- https://github.com/bitcoin/bitcoin
- https://github.com/lightningnetwork/lnd/
- https://github.com/ElementsProject/lightning
- https://github.com/ACINQ/eclair
- https://github.com/bitcoin-core/secp256k1

_posts/en/newsletters/2019-01-15-newsletter.md
```diff
@@ -0,0 +1,94 @@
...
+  between honest miners who simply extend the block chain and dishonest
+  miners who create chain forks in an attempt to steal fees from honest
+  miners.  However, when using anti-fee-sniping, nodes that have been
+  offline for a while don't know what block is at the tip of the chain
+  and so they could create multiple transactions offline that would all
+  use the same very old nLockTime value, linking those transactions
+  together in block chain analysis.  This merge fixes the problem by
+  disabling the feature if a node is offline for too long.
+
+- [C-Lightning #2214][] fixes a DoS vulnerability where a peer could
```
jnewbery: I like all the information here, but it looks like the most important information ('upgrade is highly recommended') is buried in the middle of the text. You have highlighted the upgrade advice in the **actions** section above, but I wonder if we can also make it more visible here. Perhaps something like:

```
- [C-Lightning #2214][] fixes a remote crash bug which could lead
  to loss of funds. **All users are advised to upgrade to 0.6.3 to
  get a fix for this issue.**

  Before this fix, a peer could cause the c-lightning node to crash
  by trying to get it to accept a payment with a smaller timelock
  than the node allows. As with all lightning node remote-crash
  bugs, this could allow an attacker to steal funds from open channels.
  However, in order to do this, the attacker must risk their own money.
  This is hoped to be enough of a risk to discourage most attacks.
```

_posts/en/newsletters/2019-01-15-newsletter.md
```diff
@@ -0,0 +1,94 @@
...
+
+- **Upgrade to C-Lightning 0.6.3:** this release fixes a remote DoS
+  vulnerability that could be used to crash C-Lightning nodes and
+  potentially steal money.  See the *notable code changes* section below
+  for details.  This release also includes other less critical bug fixes
+  and new features.
+
+## News
+
+- **Weak signature nonces discovered:** a preprint [paper][weak nonces]
```
jnewbery: May be worth adding somewhere that not reusing addresses avoids this issue entirely.

harding: Eh, what if you use RBF?  Furthermore, the Bitcore weak nonces only contained 64 bits of entropy, and the paper has this to say about that:

> 64-bit nonces and single-signature keys. Our lattice attack only applies when at least two signatures with a small nonce are created using the same secret key. A single 64-bit nonce requires only 2^32 time to break using Shanks’s baby-step-giant-step algorithm [34] or the Pollard rho algorithm [28], feasible with only modest computation resources. Applying this attack to all of the 2^30 Bitcoin signatures was beyond our resources, but we investigated a random sample.

I'm unfamiliar with the methods they describe, but IIUC it seems plausible that an attacker could scan their mempool for signatures with <=64-bit nonces and recover the corresponding private keys even with only one signature.

That said, if I ever start a band, I'm naming it Shanks's Baby-Step-Giant-Step Algorithm.

jnewbery: Ah, you're right. Thanks for clarifying! Can I update my suggestion to "May be worth adding somewhere that not reusing addresses does not avoid this issue"? :slightly_smiling_face: 

Something like:

The method in the Breitner-Heninger paper was used to retrieve private keys where the address had been reused and the the key had been used to sign multiple transactions. However, even keys for addresses that have not been reused are vulnerable to attack if the nonce generation is biased or too small, either through using the same method if the key has been used to sign multiple times (eg for Replace-by-fee), or through simply brute-forcing the key using the baby-step giant-step or Pollard's Rho methods.

harding:
Pushed an update that I think should address all of @practicalswift's and @jnewbery's concerns.  Thanks!

jnewbery:
ACK 7e3bf2aa61ca89b3b7667eb7f8258573e5f30d9e. Thanks Dave!

PR #103, opened 2019-01-16
add .ruby-version file

adamjonas:
This file is necessary for the [netlify](//netlify.com) build process and is a good practice for ruby apps in general.


jnewbery:
Works fine for me. @adamjonas - can you also update the comment in the `Gemfile` that says 'If you update the version here...' to also reference this file.

@harding , @moneyball - can you confirm this doesn't cause any issues for you?

jnewbery:
tested ACK 0bd5a2281ca6d5ba33310a5021d4e685bdc20fe0

harding:
Tested ACK 0bd5a2281ca6d5ba33310a5021d4e685bdc20fe0

PR #104, opened 2019-01-20
Newsletters: add #30 (2019-01-22)

harding:
A short newsletter this week due to a lack of news (but with plenty of interesting merges).  While scanning merges, I noticed that C-Lightning changed the name of a file we had previously linked to, so this PR also updates those links.

### DIFF
--- a/_posts/en/newsletters/2018-11-20-newsletter.md
+++ b/_posts/en/newsletters/2018-11-20-newsletter.md
@@ -346,7 +346,7 @@ newsletter's author.
 [cjp rz]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-November/001498.html
 [zmn rz]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-November/001547.html
 [zmn rz packetswitch]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-November/001553.html
-[cl plugin]: https://github.com/ElementsProject/lightning/blob/master/doc/plugins.md
+[cl plugin]: https://github.com/ElementsProject/lightning/blob/master/doc/PLUGINS.md
 [rusty wrapup]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-November/001569.html
 [ln1.1 outline]: https://github.com/lightningnetwork/lightning-rfc/wiki/Lightning-Specification-1.1-Proposal-States
 [input consolidations]: https://en.bitcoin.it/wiki/Techniques_to_reduce_transaction_fees#Consolidation
--- a/_posts/en/newsletters/2019-01-22-newsletter.md
+++ b/_posts/en/newsletters/2019-01-22-newsletter.md
@@ -1,0 +1,127 @@
+---
+title: 'Bitcoin Optech Newsletter #30'
+permalink: /en/newsletters/2019/01/22/
+name: 2019-01-22-newsletter
+type: newsletter
+layout: newsletter
+lang: en
+---
+This week's newsletter describes a proposed LN feature to allow making
+spontaneous payments and provides our longest-ever list of notable code
+changes to popular Bitcoin infrastructure projects.
+
+## News
+
+- **PR opened for spontaneous LN payments:** LN protocol developer
+  Olaoluwa Osuntokun opened a [pull request][spontaneous payments] to
+  allow an LN node to send a payment to another node without first
+  receiving an invoice.  This takes advantage of LN's Tor-like onion
+  routing by allowing a spender to choose a preimage, encrypt it so
+  that only the receiver's node can decrypt it, and then route a payment
+  along LN like normal using the hash of the preimage.  When the
+  payment reaches the receiver, they decrypt the preimage and disclose
+  it to the routing nodes in order to claim the payment.
+
+    Spontaneous payments
+    help in cases where users just want to do ad hoc payment
+    tracking, for example you initiate a 10 mBTC withdrawal from an
+    exchange and either 10 mBTC shows up in your balance within a few
+    moments or you contact support.  Or you just publish your node's
+    information and users can send you donations without having to get
+    an invoice first.  For tracking specific payments, users
+    should still continue to generate invoices which can be uniquely
+    associated with particular orders or other expected payments.
+
+    Osuntokun's pull request for LND is still marked as a work in
+    progress as of this writing, so we don't know yet when the feature
+    will become generally available to LND users or whether other LN
+    implementations will also provide the same feature in a compatible
+    way.
+
+## Optech recommends
+
+If you prefer listening to audio rather than reading the weekly Optech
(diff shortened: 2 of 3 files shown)


_posts/en/newsletters/2019-01-22-newsletter.md
```diff
@@ -0,0 +1,111 @@
...
+
+## News
+
+- **PR opened for spontaneous LN payments:** LN protocol developer
+  Olaoluwa Osuntokun opened a [pull request][spontaneous payments] to
+  allow an LN node to send a payment to another node without first
+  receiving an invoice.  This takes advantage of LN's Tor-like onion
+  routing by allowing a spender to choose a preimage, encrypt it so
+  that only the receiver's node can decrypt it, and then route a payment
+  along LN like normal using the the hash of the preimage.  When the
```
practicalswift: "the the" should be "the" :-)

_posts/en/newsletters/2018-11-20-newsletter.md
```diff
@@ -346,7 +346,7 @@ newsletter's author.
 [cjp rz]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-November/001498.html
 [zmn rz]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-November/001547.html
 [zmn rz packetswitch]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-November/001553.html
-[cl plugin]: https://github.com/ElementsProject/lightning/blob/master/doc/plugins.md
+[cl plugin]: https://github.com/ElementsProject/lightning/blob/master/doc/PLUGINS.md
```
jnewbery: Have you considered linking to a specific commits, eg https://github.com/ElementsProject/lightning/blob/018a3f1d58e38e464e32cbee26826a0bd63aab59/doc/PLUGINS.md to prevent links to files in repos from breaking like this?

_posts/en/newsletters/2019-01-22-newsletter.md
```diff
@@ -0,0 +1,111 @@
...
+  payment reaches the receiver, they decrypt the preimage and disclose
+  it to the routing nodes in order to claim the payment.
+
+    Allowing spontaneous payments allows nodes to accept donations
```
jnewbery: You've led with donations here, and then also mentioned 'ad hoc payment tracking'. I'm not sure if donations will be predominant use case here. You could consider starting with a more general statement like 'Spontaneous payments may open up new use cases where invoices and proof of payment are not required. For example...'

_posts/en/newsletters/2019-01-22-newsletter.md
```diff
@@ -0,0 +1,111 @@
...
+    without using extra software for generating invoices.  It can also
+    help in cases where users just want to do ad hoc payment
+    tracking, for example you initiate a 10 mBTC withdrawal from an
+    exchange and either 10 mBTC shows up in your balance within a few
+    moments or you contact support.  For serious payment tracking, users
```
jnewbery: I don't much like the 'serious' here. Who's to say ad hoc payments aren't serious? Perhaps s/serious payment tracking/more significant payments/ or similar.

_posts/en/newsletters/2019-01-22-newsletter.md
```diff
@@ -0,0 +1,111 @@
...
+  and setting the default onchain feerate for justice transactions
+  (breach remedy transactions).
+
+- [LND #2198][] gives the `sendcoins` RPC a new `sweepall` parameter
+  that will spend all of the wallet's bitcoins to the specified address
+  without the user having to manually specify the amount.
+
+- [C-Lightning #2232][] extends the `listpeers` command with a new
+  `funding_allocation_msat` field that returns the amounts initially
+  placed into a channel by each peer.
```
jnewbery: I think adding some text similar to that in the PR "At the moment there are only one-sided channel opens, but dual-funded channels will [hopefully] be available in the near future" adds more context.

_posts/en/newsletters/2019-01-22-newsletter.md
```diff
@@ -0,0 +1,111 @@
...
+
+- [C-Lightning #2234][] extends the `listchannels` RPC to take a
```
jnewbery: I was going to ask whether this was the right PR since the PR title is so unspecific and lacking in notes, but you're right. Good code archeology!

_posts/en/newsletters/2018-11-20-newsletter.md
harding: Yeah.  I think it's a toss-up between ensuring our links don't break and helping users find up-to-date information.  For now, I opt towards the later, but I have no strong preference.  FYI, the htmlproofer command used in the Makefile to check internal links (among other things) can also check external links---this is noisy and wasteful of network resources, so it shouldn't be done automatically from Travis, but we can run it every few months and fix up links then.

_posts/en/newsletters/2019-01-22-newsletter.md
harding: Slight push back on this one.  I don't actually see why that's relevant.  Also, the code doesn't implement that (it's left as a fixme), so I'm not sure it should be described in those terms.  Code:

```diff
+
+               // FIXME @conscott : Modify this when dual-funded channels
+               // are implemented
+               json_object_start(response, "funding_allocation_msat");
+               if (channel->funder == LOCAL) {
+                       json_add_u64(response, pubkey_to_hexstr(tmpctx, &p->id), 0);
+                       json_add_u64(response, pubkey_to_hexstr(tmpctx, &ld->id),
+                                       channel->funding_satoshi * 1000);
+               } else {
+                       json_add_u64(response, pubkey_to_hexstr(tmpctx, &ld->id), 0);
+                       json_add_u64(response, pubkey_to_hexstr(tmpctx, &p->id),
+                                       channel->funding_satoshi * 1000);
+               }
+               json_object_end(response);
```

_posts/en/newsletters/2019-01-22-newsletter.md
harding: Thanks!  Actually, I almost never read the PRs, so I'm glad you check them.  In the case of C-Lightning, the code archeology always goes the other way---the trick isn't finding out what the PR does, it's finding out which PR the commit goes with!

harding:
Pushed a commit adding an "Optech recommends" section point to Max Hillebrand's readings of the Optech newsletter.  I listened to a few before posting, and they really were great!  (Also, I'm blown away by the fact that he stuck with the year-in-review newsletter for 40 minutes of reading!  And he still thanked us at the end instead of damning us for publishing something so long!)

_posts/en/newsletters/2018-11-20-newsletter.md
harding: I didn't know Travis now supported scheduled jobs, nice!  I've missed that feature from Jenkins.  I'll open a PR implementing that for our Makefile, and if it works for you, you should have access to add the scheduled job.

PR #106, opened 2019-02-02
Newsletters: add #32 (2019-02-05)

harding:
Todo:

- [x] Add info about the Chaincode Residency (@jnewbery)
- [ ] (Optional) Closer to publication, check to see if LND 0.5.2 final is [released](https://github.com/lightningnetwork/lnd/releases) and add short note (@harding)

### DIFF
--- a/_posts/en/newsletters/2019-02-05-newsletter.md
+++ b/_posts/en/newsletters/2019-02-05-newsletter.md
@@ -1,0 +1,302 @@
+---
+title: 'Bitcoin Optech Newsletter #32'
+permalink: /en/newsletters/2019/02/05/
+name: 2019-02-05-newsletter
+type: newsletter
+layout: newsletter
+lang: en
+---
+This week's newsletter includes an announcement of the 2019 Chaincode
+Residency program, summarizes a few talks from the Stanford Blockchain
+Conference, and provides the usual list of notable code changes in
+popular Bitcoin infrastructure projects.
+
+## Action items
+
+- **Apply to the Chaincode Residency:** Bitcoin Optech encourages any
+  engineer who is interested in spending summer contributing to open source
+  Bitcoin and Lightning projects to apply to the Chaincode Residency. Full
+  details of the residency are in the *News* section below.
+
+## News
+
+- **2019 Chaincode Residency:** Chaincode Labs opened applications for its
+  [fourth residency program][residency] to be held in New York over
+  summer 2019. The program combines a 3 week seminar and discussion series
+  covering Bitcoin and Lightning protocol development with a two month period for
+  working on an open source Bitcoin or Lightning project under the guidance of an
+  established protocol developer. The list of [confirmed speakers and
+  mentors][residency speakers] includes some of the most prolific contributors to
+  Bitcoin and Lightning. The previous Chaincode Residency (focused on Lightning
+  Network applications) was covered in [Newsletter #21][].
+
+    Chaincode is inviting developers who want to contribute to open source
+    Bitcoin and Lightning protocol projects to [apply to the
+    residency][residency apply]. Applicants from all backgrounds are welcomed,
+    and Chaincode will cover travel and accommodation costs and provide a stipend
+    to support living expenses.
+
+[residency]: https://residency.chaincode.com
+[residency speakers]: https://residency.chaincode.com/#mentors
+[residency apply]: https://residency.chaincode.com/#apply
+
+## Notable talks from the Stanford Blockchain Conference
+
+The [third edition][sbc] of this annual conference (formerly named BPASE)
+included more than two dozen presentations across three days.  We found
+the following topics especially interesting from a Bitcoin perspective,
+although we encourage anyone who wants to learn more to look at the
+[transcripts][] provided by Bryan Bishop or the videos provided by
+the organizers ([day 1][], [day 2][], [day 3][]).
+
+- **Accumulators for blockchains** presented by Benedickt Bünz
+  ([transcript][accumulators txt], [video][accumulators vid]).  Bitcoin
+  full nodes maintain a ledger (called the UTXO set) that stores the
+  ownership information for every currently-spendable grouping of
+  bitcoins.  Currently, that ledger contains over 50 million entries
+  and uses about three gigabytes of disk space.  This allows nodes who
+  receive a transaction to ensure the bitcoins being spent exist in
+  the UTXO set, retrieve the ownership information for those bitcoins,
+  and verify that information against the signature and other witness
+  data provided in the transaction.
+
+    But what if we also asked the spender to provide a copy of the
+    ownership information in their transaction along with a
+    cryptographic proof that the the information is actually in the UTXO
+    set?  In that case, we wouldn't need to store the whole set, we'd
+    only need to store a commitment to a set we knew was accurate and
+    that could be referenced by accurate proofs.  RSA accumulators are
+    one idea (among several others) for how to create that commitment
+    and its related proofs.  Using an accumulator, the size of the UTXO
+    commitment that nodes would store on disk would be tiny compared to
+    the full state.  Transactions would increase in size due to needing
+    to provide ownership data and a proof that they were part of the
+    current UTXO set, but the size increase would be modest (assuming
+    current typical transactions, about 70 bytes of ownership
+    information per input plus a proof that could be aggregated down to
+    about 325 bytes per block).
+
+    Other considerations affect the suitability of accumulators to
+    the task, including being relatively new to cryptography,
+    requiring either a well-studied trusted setup or a
+    more novel untrusted setup, and potentially making blocks take
+    longer to verify than the current system given that accumulator
+    verification is about 100x slower than alternative systems using
+    merkle trees.  In a new development compared to a previous
+    version of this talk given at Scaling Bitcoin 2018, the speaker
+    describes a potential major speedup to practical verification.
+
+    In summary, RSA accumulators remain an interesting area of investigation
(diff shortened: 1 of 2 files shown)


jnewbery:
I've added my proposed wording for the Chaincode residency announcement: https://github.com/bitcoinops/bitcoinops.github.io/pull/106/commits/523761a3d3e94439d9bdfe64d929ab0b73644c21. As previously discussed, I'm recusing myself from whether we include this, so the decision for whether to include, edit or exclude the wording is with @harding, @moneyball and @bitschmidty.

I've had a quick scan over the rest of the newsletter and it looks good. I plan to do a full review tomorrow.

harding:
Untested ACK 523761a .  Thanks!

moneyball:
Untested ACK 523761a

practicalswift:
ACK 523761a3d3e94439d9bdfe64d929ab0b73644c21

_posts/en/newsletters/2019-02-05-newsletter.md
```diff
@@ -0,0 +1,283 @@
...
+although we encourage anyone who wants to learn more to look at the
+[transcripts][] provided by Bryan Bishop or the videos provided by
+the organizers ([day 1][], [day 2][], [day 3][]).
+
+- **Accumulators for blockchains** presented by Benedickt Bünz
+  ([transcript][accumulators txt], [video][accumulators vid]).  Bitcoin
+  full nodes maintain a ledger (called the UTXO set) that stores the
+  ownership information for every currently-spendable grouping of
+  bitcoins.  Currently, that ledger contains over 50 million entries
+  and uses about 3.8 gigabytes of disk space.  This allows nodes who
```
jnewbery: https://statoshi.info/dashboard/db/unspent-transaction-output-set says 2.76GiB. Any idea where the discrepancy comes from?

_posts/en/newsletters/2019-02-05-newsletter.md
```diff
@@ -0,0 +1,283 @@
...
+  receive a transaction to look up the bitcoins being spent by that
```
jnewbery: Storing the UTXO set allows the node to make two separate checks at verification time:
1. _existence_ of the UTXO
2. contents of the ScriptPubKey

Those are two functions that can be separated, as is done in Cory Fields's [UHS proposal](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2018-May/015967.html), Tadge Dryja's utreexo proposal, or the accumulator proposal described here.

Consider rewording this sentence to: "...allows nodes who receive a transaction to verify that the bitcoins being spent exist in the UTXO set, retrieve the ownership information...".

I think that emphasizes point (1) a bit better. The current wording makes it sound a bit like the 'looking up' part is incidental.

_posts/en/newsletters/2019-02-05-newsletter.md
```diff
@@ -0,0 +1,283 @@
...
+    verification is about 100x slower than alternative systems using
+    merkle trees.  In a new development compared to a previous
+    version of this talk given at Scaling Bitcoin 2018, the speaker
+    describes a potential major speedup to practical verification.
+
+    In summary, RSA accumulators remain an interesting area of research
+    in to how to reduce full node storage requirements.  At present,
+    it's debatable whether this would provide much advantage for
+    Bitcoin---with today's disk sizes, few users are probably
+    discouraged by having to store a mere 4 GB---but it could
```
jnewbery: It may be worth noting that the chainstate lookups need to be reasonably fast for blocks to be validated quickly. Ideally the UTXO would all be stored in memory.

Up to you whether you think this is worth including. It might be too much detail, but it is a consideration when thinking about Bitcoin scalability regarding UTXO set size growth.

_posts/en/newsletters/2019-02-05-newsletter.md
```diff
@@ -0,0 +1,283 @@
...
+    significantly improve usability if block sizes were to grow
+    considerably or if Bitcoin were to change to a more trusted model
```
jnewbery: Nit: I'm a little wary of talking about a 'change to a more trusted model'. I would expect any 'assume UTXO set' security model would be optional, and full node operators would have the option of fully validating from genesis if they wanted.

_posts/en/newsletters/2019-02-05-newsletter.md
```diff
@@ -0,0 +1,283 @@
...
+  your node in some other way, such as by connecting to a part of the
+  Internet from which you don't have many other peers).  Previously,
+  Bitcoin Core banned the IP addresses of misbehaving peers for a period
+  of time (default of 1 day); this was easily circumvented by attackers
+  with multiple IP addresses.  This solution gives those peers a chance
+  to be useful but provides priority to potentially more helpful peers.
+  If you manually ban a peer, such as by using the `setban` RPC,
+  connections from that peer will still be rejected.
+
+- [Bitcoin Core #13926][] adds a new `bitcoin-wallet` tool to the
```
jnewbery: Nit: note that this is an offline tool: "adds a new offline `bitcoin-wallet` tool..."

_posts/en/newsletters/2019-02-05-newsletter.md
```diff
@@ -0,0 +1,283 @@
...
+  executables Bitcoin Core provides.  Without needing to use any RPCs,
+  this tool can currently create a new wallet file or display some basic
+  information about an existing wallet, such as whether the wallet is
+  encrypted, whether it uses an HD seed, how many transactions it
+  contains, and how many address book entries it has.  This helps people
+  who have a wallet file that hasn't been synced to the most recent chain
+  tip; they can do a quick inspection on the wallet to see if it's
+  interesting before they perform the lengthy rescan necessary to import
+  it.
```
jnewbery: Nit: note that the tool is expected to be enhanced in future to offer more functionality, such as salvaging damaged wallet files or topping up keypools.

moneyball:
utACK. I have a few questions noted above but no showstoppers. Very educational newsletter!

_posts/en/newsletters/2019-02-05-newsletter.md
harding: I found "offline" a bit ambiguous here (the way most people use `bitcoin-cli` is offline too, in the sense that it's only querying over localhost).  I changed a few words in the following sentence that should hopefully convey the same point, but please feel free to edit more if you think it's not clear enough.

_posts/en/newsletters/2019-02-05-newsletter.md
harding: At the current height, `gettxoutsetinfo` on my node returns the following two fields:

```json
  "bogosize": 3814123285,
  "disk_size": 2971132652
```

I previously went with bogosize, which is a deterministic size that all nodes should report the same when on the same chaintip.  Looks like Lopp is using disk_size, which can vary between different nodes.  I think there's an argument to be made for bogosize being the true size of the data, but given that the disk size is what people really care about (and what affects performance the most), I pushed a commit updating this to "about three gigabytes".

harding:
Pushed an updated commit addressing the excellent feedback from @jnewbery and @moneyball (thanks!).  It makes the newsletter longer, but hopefully the extra clarity and accuracy is worth it.

@moneyball To answer your questions:

> Does miniscript support functions? ie can the personal policy be in a function instead of inlined?

I don't see that in the described grammar and I wouldn't expect it to be added, as it'd complicate the language significantly.  However, when using a miniscript library in your program, I'd expect you to be able to extend the base class with custom functions or just throw normal functions into policy definitions, e.g.: `policy = "thres(3,C,C,C," + myscript(foo, bar) + ")"`

> Will Schnorr/Taproot allow for that personal policy to be private and not broadcast to the blockchain

Presumably miniscript will allow users of Schnorr-based multiparty signature schemes (e.g. musig) to specify they want that rather than onchain multisig.  In that case, their use of multisigs will be invisible to block chain analysts.  For Taproot, the actual policy used will remain private if the funds are spent cooperatively between the people who controlled those funds; in the non-cooperative case, parts or all of the policies used will have to be revealed on the block chain.

> (or even shared with the participants in the 3 of 4)?

The person or people responsible for securing the funds sent to a particular encumbrance (e.g. script, merklized script, taproot, etc...) needs to know the entirety of terms in that encumbrance in order to ensure nobody has inserted any conditions that allow them unilateral ability to spend the funds.  Miniscript doesn't change that obligation but it does make the review easier by allowing a miniscript-compatible encumbrance to be decompiled back into a policy that can be checked for equality at some level to the desired policy (e.g. the outer `thres` in the example in the newsletter).

Unrelated to miniscript, it might be possible for someone to use zero-knowledge proofs to prove a set of opcodes wouldn't reduce the security of a script without revealing those opcodes to any outside parties when cooperatively creating a receive address, but that's not something I'm aware of anyone working on.  Additionally, even if you did that, you might need to reveal the encumbrance when spending funds sent to the address.

> If yes then will miniscript be able to support that?

I think it should be easy for miniscript to be extended to support multi-party Schnorr-based signatures, generic MAST, and Taproot-enabled MAST.  Assuming we don't lose any opcodes or flexibility along the way to implementing those features, I can't see why any existing miniscript couldn't make optimal use of those improvements for both efficiency and privacy.

_posts/en/newsletters/2019-02-05-newsletter.md
jnewbery: Looks great. Thanks!

jnewbery:
Merged.

> Very educational newsletter!

Agree. I really enjoyed this one. It was a great summary of a lot of very technical information. Thanks Dave!

Issue #35, opened 2018-07-20
[Post-launch] Clean up HTML/CSS separation


harding:
After launch craziness is over, payoff technical debt by moving CSS rules specified in `style` tags to a SASS/CSS file.

I volunteer to do this.

jnewbery:
Thanks! 🙂

PR #112, opened 2019-02-24
Newsletters: add #35 (2019-02-26)

harding:


_includes/references.md
```diff
@@ -19,42 +20,30 @@
...
+in the included file is redefined, the cached file will be used, so the
+loop will only be run once no matter how many times this file is
+included in documents.  See https://github.com/jekyll/jekyll/pull/7108
+for details --> {% endcomment %}
+{% for i in (1..400) %}
+{% assign i_padded = "0000" | append: i | slice: -4, 4 %}
+[BIP{{i}}]: https://github.com/bitcoin/bips/blob/master/bip-{{i_padded}}.mediawiki
+{% endfor %}
+
+{% comment %}<!-- Later link definitions supercede earlier definitions.
```
practicalswift: Should be "supersede"?

practicalswift:
ACK 098b161021fca6c3a8e30781baa879083b46eef8

Thanks for yet another excellent newsletter!

_posts/en/newsletters/2019-02-26-newsletter.md
```diff
@@ -0,0 +1,152 @@
...
+      and signatures for k-of-n multisig security, MuSig can provide the
+      same security using just one public key and one signature---reducing
+      block chain space, improving verification efficiency, increasing
+      privacy, and allowing much larger sets of signers than supported
+      by Bitcoin Script's current byte-size and signature-operation
+      limits.  However, the multi-step protocol requires especially
+      careful management of secret nonces; Poelstra's post details
+      how libsecp256k1-zkp attempts to minimize the risk of
+      nonce-related failures and teases the possibility of even better
+      solutions in the future.
```
moneyball: I'm not sure you want to cover this level of detail, but another point you could make is that there are a few approaches to using Musig:
1) Threshold signatures, as implied in your text above, but you might want to point out that a drawback with this approach is loss of accountability. If the signing parties care about public or private auditing, this could be a non-starter for them.
2) It is possible to achieve k of n multisig with accountability and still get the benefits of 1 sig/1 pubkey for the direct spending case (ie most common signing scenario) if Taproot is used. It can be structured in a couple of different ways:
a) If there is a clear spending case that is most likely and most convenient, such as 2 online keys in a 2of3 mulitsig hot wallet, then the Taproot direct spending key can be the threshold 2 online keys, and the keytree can include the other spending options (such as the 3rd key, perhaps in a cold wallet) paired with either of the online keys. While this configuration gives you accountability, the drawback is that in the scenarios where the cold wallet key is used, you would lose the blockchain efficiency, privacy, and fungibility benefits as the spending script would be part of the transaction.
b) If all n keys of a k of n are all online, making it relatively convenient to do the interactive signing protocol for any of the n keys, then it could make sense to structure the direct spending key as an n of n key, with all k pairing signing configurations as part of the keytree. In this scenario, any k signing participants could prove to the other n - k parties that they have sufficient signing power to approve spending the funds, so the other n - k parties might as well sign the n of n direct spending key, with the benefit being that in any signing scenario there would only be 1 signature and 1 pubkey broadcast. The accountability in this scenario would be private and not public, although it could be made public if it were important for the use case.

moneyball:
utACK

harding:
Note: while investigating @moneyball's concerns above about musig audibility, I noticed that [Poelstra](https://download.wpsoftware.net/bitcoin/2018-10-scaling-proposal/slides.pdf) spells "adaptor signatures" with an "o" but my spell checker wants to spell it with an "e".  Here's some random (but detailed) [article](https://writingexplained.org/adapter-vs-adaptor-difference) about the spelling.  I don't care myself either way, but if anyone else wants to express a preference for Poelstra's spelling, I'm happy to change the spelling here.

_posts/en/newsletters/2019-02-26-newsletter.md
harding: Re: point 1, are you sure there's a loss of "private auditing"?  Can't the subset of participating signers publish the transcript of the data exchanged during signature generation?  (Alternatively, can't participating signers each commit to their intention to sign before starting and then commit to that set of commitments in signature, e.g. using sign-to-contract?)

harding: Ignore my previous comment.  After sleeping on it, I realized the methods I describe only work in the honest case; in the pathological case, the colluding signers can spend the money without generating an audit trail.  Accountability that only works in the honest case is pretty useless.  :-)

I'll try to work a brief summary of your concern in point 1 into the newsletter description.  Also your mention of taproot reminds me that Poelstra's blog post doesn't itself mention creating or verifying the pay-to-contract pubkeys used for taproot; I need to check to see if the library supports those directly, which might also be worthy of a mention.  (I know Blockstream has code for that somewhere, since it's integral to the Elements design and I've used it myself via their old [contracthash](https://github.com/Blockstream/contracthashtool) tool.)  I'm probably not going to mention your point 2 workarounds, which are clever but also IMO something not directly related to the functions being made available in libsecp256k1-zkp (but I'll definitely keep them in mind if we cover threshold signing in more detail in the future).

Thanks for teaching me something!

moneyball:
utACK c02adf9

_posts/en/newsletters/2019-02-26-newsletter.md
```diff
@@ -0,0 +1,156 @@
...
+  describing tools and techniques developers have found to improve their
+  efficiency.  Although some are specific to Bitcoin Core and C++
+  development, others more generally apply to anyone who develops with
+  git or GitHub.
+
+- [LND #2419][] creates a time-based buffer for certain objects stored
+  in memory that allows them to be freed after a certain period of
+  non-use.  Previously, some unused objects were being kept in memory
+  indefinitely, leading to user-reported crashes when the program
+  exceeded the system's free memory.
```
jnewbery: This seems to vague to be particularly interesting to readers. The underlying issue could potentially be a DoS vector, so I'd suggest either:

- update this summary to say that this closes a potential DoS vector where an adversary could asymmetrically exhaust node memory.
- remove it entirely.

_posts/en/newsletters/2019-02-26-newsletter.md
```diff
@@ -0,0 +1,156 @@
...
+- [C-Lightning #2380][] requires transactions have at least one
+  confirmation before the wallet will attempt to spend their bitcoins by
+  default.  This fixes a problem where the wallet would attempt to spend
+  its own unconfirmed change outputs but those payments would sometimes
+  get stuck because the earlier payments weren't confirming quickly.
+  Several RPCs related to payments receive a `minconf` parameter that
+  defaults to `1` but can be set to `0` to continue the old behavior or
+  set to a higher value if desired.
+
+- [Eclair #821][] adds more than a dozen heuristics used to help find
```
jnewbery: I only count three heuristics: `cltvDeltaFactor`, `ageFactor` and `capacityFactor`.

_posts/en/newsletters/2019-02-26-newsletter.md
harding: I especially appreciated this criticism.  I was told this solved a significant number of user-reported problems, but I couldn't see how exactly it did that, so I was indeed vague.  It turned out that there was a second half to the PR that was just merged today.  I'm deleting this item for this week and will reconsider it (with improvements) for next week after I've reviewed the second PR.  Thanks for calling me out on my vague hand waving!

harding:
Pushed updates for @jnewbery's comments.  Thanks to him, @moneyball, and @practicalswift for the excellent feedback this week!

jnewbery:
utACK 81932a20f6ebb65a4874a2b21954e7ef76f052ef. Great stuff Dave! :boom:

PR #117, opened 2019-03-06
Revamp index and content listings

harding:
We originally created the homepage for this site before we had any content.  This revamps it a bit by doing the following:

1. Linking to all the content we now have: workshops, newsletters, dashboard, blog posts, and a WIP scaling book.
2. Replaces the "posts" listing on the main page with a link to our most recent content in both the blog and the newsletters, limiting the display to the most recent three items.
3. Adds a new index page for the blog posts and links to it from the main page
4. Displays excerpts for all the newsletters and blog posts.  This required writing a few excerpts for pages that didn't have nice ones by default

I'm biased, but I think it makes the homepage much more approachable if we were to do something like #116

PR #120, opened 2019-03-14
Update about page

### LINKED ISSUE #83: Add funding model
We've always been transparent about our funding model in interviews and podcasts (eg https://bitcoinmagazine.com/articles/chaincode-devs-google-alumni-create-industry-group-help-bitcoin-scale/ and https://talesfromthecrypt.libsyn.com/tales-from-the-crypt-36-john-newbery). We should put that information on the website so it's easier for people to find.

jnewbery:
This PR:

- moves the introductory text and sponsors to the about page, making the front page less cluttered
- adds a 'funding' section to the about page (closes #83)

### DIFF
--- a/index.md
+++ b/index.md
@@ -8,31 +8,22 @@ layout: home
 {:.logo}
 ![Optech Logo](/img/logos/optech-notext.png)
 
-The Bitcoin Operations Technology Group (Optech) works to bring the best
-open source technologies and techniques to Bitcoin-using businesses in
-order to lower costs and improve customer experiences.
+Bitcoin Optech helps Bitcoin users and businesses integrate scaling
+technologies.
 
-An initial focus for the group is working with its member organizations to
-reduce transaction sizes and minimize the effect of subsequent transaction fee
-increases.  We provide [workshops][], [documentation][scaling book],
-[weekly newsletters][], [original research][dashboard], [case studies
-and announcements][blog], and help facilitate improved relations between
-businesses and the open source community.
+We provide [workshops][], [documentation][scaling book], [weekly
+newsletters][], [original research][dashboard], [case studies and
+announcements][blog], and help facilitate improved relations between businesses
+and the open source community.
+
+[Learn more about us][about].
 
 [scaling book]: https://github.com/bitcoinops/scaling-book
 [workshops]: /workshops
 [weekly newsletters]: /en/newsletters/
 [dashboard]: https://dashboard.bitcoinops.org/
 [blog]: /en/blog/
-
-Optech does not exist to make a profit, and all materials and documentation
-produced are released under the MIT license. We are supported by our generous
-founding sponsors and contributions from member companies.
-
-If you're an engineer or manager at a Bitcoin company or an open source contributor and you'd like to be a part of this, please
-contact us at [info@bitcoinops.org](mailto:info@bitcoinops.org).
-
-{% include newsletter-signup.html %}
+[about]: /about
 
 <h2>Recent newsletters and blog posts</h2>
 <ul class="post-list">
@@ -54,6 +45,6 @@ contact us at [info@bitcoinops.org](mailto:info@bitcoinops.org).
 
 <p class="rss-subscribe">View more <a href="/en/newsletters/">newsletters</a> or <a href="/en/blog">blog posts</a>. Learn about new content by subscribing via email or <a href="{{ "/feed.xml" | relative_url }}">RSS</a>.</p>
 
-{% include members.html %}
+{% include newsletter-signup.html %}
 
-{% include sponsors.html %}
+{% include members.html %}
--- a/about.md
+++ b/about.md
@@ -21,12 +21,19 @@ businesses and the open source community.
 [dashboard]: https://dashboard.bitcoinops.org/
 [blog]: /en/blog/
 
-Optech does not exist to make a profit, and all materials and documentation
-produced are released under the MIT license. We are supported by our generous
-founding sponsors and contributions from member companies.
-
 If you're an engineer or manager at a Bitcoin company or an open source contributor and you'd like to be a part of this, please
 contact us at [info@bitcoinops.org](mailto:info@bitcoinops.org).
+
+## Funding
+
+Optech does not exist to make a profit, and all materials and documentation
+produced are released under the MIT license.
+
+Seed funding was provided by Wences Casares and John Pfeffer to cover outside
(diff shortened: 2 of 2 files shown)


moneyball:
tACK. I like the improvements. Once we have the remaining todo items done I think it is ready to merge.

_includes/associates.html
```diff
@@ -5,7 +5,7 @@
     <ul class="associates n-column-list">
```
harding: You can get Jekyll to sort this list for you by doing:

```diff
$ git diff
diff --git a/_includes/associates.html b/_includes/associates.html
index a603edc..419c61b 100644
--- a/_includes/associates.html
+++ b/_includes/associates.html
@@ -1,4 +1,4 @@
-{% assign associates = site.data.associates.associates %}
+{% assign associates = site.data.associates.associates | sort: "name" %}
 
 <div id="associates">
 
```

index.md
```diff
@@ -8,29 +8,9 @@ layout: home
...
-
-Optech does not exist to make a profit, and all materials and documentation
-produced are released under the MIT license. We are supported by our generous
-founding sponsors and contributions from member companies.
-
-If you're an engineer or manager at a Bitcoin company or an open source contributor and you'd like to be a part of this, please
-contact us at [info@bitcoinops.org](mailto:info@bitcoinops.org).
+Bitcoin Optech works to bring the best open source technologies and techniques
+to Bitcoin-using businesses in order to lower costs and improve customer
+experiences. [Learn more](./about).
```
harding: Nitpick: I have a slight preference for "/about" rather than "./about".  Always using absolute paths for internal links is nice because it's context-free when grepping for stuff or doing global substitutions.

harding:
I like it also.  One suggestion (certainly not required): I think it would be better if the "Subscribe to our newsletter" content was moved to immediately above the "Member companies" sub-head.  I feel like putting it at the top looks aggressive, but putting it below a listing of some of our (hopefully) high-quality content is just providing a reasonable keep-in-touch option.

_includes/associates.html
jnewbery: Thanks. That's a much better idea. Note that this sorts by first name, rather than last name, but I don't have any preference either way (as long as it's a stable and arbitrary ordering 🙂)

index.md
jnewbery: Fixed!

jnewbery:
> I think it would be better if the "Subscribe to our newsletter" content was moved to immediately above the "Member companies" sub-head.

Yes, good idea. Done!

index.md
```diff
@@ -8,31 +8,9 @@ layout: home
...
-
-Optech does not exist to make a profit, and all materials and documentation
-produced are released under the MIT license. We are supported by our generous
-founding sponsors and contributions from member companies.
-
-If you're an engineer or manager at a Bitcoin company or an open source contributor and you'd like to be a part of this, please
-contact us at [info@bitcoinops.org](mailto:info@bitcoinops.org).
-
-{% include newsletter-signup.html %}
+Bitcoin Optech works to bring the best open source technologies and techniques
```
jnewbery: I think since this text appears on the about page, we could make the text even shorter and snappier on the index page. Something like:

"Bitcoin Optech helps Bitcoin users and businesses integrate scaling technologies. Learn more."

Thoughts? @harding - you raised a point about the word 'scaling' here: https://github.com/bitcoinops/bitcoinops.github.io/pull/117#issuecomment-470618659. Do you have any thoughts about its use here?

_includes/associates.html
harding: I gave up on last name sorting years ago after starting to deal with the complexity of multiple cultures, but if you want that, you can either use separate fields for given and family names or you can just use an array, e.g. in the YAML:

```
- name: [Harding, David]
```

The sort code from my earlier comment works with that, but you also need to change the way the names are printed.  E.g.:

```diff
-            <h3 itemprop="name">{{ associate.name }}</h3>
+            <h3 itemprop="name">{% for part in associate.name reversed %}{{part}} {% endfor %}</h3>
```

index.md
harding: If you think the term is appropriate, then we shouldn't stop using it.  In any case, we can always change this page again later if we move to a different phrasing, so there's no harm in this change.

_includes/associates.html
jnewbery: I think it's fine to order by first name, but thanks for sharing this!

index.md
jnewbery: Thanks. I've updated the text in the front page.

jnewbery:
Adam and Carl updated. This is ready for final review.

moneyball:
tACK 283cd5c

I do think that in the future, as a separate PR, we can either (a) reduce the size of the Optech logo to better match the now-much-shorter text at the top of the home page or (b) add a bit more text description back before Learn More. But, there is so much goodness in this PR, that I think we should merge now. I'll file a separate Issue for my comment above, and we can discuss it there.

jnewbery:
> there is so much goodness in this PR, that I think we should merge now.

We can easily merge all the changes to the associates now and wait until we're happy with the index/front page before merging those parts.

index.md
```diff
@@ -8,31 +8,7 @@ layout: home
...
-
-Optech does not exist to make a profit, and all materials and documentation
-produced are released under the MIT license. We are supported by our generous
-founding sponsors and contributions from member companies.
-
-If you're an engineer or manager at a Bitcoin company or an open source contributor and you'd like to be a part of this, please
-contact us at [info@bitcoinops.org](mailto:info@bitcoinops.org).
-
-{% include newsletter-signup.html %}
+Bitcoin Optech helps Bitcoin users and businesses integrate scaling technologies. [Learn more](/about).
```
bitschmidty: My suggestion would be to add another sentence or two here. Visually this looks "empty" on the page.

jnewbery:
#123 is merged. This PR now just contains the changes to the front page and about page.

> Visually this looks "empty" on the page.

I agree. I wanted to make the front page short and snappy, but as it is in this PR it looks a bit empty. We could add text or increase the font size :slightly_smiling_face:

jnewbery:
Updated front page to @bitschmidty's suggestion here: https://github.com/bitcoinops/bitcoinops.github.io/issues/122#issuecomment-474061811 with one minor change: 'Learn more' looked a bit lonely in a paragraph on its own, so I changed it to 'Learn more about us'.

moneyball:
tACK. I made a trivial change.

PR #40, opened 2018-07-21
Add newsletter #5 2018-07-24

harding:
One link depends merge of: https://github.com/bitcoin-core/bitcoincore.org/pull/576 (alternatively, link can be removed; it's labeled as TODO).

Includes a few helper functions, as discussed last week.  Also, this week, I clicked through all the links to make sure they pointed at the right thing (sorry about that last week).

    #include <feel_free_to_push_to_my_branch>

PR #127, opened 2019-03-31
Newsletters: add #40 (2019-04-02)

harding:


_posts/en/newsletters/2019-04-02-newsletter.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+That's your entire scriptPubKey.  You can use that in the output of a
+transaction and send it.  Note that bech32 scriptPubKeys can vary in
+size from 4 to 42 vbytes, so you need to consider that in your
+fee estimation code.  (With current segwit P2WPKH and P2WSH, the
+scriptPubKey sizes are 22 and 34 vbytes, respectively.)
+
+Your code doesn't need to be written in Python.  Reference libraries
+are also provided for c, c++, go, haskell, javascript, ruby, and rust.
```
practicalswift: Nit: Make capitalisation consistent with how we write Python.

```suggestion
are also provided for C, C++, Go, Haskell, JavaScript, Ruby, and Rust.
```

practicalswift:
ACK c448153b5a61a94947385bfe48d9264665a0a3fd modulo nit

Very nice newsletter! Thanks a lot!

_posts/en/newsletters/2019-04-02-newsletter.md
```diff
@@ -0,0 +1,269 @@
...
+  backwards compatible with all widely-used software.  As
+  discussed in an [issue][Bitcoin Core #15560] and [the project's weekly
+  meeting][core meeting segwit], starting with Bitcoin Core 0.20
+  (expected about a year from now), Bitcoin Core will default to native
+  segwit addresses (bech32) that provide additional fee savings and
+  other benefits.  Currently, many wallets and services [already support
+  bech32 addresses][bech32 adoption], and if the Bitcoin Core project
+  sees enough additional adoption in the next six months to warrant an
+  earlier switch, it will instead default to bech32 receiving addresses
+  in Bitcoin Core 0.19.  Legacy address will continue to be provided
```
moneyball: addresses?

_posts/en/newsletters/2019-04-02-newsletter.md
```diff
@@ -0,0 +1,269 @@
...
+  if the user requests them in the GUI or by RPC, and anyone who doesn't
+  want the update will be able to configure their default address type.
```
moneyball: Just to be clear, this is because of the existing addresstype option? (not because of a recent discussion/decision by the Core project)

_posts/en/newsletters/2019-04-02-newsletter.md
```diff
@@ -0,0 +1,269 @@
...
+`bc` in our code so that we can later ensure that the addresses we parse
+are for the network we expect.
+
+```python3
+HRP='bc'
+```
+
+Finally, we have a few addresses we want to check---one that should work
+and two that should fail.  (See [BIP173][] for a complete set of
+reference test vectors.)
```
moneyball: optional micro-optimization: deep link to the test vector section? https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#Test_vectors

_posts/en/newsletters/2019-04-02-newsletter.md
```diff
@@ -0,0 +1,269 @@
...
+In [16]: segwit_addr.decode(HRP, typo_address)
+Out[16]: (None, None)
+
+In [17]: segwit_addr.decode(HRP, wrong_network_address)
+Out[17]: (None, None)
+```
+
+If we get back a None for the first value (the witness version), the
+address is invalid on our chosen network, so you want to throw an
+exception up the stack so that whatever process is interfacing with the
```
moneyball: While the Python reference library doesn't provide reason(s) for the decode failure, presumably this would be possible to do allowing for a more actionable error message to be shown to the user. If this is true, should we speak to that here? Or, maybe in a subsequent week for "improving the user experience"?

_posts/en/newsletters/2019-04-02-newsletter.md
```diff
@@ -0,0 +1,269 @@
...
+[eclair api docs]: https://acinq.github.io/eclair/#introduction
+[johoe's mempool statistics]: https://jochen-hoenicke.de/queue/#0,1w
+[p2sh.info's fee estimate tracker]: https://p2sh.info/dashboard/db/fee-estimation?orgId=1
+[trampoline thread]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-March/001939.html
+[core meeting segwit]: http://www.erisian.com.au/meetbot/bitcoin-core-dev/2019/bitcoin-core-dev.2019-03-28-19.01.log.html#l-83
+[bech32 adoption]: https://en.bitcoin.it/wiki/Bech32_adoption
+[bech32 reference libraries]: https://github.com/sipa/bech32/tree/master/ref
+[segwit addr to bytes]: https://github.com/sipa/bech32/blob/master/ref/python/tests.py#L30
+[bitgo segwit]: https://blog.bitgo.com/native-segwit-addresses-via-bitgos-api-4946f2007be9
+[gemini reddit]: https://old.reddit.com/r/Bitcoin/comments/b66n0v/psa_gemini_is_full_on_with_native_segwit_and_uses/
```
moneyball: Should we be linking to "old" Reddit? ie is there more danger this link breaks in the future than using www.reddit.com?

_posts/en/newsletters/2019-04-02-newsletter.md
```diff
@@ -0,0 +1,269 @@
...
+programmer should be able to implement it from scratch in their preferred
+language without requiring anything beyond what most programming
+languages provide in their builtins and standard library.
+
+**Other bech32 sending support updates:** BitGo [announced][bitgo
+segwit] that their API now supports sending to bech32 addresses; see
+their announcement for additional details about bech32 receiving support.
+The Gemini exchange also [apparently][gemini reddit] added bech32
+sending support this week, and users report that Gemini will accept
+deposits to bech32 addresses as well.
```
moneyball: I suggest we say that Gemini now _defaults_ to bech32 addresses when generating a new receiving address on Gemini (I personally tested this).

moneyball:
tACK good stuff!

_posts/en/newsletters/2019-04-02-newsletter.md
harding: This sentence has multiple parts, so I'm not sure what you're specifically referring to, but everything is supported today.

1. If `addresstype=bech32` on master, then you can select a legacy address by unchecking the "Generate native segwit (bech32) address" box that's checked by default in the GUI (with that `addresstype` configuration).  


2. You can request any address type you want (P2PKH, P2SH-P2WPKH, or P2WPKH) via the second parameter of the `getnewaddressrpc` or you can generate the address you want using a descriptor or other RPC.

3. You can use `addresstype={legacy,p2sh-segwit,bech32}` as a configuration option to set your local default.

It was explicitly discussed (and, AFAICT, universally agreed) that users should continue to receive choice, so I'm imagining that it'll be done by the same basic mechanism (although maybe the GUI text will be rephrased).

_posts/en/newsletters/2019-04-02-newsletter.md
harding: Ha, you're on to my plan.  I was initially going to do both a "python backend" and "JS frontend" section this week, but the Python section seemed long enough.  My plan for next week's newsletter is to code up basically a version of [this](http://bitcoin.sipa.be/bech32/demo/demo.html) via a similar tutorial in the newsletter using the reference JS ECC library.  I don't want to explicitly commit to that in this week's newsletter because I haven't written it yet and so I don't know if I'll encounter any problems.

_posts/en/newsletters/2019-04-02-newsletter.md
```diff
@@ -0,0 +1,269 @@
...
+
+- [Bitcoin Core #15637][] modifies the mempool-related RPCs (such as
+  `getrawmempool`) to rename the `size` field to `vsize`.  The previous
+  value was also the vsize, so the calculation has not changed.  This
+  merged PR simply makes it clear that this is a vsize and not a
+  stripped size.  For backwards compatibility, you can start Bitcoin
+  Core with the `deprecatedrpc=size` parameter to continue using the old
+  field name, although this will be removed in a future release.
+
+- [LND #2759][] lowers the default [CLTV delta][bolt2 delta] from 144
```
jnewbery: I think there are some issues with this explanation. I've pushed a commit that I think improves it, but I'm not 100% clear of the exact details of lightning. Can you review and let me know what you think?

jnewbery:
I think this all looks good apart from the explanation of LND PR 2759. I've pushed a commit which I think improves it, but would appreciate a lightning expert looking over it if possible. Perhaps @cdecker, @Roasbeef, @halseth or @cfromknecht would be kind enough to review that section?

_posts/en/newsletters/2019-04-02-newsletter.md
```diff
@@ -0,0 +1,269 @@
...
+Pieter Wuille's [code][segwit addr to bytes] does this quite succinctly:
+
+```python3
+>>> witver, witprog = segwit_addr.decode(HRP, good_address)
+>>> bytes([witver + 0x50 if witver else 0, len(witprog)] + witprog).hex()
+'00146eafa604a503a0bb445ad1f6daa80f162b5605d6'
+```
+
+That's your entire scriptPubKey.  You can use that in the output of a
+transaction and send it.  Note that bech32 scriptPubKeys can vary in
```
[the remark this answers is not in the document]
harding: Hmm, I feel the opposite from you about this.  I think it's better to generalize the problem as "consider the actual output size whatever it uses" rather than special casing P2PKH, P2SH, P2WPKH, and P2WSH (plus prepare more special casing for the future).  I only have the parenthetical so that people aren't thinking that supporting segwit spending currently requires almost doubling the scriptPubKey size from P2PKH/P2SH (which means doubling the fee it pays).

_posts/en/newsletters/2019-04-02-newsletter.md
```diff
@@ -0,0 +1,269 @@
...
+- [Bitcoin Core #15643][] changes the Python script Bitcoin Core
+  developers use to merge commits so that the git merge message includes
+  a list of which developers approved (ACKed) the version of a PR that
+  was merged.  (This internal project change is perhaps not notable by
+  itself, but one of the tool's other features---copying the full PR
+  description into the merge message---makes it much easier for the
+  author of this section to write these merge summaries, so he
+  encourages other Bitcoin projects to investigate the advantages of
+  using this tool for automatically creating better git-based
+  documentation, as well as improving their security and auditability.)
```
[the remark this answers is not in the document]
harding: Ha, I had the exact same thought process while writing.  I don't think it's interesting enough to include.

I think I'd like to see a parallel git repository that's periodically updated to include all bitcoin/bitcoin discussion.  (I think this used to exist.)  The github-merge.py tool could then optionally reference the most recent sha512 tree commit from that repository in each PR to truly reference all the discussion.

_posts/en/newsletters/2019-04-02-newsletter.md
harding: I think my description was correct.  I'll write an explanation of why, but I suggest that we not publish this unless we get an actual LN expert (such as one of those you @ mentioned) to review whatever version we go with.  (I'll leave it up to you to keep or drop at merge time tomorrow; we can put it in next week's newsletter after we ensure it's accurate.)

harding:
965ea0852ef1df1c6db4f1cb24128c3a2662a05a LGTM despite my initial debate of it.  I'm ok-ish printing it without review from an LN expert.  I'll make the other suggested edits.

_posts/en/newsletters/2019-04-02-newsletter.md
```diff
@@ -0,0 +1,269 @@
...
+
+- **Trampoline payments for LN:** Pierre-Marie Padiou started a
+  [thread][trampoline thread] on the Lightning-Dev mailing list
+  suggesting that Alice could send a payment to Zed even if she didn't
+  know a path to his node by first sending a payment to an intermediate
+  node (Dan) and asking Dan to figure out the route the rest of the way
+  to Zed.  This would especially benefit Alice if she ran a lightweight
+  LN node that didn't attempt to keep track of the entire network.  For
+  increased privacy, Alice could use several intermediate nodes in
+  succession rather than just one (each one receiving it's own
```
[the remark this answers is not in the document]
harding: I think you interpreted it the same way I read the email, and I don't understand how your suggested change affects the meaning.  I am removing "in succession" because it actually doesn't add anything to the sentence.

harding:
All suggested edits pushed, except for those I left notes about.  I'll try to review any feedback that comes in overnight by 08:00 EDT tomorrow; after that, @jnewbery or @moneyball, please feel free to push any edits you think are necessary.

Thanks @practicalswift @moneyball @jnewbery !

cdecker:
> I think this all looks good apart from the explanation of LND PR 2759. I've pushed a commit which I think improves it, but would appreciate a lightning expert looking over it if possible.

The description is correct, and the lowering of the CLTV is definitely a welcome change :-)

harding:
@cdecker Thanks!

jnewbery:
Thanks @cdecker!

I've made one minor edit ('over a 95% savings' -> 'saving over 95%').

PR #134, opened 2019-04-09
Layout: add default opengraph image

### LINKED ISSUE #133: Add twitter cards markup
Our twitter stream right now is visually dull: https://twitter.com/bitcoinoptech. There's no image for the links so twitter just defaults to its default grey doc image. We should make it a bit more appealing to the eye with some twitter card markup: https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started.html

I'm open to suggestions about what we put there, but some things we could consider:

- default to the Bitcoin Optech logo:

Screenshot from 2019-04-09 10-34-29

- default to just the newsletter number in the Optech grey font, eg:

Screenshot from 2019-04-09 10-33-48

- occasionally have a different image if there's a special issue (eg end-of-year, Scaling Bitcoin special, segwit 2-year anniversary, etc)

harding:
Closes #133

Validator: https://cards-dev.twitter.com/validator
Test URL: http://dg2.dtrt.org/en/newsletters/2019/04/09/
How this change works: https://github.com/jekyll/jekyll-seo-tag/blob/master/docs/advanced-usage.md#setting-a-default-image

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -40,3 +40,9 @@ exclude:
    - README.md
 
 show_excerpts: true
+
+defaults:
+  - scope:
+      path: ""  ## all pages
+    values:
+      image: https://bitcoinops.org/img/logos/optech-notext.png


harding:
Oh, and the image can be overridden for a particular post by including an `image: <fully-qualified-url>` tag in the post's header YAML.

jnewbery:
ACK b516c25f23c2086af3da3a392836746d35a88be7

Thanks for grabbing this!

PR #136, opened 2019-04-19
Add Crypo Garage

jnewbery:
Crypto Garage are now members and have agreed to have their logo on the website


moneyball:
post merge tACK

PR #142, opened 2019-05-14
News46: mention xpub in addition to derivation path

harding:
One of the participants in this discussion pointed out to me that suggestion on the mailing list was to include both the xpub and the derivation path.


jnewbery:
LGTM

PR #46, opened 2018-07-27
Update LICENSE

jnewbery:
Adds MIT License.

Since this changes the license, I guess we need approval from all contributors: @harding , @jamesob , @moneyball 

fixes #45

PR #143, opened 2019-05-15
NL-#46 : fix wrong address type on vbytes analysis

ariard:
Seems to me a nit which is in opposite to the below board


harding:
I believe the existing text is correct.  The costs to spend (scriptSig or witness):

- P2PKH: 107 vbytes
- P2WPKH: 26.75 vbytes
- Taproot: 16.25 vbytes

So for P2WPKH vs Taproot, thats:


    1 - (16.25 / 26.75) = 0.393

So I think it's reasonable to say, "the cost to spend a single-key Taproot is 40% cheaper than P2WPKH."

ariard:
Ah sorry that's right, read too fast, the board doesn't show witnesses!

PR #144, opened 2019-05-20
Newsletters: add news47 (2019-05-21)

harding:
Regular length this week.

CC: @ajtowns (section about bip-anyprevout)

### DIFF
--- a/_posts/en/newsletters/2019-05-21-newsletter.md
+++ b/_posts/en/newsletters/2019-05-21-newsletter.md
@@ -1,0 +1,225 @@
+---
+title: 'Bitcoin Optech Newsletter #47'
+permalink: /en/newsletters/2019/05/21/
+name: 2019-05-21-newsletter
+type: newsletter
+layout: newsletter
+lang: en
+---
+This week's newsletter describes the bip-anyprevout soft fork proposal, summarizes
+a few technical talks from the Magical Crypto Friends conference, and
+includes our regular sections on bech32 sending support and notable
+changes to popular Bitcoin infrastructure projects.
+
+## Action items
+
+*None this week.*
+
+## News
+
+- **Proposed anyprevout sighash modes:** two weeks ago, Anthony Towns
+  [posted][email bip-anyprevout] a proposed BIP to the Bitcoin-Dev and Lightning-Dev mailing
+  lists for consideration.  The idea, [bip-anyprevout][], provides two
+  new signature hashing (sighash) modes that allow a signature to commit
+  to fewer details about the funds being spent than the default
+  [bip-taproot][] and [bip-tapscript][] sighash mode.  This enables the
+  functionality previously described by [BIP118][] with modification so
+  that it works with Taproot and reduces the risk of accidental misuse.  One of
+  the new sighash modes is
+  directly compatible with the proposed [Eltoo][] layer for LN,
+  requiring only modifications for Taproot and the addition of a
+  *chaperone signature* (described later).  A second sighash mode
+  commits to more data than necessary for Eltoo in a way that may make
+  it useful for atypical commitments in Eltoo or for use in other protocols.
+
+    A significant advantage of this proposal over BIP118 noinput is that
+    it can make use of Taproot cooperative spends, allowing the two or
+    more parties to an LN channel or other contract protocol to optionally spend
+    their funds without revealing any of the contract terms (including
+    that anyprevout was in use).
+
+    For a quick look at anyprevout, we consider it in the context of a
+    two-party Eltoo-based LN channel.  As a reminder, the key feature of
+    Eltoo is that each balance change in a channel (state update) can be
+    spent by *any* later state update, so there's no risk of publishing an
+    old state to the block chain like there is with the current
+    penalty-based LN channels.  Eltoo calls this capability *rebinding*
+    and BIP118 proposed to make rebinding available by allowing
+    signatures to skip committing to the input part of the spending
+    transaction---allowing anyone to modify that part of the
+    transaction to bind any later state they knew about.
+
+    The `SIGHASH_ANYPREVOUTANYSCRIPT` mode (any previous output, any
+    script) defined by bip-anyprevout works similarly to BIP118 noinput,
+    with the following changes.  To use anyprevoutanyscript, the public
+    key to which a signature is compared will need to use a special
+    prefix (0x00 or 0x01; not to be confused with the pubkeys used for
+    [bip-taproot's][bip-taproot] output key that uses the same prefix in
+    a different context).  Additionally, the script being evaluated will
+    need to contain at least one signature that doesn't use
+    anyprevoutanyscript or the other new sighash mode described later.
+    This non-anyprevout signature is called the *chaperone signature*
+    because, under reasonable assumptions, it prevents an anyprevout
+    signature from being misused.  (See [Newsletter #34][] for details
+    about the replay problem.)  With the correct prefix and a chaperone
+    signature, anyprevoutanyscript allows the signature to skip
+    committing to the identifier for output being spent (the outpoint),
+    that previous output's scriptPubKey, and the hash of the executed
+    taproot leaf (tapleaf).  The transaction digest to which the signature commits still includes
+    some details about the prevout, such as its bitcoin value.
+
+    Additionally, bip-anyprevout also defines another sighash mode:
+    `SIGHASH_ANYPREVOUT` that also requires the same special pubkey
+    prefix and chaperone signature, but it includes the prevout's
+    scriptPubKey and tapleaf hash in the signature.  Whereas
+    anyprevoutanyscript can allow Eltoo-like rebinding where any later
+    state can bind to any earlier state (but earlier states can't bind
+    to later states), there may be alternative protocols (and times
+    within the Eltoo protocol) where the participants want to ensure
+    that state *n* can only bind to state *n-1* and not any other state.
+
+    The proposal has begun receiving [feedback][anyprevout list] on the
+    mailing lists, so we'll provide updates in subsequent newsletters
+    summarizing any significant discussions.
+
+- **Talks of technical interest at Magical Crypto Friends conference:**
+  Bryan Bishop provided [transcripts][mcf transcripts] of about a dozen
+  talks and panels from the MCF conference two weekends ago, and the
+  conference organizers have uploaded most [videos][mcf vids].  Although
+  only one of the talks described any specific new developments, several
+  of them did discuss details and implications of technology such as
+  confidential transactions, taproot, schnorr, and other ideas related
+  to Bitcoin.  We found the following talks particularly interesting:
+
+    - A [talk][mcf andytoshi] by Andrew Poelstra about the cryptography used in
+      cryptocurrencies.  In particular, he focuses on the difficulty of
+      building systems where *everything* needs to be done correctly in
+      order to resist attack.
+
+    - A [panel][mcf nonet] by Rodolfo Novak, Elaine Ou, Adam
+      Back, and Richard Myers about using Bitcoin without direct access
+      to the Internet.  Discussion topics included satellite-based block
+      propagation, mesh networking, amateur radio, and physically
+      carrying data (sneakernets) and how they can make Bitcoin more
+      robust for current users and more accessible for users in areas
+      with limited network access.  We found particularly interesting a
+      side discussion about the security of relaying Bitcoin data---in
+      short, the Bitcoin protocol is already designed to trustlessly
+      accept data from random peers, so non-net relay doesn't
+      necessarily change the trust model.
+
+   - A [panel][mcf future ln] by Will O'Beirne, Lisa Neigut, Alex Bosworth with
+     moderation by Leigh Cuen discussing the future of LN, mostly the
+     short-term and medium-term conclusion of current development
+     efforts surrounding the LN 1.1 specification.  There are no hyped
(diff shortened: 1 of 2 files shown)


_posts/en/newsletters/2019-05-21-newsletter.md
```diff
@@ -0,0 +1,221 @@
...
+    anyprevoutanyscript or the other new sighash mode described later.
+    This non-anyprevout signature is called the *chaperone signature*
+    because, under reasonable assumptions, it prevents an anyprevout
+    signature from being misused.  (See [Newsletter #34][] for details
+    about the replay problem.)  With the correct prefix and a chaperone
+    signature, anyprevoutanyscript allows the signature to skip
+    committing to the identifier for output being spent (the outpoint),
+    that previous output's scriptPubKey, and the hash of the executed
+    taproot leaf (tapleaf).  The signature still commits to other
+    details about the prevout, such as its bitcoin value.
```
moneyball: Just to make sure I understand, is it the case that ANYPREVOUTANYSCRIPT does not commit to a specific pubkey in the output to be used, but ANYPREVOUT does?

_posts/en/newsletters/2019-05-21-newsletter.md
```diff
@@ -0,0 +1,221 @@
...
+services haven fallen behind by not providing sending support.
+
+If you haven't implemented bech32 sending support yet, we suggest you
+try to get it implemented by 24 August 2019 (the two-year anniversary of
+segwit activation).  Not long after that, Bitcoin Core's next release is
+expected to begin defaulting to bech32 receiving addresses in its GUI
+and perhaps also its API methods (see Newsletters [#40][Newsletter #40]
+and [#42][Newsletter #42]).  We expect other wallets to do the
+same---except for the ones that have already made bech32 their default
+(or even their only supported address format).
```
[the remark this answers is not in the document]
harding: Wasabi, for sure.  Also, I think if you enable segwit receiving in Electrum, you can only receive to bech32 addresses (but I haven't tested this myself).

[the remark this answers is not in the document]
harding: Oh, I realized your comment might be asking me to name the wallet(s) here.  I don't want to do that because one of the topics on the 24-week outline is specifically talking about bech32-only wallets and how they're probably going to drag everyone into supporting bech32 sending eventually ("tyranny of the minority").

_posts/en/newsletters/2019-05-21-newsletter.md
```diff
@@ -0,0 +1,221 @@
...
+users as a website in 2019 that's covered in Shockwave/Adobe Flash
+elements and that claims it's best viewed in Internet Explorer 7 (or
+see an even more [imaginative comparison][nullc bank analogy] written by Gregory
+Maxwell.)
+
+Bech32 sending is not some experimental new technology that still needs
+testing---native segwit unspent outputs currently hold [over 200,000
+bitcoins][].  Bech32 sending is also something that's easy to implement
+(see Newsletters [#38][news38 bech32] and [#40][news40 bech32]).  Most
+importantly, as more and more wallets and services upgrade to bech32
```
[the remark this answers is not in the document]
harding: Wasabi, Bitcoin Core GUI, maybe Electrum, and probably a few other wallets also have bech32 receiving by default.  I don't think it'd be fair to just link to Gemini here.

_posts/en/newsletters/2019-05-21-newsletter.md
harding: Not the pubkey.  It's the scriptPubKey (output script), the thing that specifies the conditions that must be fulfilled in order for the output to be spent.  Imagine you have two outputs each with different (but similar) scripts:

    <600000> OP_CLTV OP_DROP <A pubkey> OP_CHECKSIGVERIFY <Chaperone pubkey> OP_CHECKSIG
    <601000> OP_CLTV OP_DROP <A pubkey> OP_CHECKSIGVERIFY <Chaperone pubkey> OP_CHECKSIG

With anyprevoutanyscript, the same signature by A could spend either of those outputs (assuming it committed to a spend with an nLockTime over 601,000); with anyprevout, the signature could only spend whichever of those output scipts it explicitly committed to.

_posts/en/newsletters/2019-05-21-newsletter.md
```diff
@@ -0,0 +1,221 @@
...
+> "They can only pay legacy addresses."<br>
+> "Oh.  Let's look for another service that supports current technology."
+
+Services that only support legacy addresses are likely to become a cue
+to users that minimal development effort is being put into maintaining
+their Bitcoin integration.  We expect that it'll send the same signal to
+users as a website in 2019 that's covered in Shockwave/Adobe Flash
+elements and that claims it's best viewed in Internet Explorer 7 (or
+see an even more [imaginative comparison][nullc bank analogy] written by Gregory
+Maxwell.)
```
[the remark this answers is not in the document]
harding: Heh.  I didn't even look at the title of the thread or notice who he was arguing with.  I'm narrowing the context down to just Maxwell's last comment.  The title will still be displayed and anyone who's really curious will still be able to see the other context, but hopefully it's less of a possible affront that way.  What do you think?

harding:
Thanks for the reviews!  Pushed edits for several of @moneyball's comments; left replies for the rest.

_posts/en/newsletters/2019-05-21-newsletter.md
```diff
@@ -0,0 +1,221 @@
...
+*None this week.*
+
+## News
+
+- **Proposed anyprevout sighash modes:** two weeks ago, Anthony Towns
+  posted a proposed BIP to the Bitcoin-Dev and Lightning-Dev mailing
+  lists for consideration.  The idea, [bip-anyprevout][], provides the
+  functionality previously described by [BIP118][] with modifications to
+  work with Taproot and to reduce the risk of accidental misuse.  One of
+  the signature hashing (sighash) modes available with bip-anyprevout is
```
jnewbery: I think this jumps into talking about 'sigature hashing (sighash) modes' a little too quickly, without introducing the concept first. Perhaps just one additional introductory sentence would help:

' ... and to reduce the risk of accidental misuse.  The BIP proposes two news ways for a signature to commit to the coin that it spends. The first of these signature hashing (sighash) modes is directly compatible ...'

_posts/en/newsletters/2019-05-21-newsletter.md
```diff
@@ -0,0 +1,221 @@
...
+  requiring only modifications for Taproot and the addition of a
+  *chaperone signature* (described later).  A second sighash mode
+  commits to more data than necessary for Eltoo in a way that may make
+  it useful for atypical commitments in Eltoo or for use in other protocols.
+
+    A significant advantage of this proposal over BIP118 noinput is that
+    it can make use of Taproot cooperative spends, allowing the two or
+    more parties to an LN channel or other contract protocol to spend
+    their funds without revealing any of the contract terms (including
+    that anyprevout was in use).  However, if they do need to reveal the
```
jnewbery: I think this sentence "However, if they do need to reveal the..." is unnecessary, given the thorough treatment of taproot that you gave last week, and would be slightly in favour of removing it.

_posts/en/newsletters/2019-05-21-newsletter.md
```diff
@@ -0,0 +1,221 @@
...
+    terms, any party to the terms can always do that.
+
+    For a quick look at anyprevout, we consider it in the context of a
+    two-party Eltoo-based LN channel.  As a reminder, the key feature of
+    Eltoo is that each balance change in a channel (state update) can be
+    spent by any later state update, so there's no risk of publishing an
```
jnewbery: Consider italicizing the _any_ for emphasis: "each balance change in a channel (state update) can be spent by _any_ later state update"

_posts/en/newsletters/2019-05-21-newsletter.md
```diff
@@ -0,0 +1,221 @@
...
+[bech32 easy]: {{news38}}#bech32-sending-support
+[mcf transcripts]: https://diyhpl.us/wiki/transcripts/magicalcryptoconference/2019/
+[nullc bank analogy]: https://www.reddit.com/r/Bitcoin/comments/9iw1p2/hey_guys_its_time_to_make_bech32_standard_on/e6onq8t/?context=3
+[over 200,000 bitcoins]: https://p2sh.info/dashboard/db/p2wpkh-statistics?orgId=1
+[news38 bech32]: {{news38}}#bech32-sending-support
+[news40 bech32]: {{news40}}#bech32-sending-support
+[v2 transport protocol]: https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2019-March/016806.html
+[anyprevout list]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-May/001992.html
+[mcf vids]: https://www.youtube.com/playlist?list=PLWqtMh0tDnEGBHDS8CalOpkVZhlIgRlAC
+[mcf andytoshi]: https://www.youtube.com/playlist?list=PLWqtMh0tDnEGBHDS8CalOpkVZhlIgRlAC
```
jnewbery: These all link to the playlist instead of the specific video/start time

_posts/en/newsletters/2019-05-21-newsletter.md
```diff
@@ -0,0 +1,221 @@
...
+(BIPs)][bips repo].*
+
+- [Bitcoin Core #15006][] extends the `createwallet` RPC with a new
+  `passphrase` parameter that allows creating a new wallet that's
+  encrypted by default.  Existing wallets can still be converted to
+  encrypted with the `encryptwallet` RPC.
+
+- [Bitcoin Core #15870][] will cause the `importwallet`,
+  `importpubkey`, `importaddress`, and `importprivkey` RPCs to fail if
+  any blocks have been pruned.  Previously, it failed if pruning was
```
jnewbery: I think in this case it'd be clearer to first say what the current behaviour is, and then what the change is. For example:

[Bitcoin Core #15870][] makes `importwallet`, `importpubkey`, `importaddress`, and `importprivkey` RPCs succeed more often for pruned nodes.  Previously, those calls would always fail on nodes which were configured to prune the blockchain. After this change, they will only fail if blocks have already been pruned. Pruning can be configured for manual operation...

_posts/en/newsletters/2019-05-21-newsletter.md
```diff
@@ -0,0 +1,221 @@
...
+includes our regular sections on bech32 sending support and notable
+changes to popular Bitcoin infrastructure projects.
+
+## Action items
+
+*None this week.*
+
+## News
+
+- **Proposed anyprevout sighash modes:** two weeks ago, Anthony Towns
```
jnewbery: AJ's mailing list post (https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2019-May/016929.html) adds some nice context to the proposed BIP, so I think it'd be nice to link to it from here.

_posts/en/newsletters/2019-05-21-newsletter.md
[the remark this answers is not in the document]
moneyball: Yeah I wasn't suggesting exclusively Gemini, and citing even more is better as it communicates how real this is becoming.

_posts/en/newsletters/2019-05-21-newsletter.md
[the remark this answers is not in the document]
moneyball: I actually was just curious as I wasn't aware of any wallet that ONLY accepts bech32. We can close this out.

_posts/en/newsletters/2019-05-21-newsletter.md
moneyball: Then, shouldn't we say "commits to other details about the prevout, such as its bitcoin value and pubkey."?

harding: I think that would confuse the issue, as we're discussing the bip-taproot/bip-tapscript signature hashing algorithm here (which does not directly commit to the pubkey), rather than the bip-schnorr verification protocol (which does require the signature commit to the pubkey).  I'm adding some extra words to maybe make that clearer.

harding:
Pushed edits that hopefully address @jnewbery's feedback and a follow-up comment from @moneyball.  (Thanks!)

_posts/en/newsletters/2019-05-21-newsletter.md
```diff
@@ -0,0 +1,225 @@
...
+    `SIGHASH_ANYPREVOUT` that also requires the same special pubkey
+    prefix and chaperone signature, but it includes the prevout's
+    scriptPubKey and tapleaf hash in the signature.  Whereas
+    anyprevoutanyscript can allow Eltoo-like rebinding where any later
+    state can bind to any earlier state (but earlier states can't bind
+    to later states), there may be alternative protocols (and times
+    within the Eltoo protocol) where the participants want to ensure
+    that state *n* can only bind to state *n-1* and not any other state.
+
+    The proposal has begun receiving [feedback][anyprevout list] on the
```
ajtowns: Could be worth linking directly to https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-May/001996.html which has more specific eltoo details.

jnewbery: I'm going to leave this link out and just retain the link to the first post in that thread. I'm sure there'll be more coverage of anyprevout in future newsletters where we can include links to interesting posts.

bitschmidty:
Its merged! 

Thanks @harding for authoring and @jnewbery @moneyball and @ajtowns for reviewing.

PR #146, opened 2019-05-22
Blog: add post with all bech32 content to date

### LINKED ISSUE #128: Add blog post summarizing 24 weeks of bech32
Once we complete the 24 weeks of bech32 in August, we should consider putting all of the content in a single post linked from our web site, so that it becomes an easy reference for future readers.

harding:
Closes #128 

This moves each bech32 sending support section into a separate includable file and then includes each of them in two different places: first, in the newsletter that originally included them; second, in a blog post backdated to the start of our series.  I think this makes the combined content highly accessible, as it's all easily searchable using find-in-page.

This required moving links around and also moving where we included references.md in some previous newsletters (because it contains variable definitions, it must appear before we use those variables, which means including it before the subsequently-included bech32 sections.  However, it can't go at the very top of the file, since that breaks the automatic first-paragraph summary used on the newsletter index page.)

A separate commit in this PR also adds a script that automates most of the work of creating the link reference definitions for old newsletters.  Manually creating those links was annoying for me, even if it only took 30 seconds a week.

When making a change like this that makes a lot of backend changes but minimal frontend changes, my preferred testing method is comparing diffs of the rendered site.  E.g.:

    git checkout 4c551cc0    ## before these changes
    make
    cp -a _site _old_site
    git checkout <this PR>
    make
    diff -ruN _old_site _site | colordiff | less -R

I think that looks good.  (You can drop the -N from diff to suppress new files so that you only see the changes to existing files, which is a minimal diff.)

### DIFF
--- a/_posts/en/2019-03-19-bech32-sending-support.md
+++ b/_posts/en/2019-03-19-bech32-sending-support.md
@@ -1,0 +1,87 @@
+---
+title: Bech32 Sending Support
+permalink: /en/bech32-sending-support/
+name: 2019-03-19-bech32-sending-support
+type: posts
+layout: post
+lang: en
+
+excerpt: >
+  Copies of all published parts of our 24-part weekly series on bech32 sending
+  support from March 19th to August 28th, 2019.
+
+---
+<style>
+/* put a little extra space between the H2s to maybe help
+ * readers understand each of these was originally published independently
+ * of the others */
+h2:not(:first-of-type) { margin-top: 3em; }
+</style>
+
+{% include references.md %}
+
+On this page are copies of all published parts of our 24-part weekly series
+on bech32 sending support, from March 19th to August 28th, 2019.
+
+1. toc
+{:toc}
+
+## Introduction to Bech32 Sending Support
+
+*Originally published in [Newsletter #38][].*
+
+{% include specials/bech32/01-intro.md %}
+
+## Bech32 usage statistics
+
+*Originally published in [Newsletter #39][].*
+
+{% include specials/bech32/02-stats.md %}
+
+## Using the bech32 reference libraries
+
+*Originally published in [Newsletter #40][].*
+
+{% include specials/bech32/03-python-ref.md %}
+
+## Locating typos in bech32 addresses
+
+*Originally published in [Newsletter #41][].*
+
+{% include specials/bech32/04-ecc.md %}
+
+## Fee savings with native segwit
+
+*Originally published in [Newsletter #42][].*
+
+{% include specials/bech32/05-fee-savings.md %}
+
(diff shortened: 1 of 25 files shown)


moneyball:
tACK. Note that I only built locally and tested - I didn't review the actual changes, so someone else should do that.

I'm glad we're doing this now so that we have something easily shareable. It is already an impressive compilation.

I assume we plan to append each week's section to this when we publish the weekly newsletter?

Also this seems to belong in our scaling book too ... maybe repurpose it there in August when all 24 are complete?

harding:
> I assume we plan to append each week's section to this when we publish the weekly newsletter?

Yep.  Sections from this point forward will be written in the `_includes` dir and then added to both that week's newsletter and this blog post.

_posts/en/2019-03-19-bech32-sending-support.md
```diff
@@ -0,0 +1,89 @@
...
+title: Bech32 Sending Support
+permalink: /en/bech32-sending-support/
+name: 2019-03-19-bech32-sending-support
+type: posts
+layout: post
+lang: en
+
+excerpt: >
+  Copies of all published parts from our weekly series on bech32 sending
+  support, scheduled to run 24 weeks from March 19th to August 28th,
```
jnewbery: If you s/scheduled to run/running for/ we won't need to update this once the series is complete.

_posts/en/2019-03-19-bech32-sending-support.md
```diff
@@ -0,0 +1,89 @@
...
+/* put a little extra space between the H2s to maybe help
+ * readers understand each of these was originally published independently
+ * of the others */
+h2:not(:first-of-type) { margin-top: 3em; }
+</style>
+
+{% include references.md %}
+
+On this page are copies of all published parts from our weekly series
+on bech32 sending support, scheduled to run 24 weeks from March 19th to
```
jnewbery: As above, consider s/scheduled to run/running for/.

_posts/en/2019-03-19-bech32-sending-support.md
```diff
@@ -0,0 +1,89 @@
...
+
+*Originally published in [Newsletter #46][].*
+
+{% include specials/bech32/09-qrcode.md %}
+
+## Bech32 support as a proxy for competence
+
+*Originally publised in [Newsletter #47][].*
+
+{% include specials/bech32/10-snooze-lose.md %}
```
jnewbery: snooze-lose :+1:

harding:
@jnewbery the sentence has been rephrased.  Thanks!

jnewbery:
ACK 983a2fbfe05d7f4bfe72d98a4033d916befe5a27. Thanks Dave!

PR #148, opened 2019-05-24
Correction: QR codes can use mixed nodes

harding:
As noted by @shesek here: https://twitter.com/shesek/status/1131733590235131905 (thanks again!)

I've given the main updated paragraph an anchor so that we can link directly to it in a Corrections section of next week's newsletter.

This PR also fixes a dangling footnote reference left over from the content move and adds a test so that future dangling footnote refs should cause a CI failure.


jnewbery:
utACK bc7f73f15d37f8aa07faa4b5a6dd73a029a4a9ac

shesek:
LGTM, thanks for the mention!

PR #149, opened 2019-05-28
News48 (2019-05-28ish)

harding:
- [ ] Fix FIXMEs.  These are mostly minor stuff that should be easy to review at the last minute and that are easy to imagine how it'll look or where a link will go just from the text surrounding it.
    - [x] I need the final URL for the video of @bitschmidty's talk (or Mike needs it for merge time tomorrow).  The relevant line has a FIXME in a comment (near the bottom of the document).

- Note that this is a very long newsletter.  I'm sorry about that.  Please don't hesitate to suggest the culling of paragraphs or entire subsections, and to recommend that I restrain myself in the future.

- I'm unhappy with several of the diagrams in this newsletters.  Before publication, I'll tighten them up a bit (remove whitespace so the font size is larger), but if anyone sees a way to materially improve them, please let me know.  Alternatively, let me know if you think the newsletter would be better without them.

- The newsletter deliberately doesn't cover Blummer's difficulty opcode proposed soft fork.  He said he was just soliciting feedback and would come back with an actual BIP.  Since we already have an over-stuffed newsletter (our longest ever), I think it's reasonable to punt on that.

- As usually, @moneyball's SE section is added in a separate commit and I have a fixup commit on top of that with my edits so that he can review.  I'll squash those together after his review.

- Finally, I do have a fear that I've terribly misunderstood COSHV and everything I wrote about it is rubbish.  If that seems to be the case, even partly, this newsletter is certainly long enough for us to just yank that whole section and update the news item to say "we'll cover it in the future."  Those of you with commit access, please feel free to do that without consulting with me (or make whatever other edits).

Thanks as always for reviewing my overly long and poorly illustrated ramblings!


moneyball:
ACK da97f749e70dcb3a3ef18a5bbe113786d586dc88

(but still need to consider @jnewbery feedback)

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,593 @@
...
+  something.
+
+- [What are the limitations for amortizing the interactive session setup
+  of MuSig?]({{bse}}87605) Richard Myers is attempting to optimize
+  interactive setup for a low bandwidth system, but user nickler
+  emphasizes that nonces must not be reused or private keys could
+  be leaked. Nickler goes on to provide suggestions to achieve
+  Myers’ goal.
+
+- [On chain cost of Segwit version 1 versus version 0?]({{bse}}87697)
```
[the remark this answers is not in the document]
moneyball: That's @harding's chart from our newsletter :) I'll update the summary to also mention the 2nd answer.

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,596 @@
...
+    transactions during fee events.
+
+    The 30-minute presentation covers each point concisely, making it an
+    excellent high-level overview for anyone interested in learning
+    about the Bitcoin fee market and how to mitigate expected fee
+    increases.
+
+## Proposed transaction output commitments
+
+The [proposed][bip-coshv] opcode `OP_CHECKOUTPUTSHASHVERY` allows a
```
jonatack: s/VERY/VERIFY/

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,596 @@
...
+would only require them to submit a Bitcoin address and an email address
+(or some other protocol address for delivery of the committed outputs).
+
+### More advanced usage
+
+The transaction size (and thus fee cost) of using COSHV in the previous
+subsection's example is basically the same as using the pre-signed
+transaction format with current Bitcoin transactions.  However, the
+ability to use Taproot to allow the receivers to choose from any one of
+up to several billion spending conditions adds a significant amount of
```
jonatack: that's... quite a few indeed

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,596 @@
...
+### Feedback and activation
+
+The proposal received over 30 replies on the Bitcoin-Dev mailing list as
+of this writing.  All of the replies to date are either criticism or
+replies to criticism.  That's to be expected---the rule for mailing
+lists, in order to avoid wasting time, is that every post must move the
+discussion forward.  There's no place for a bunch of people to say,
+"hey, that's a really cool idea!"  We suspect 30 replies in just a few
+days from some well known Bitcoin protocol developers is its own subtle
+complement: it means the proposal is interesting enough to warrant
```
jonatack: compliment?

jonatack:
Great read; shaping up to an excellent bitcoinops issue

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,593 @@
...
+
+        This week, John Newbery [suggested][smaller v1 spk] that we use some method to
+        avoid placing this byte in the scriptPubKey.  Wuille agreed that
+        this could be useful and will attempt implementing a variation
+        where the bit will be included as part of the taproot witness
+        data.  This will reduce the cost create a taproot output by 1.0
+        vbytes (making it the same as P2WSH currently) with either zero
+        or 0.25 vbytes additional cost to spend a taproot output.
+
+- FIXME maybe minisketch thing
```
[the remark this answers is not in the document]
ajtowns: Better to give things at least a day or two before a write up in general, in my opinion; better if the newsletter tries to be detailed and informative rather than catching breaking news. YMMV etc

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,597 @@
...
+The proposed soft fork would add a new opcode,
+`OP_CHECKOUTPUTSHASHVERIFY` (abbreviated by its author as `OP_COSHV`
+with an extra *S*).  This opcode and a hash digest could be included in
+tapleaf scripts, allowing it to be one of the conditions in a Taproot
+address.  When that address was spent, if COSHV was executed, the
+spending transaction would only be valid if the hash digest of its
+outputs matched the hash digest read from the script by COSHV.
+
+Comparing this to our example above, Alice would again ask each of the
+participants for a public key (such as a Taproot
```
ajtowns: This isn't necessary -- Alice an instead generate a NUMS point P (that can't be used for signing), generate the OP_COSHV script S, calculate the taproot point Q = P+H(P,S)*G, and pay to Q. To spend, you need to know the NUMS point P, the script structure for S and the outputs that are being paid to.

The advantage of having P as a musig point instead is the recipients can then change the arrangement later if they want.

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,597 @@
...
+this into a specific distribution transaction.  Instead, she'd just hash
+the ten outputs together and use the resultant digest to create a
+tapleaf script containing COSHV.  That would be the only tapleaf in this
+Taproot commitment.  Alice could also use the participants' public keys
+to form the taproot internal key to allow them to cooperatively spend
+the money without revealing the Taproot script path.
+
+![FIXME](/img/posts/2019-05-tx-tree-2.png)
+
+Alice could then give each of the receivers a copy of all ten outputs to
```
ajtowns: Alice has to give all the receivers a copy of all the outputs, otherwise they can't claim their specific output.

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,597 @@
...
+subsection's example is basically the same as using the pre-signed
+transaction format with current Bitcoin transactions.  However, the
+ability to use Taproot to allow the receivers to choose from any one of
+up to several billion spending conditions adds a significant amount of
+flexibility to the approach.
+
+In our previous examples, getting the distribution transaction confirmed
+was an all-or-nothing endeavor.  Either all ten receivers finally
+received their money at the same time, or all of them were still waiting
+for that confirmation to occur.  However, with a MAST tree and COSHV,
```
ajtowns: I think the approach Jeremy is thinking of is a tree of outputs, ie:

 * I want to pay to 125 people
 * I make 25 COSHV outputs of 5 payments each
 * I make 5 second level COSHV outputs, each grouping 5 of the 25 COSHV outputs I just generated
 * I make a transaction that pays to a COSHV commitment of those 5 COSHV outputs

If one of the 125 people wants to collect their funds, they publish three transactions: one that expands the single COSHV commitment into 5; another that expands one of those into another 5; and a final one that expands one of those into 5 actual payments, including their own.

I think the more interesting (and scalable) MAST approach is to allow increasing fees: if Alice is paying 3 people, create a MAST tree of:

     [1 Bob, 2 Carol, 1.5 Dave, 0.4 Alice] OP_COSHV
     100 OP_CSV [1 Bob, 2 Carol, 1.5 Dave, 0.3995 Alice] OP_COSHV
     2000 OP_CSV [1 Bob, 2 Carol, 1.5 Dave, 0.39 Alice] OP_COSHV

allowing the first path to be used immediately, but adding 0.0005 BTC in fees via the second path if there hasn't been a low fee period within 100 blocks, and adding 0.01 BTC in fees if there hasn't been a low enough fee period in two weeks.

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,597 @@
...
+in size.  Since the other participants will probably want to stay in the
+factory, it's likely the person exiting will be responsible for all
+transaction fees, not just the fees for their 149 channels (outputs).
+
+However, using a multi-transaction tree construction such as above, it's
+possible for any participant to close any individual one of the their
+channels onchain (or all of them at the same time, if the MAST tree
+includes branches for those outcomes).  Also, like the joinpool, it's
+possible for any participant to kick out any other participant (keeping
+all channels intact), allowing activate participants to remove inactive
```
ajtowns: "active" not "activate"?

JeremyRubin:
Had a chance to review the OP_COSHV parts!  A few notes:

1) Overall, great summary. Captured the heart of what it's doing!
2) The S in COSHV is really to emphasize that it is OutputS, as opposed to Output, which could be separately discussed as an alternative design
3) The n! blowup only happens if you want to allow anyone to exit at any level. And because each level requires a Merkle tree lookup, this ends up being N Log N  data! Uh-oh! It's much simpler if you just commit to a balanced tree of transaction data and have two options on the tapscript tree -- expand all or expand by two. Then you get something that is O(log(N)) worst case per individual, overall O(2N), and and E[O(1)] overhead if the expand all path is used -- we can learn optimal tree structures from real-world use over time.
4) We have some issues with malleability in the current design I didn't think about w.r.t. sequence and locktime and version. There is a simple fix for it, `OP_SECURETHEBAG` (which is like OP_COSHV, but you hash everything like a txid but the COutPoints values)

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,593 @@
...
+      the subset of scripts miniscript will create, this change in
+      semantics will at most save 0.25 vbytes per tapscript.  Also,
+      although the change may also simplify development for anyone
+      writing scripts by hand, it's a bit risky as every development
+      guide to Script written to date teaches that scripts must
+      terminate with a *true* value on the stack.  Wuille summarized,
+      "so overall this feels like something with marginal costs, but
+      also at most marginal benefits."
+
+    - *Move the oddness byte:* Bitcoin public keys are graph points, so
```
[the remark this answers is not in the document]
harding: @jnewbery 

> Note that the proposal is to place the y-oddness bit in the leading byte of the control block, which has unused bits, so this doesn't increase the size of the witness

I phrased it that way because I don't understand what you and sipa were talking about.  You and he mention the control block, but IIUC, the control block isn't there for key-path spends.  For script-path spends where the control block is present, I don't think you need the Y bit as there's no signature verification being done---you can compare just the X coordinate of the output pubkey to just the X coordinate of ec_add(tweak, internal_key).

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,593 @@
...
+This technique allows spenders and receivers to lock in a set of
+payments during high fees and then only distribute the actual payments when
+fees are lower.  According to Bitcoin Core fee estimates at the time of
+writing, anyone patient enough to wait a week for a transaction to
+confirm (like the distribution transaction above) can save 99% on fees.
+Let's look at the example above in that context:
+
+| | Individual Payments | Batched Payment | Commit now, distribute later |
+|-|-|-|-|
+| Immediate (high fee) transactions | 10x141 vbytes | 1x411 vbytes | 1x141 vbytes |
```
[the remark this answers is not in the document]
harding: @jnewbery 

> I think the 141 vbytes is assuming both outputs are P2WPKH. That's not the case for column three, where the txout to the multisig address is a P2WSH

If we're magically assuming key/sig aggregation (e.g. multiparty ecdsa), then P2WPKH is usable here.

@ajtowns

> Would be nice to provide the working for the vbyte calculations in an ipynb link or a onhover tooltip

But that would defeat the efficiency gains of me making stuff up randomly as I go.  :-)

_includes/specials/bech32/11-only-bech32.md
```diff
@@ -0,0 +1,68 @@
...
+  choosing a segwit wallet seed will only be able to generate bech32
+  addresses for receiving.  Electrum warns users about the compatibility
+  issues this may create with software and services that haven't
+  upgraded to bech32 sending support yet:
+
+    ![Dialog in Electrum allowing the user to choose the address type
+    and warning them that some services may not support bech32
+    addresses](/img/posts/2019-05-electrum-choose-wallet-type.png)
+
+    Please note that it's neither required nor recommended for wallet
```
jnewbery: This looks a bit weird directly below the Electrum image and talking about the Bitcoin Core wallet.

_includes/specials/bech32/11-only-bech32.md
```diff
@@ -0,0 +1,68 @@
...
+their preferred wallet.
+
+[bech32 easy]: {{news38}}#bech32-sending-support
+[wasabi bech32 only]: https://github.com/zkSNACKs/WalletWasabi/blob/master/WalletWasabi.Documentation/FAQ.md#my-wallet-cant-send-to-bech32-addresses---what-wallets-can-i-use-instead
+["only generate filters regarding bech32 addresses"]: https://github.com/zkSNACKs/Meta/blob/master/README.md#wasabi-wallet-under-the-hood
+[wasabi wallet]: https://wasabiwallet.io/
+[trust wallet]: https://trustwallet.com/
+[electrum]: https://electrum.org/
+[news45 bech32]: {{news45}}#bech32-sending-support
+[ivgi tweet]: https://twitter.com/shesek/status/1131733590235131905
```
jnewbery: Should `[ivgi tweet]`, `[updated qr paragraph]`, `[optech twitter]` and `[optech contributors]` be in the newsletter markdown file instead of here?

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,597 @@
...
+
+This technique allows spenders and receivers to lock in a set of
+payments during high fees and then only distribute the actual payments when
+fees are lower.  According to Bitcoin Core fee estimates at the time of
+writing, anyone patient enough to wait a week for a transaction to
+confirm (like the distribution transaction above) can save 99% on fees.
+Let's look at the example above in that context.  To make later
+comparisons to Taproot more fair, we'll assume some form of key and
+signature aggregation is being used, such as [MuSig][] or (in theory)
+multiparty ECDSA (see [Newsletter #18][]).
```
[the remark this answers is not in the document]
harding: @ajtowns

> You could do a more generic analysis

I think that'd be useful in documentation that recommends commited-forward-batching (or whatever you call this technique).  Obviously, having a formula would also make it nice to provide a graph showing how much this saves per outputs to be created.

Here, I think using a single specific example with a story attached makes the most sense as it requires the least amount of mental effort on the part of the reader to skim.  You and John actually went through the table looking for errors because you want this newsletter to be accurate (thank you!), but I expect normal readers to just skip down to the bottom line (that's why it's in bold) and the conclusion sentence after it.  That's good, because all I want the reader to think at this point is "this is a useful technique for saving money; now show me how COSHV makes this better".

_posts/en/newsletters/2019-05-28-newsletter.md
harding: @ajtowns

> This [Alice asking participants for their pubkeys] isn't necessary

It is, because that pubkey (as a Taproot address) is where the output will ultimately be sent.  I say "pubkey" here instead of "address" so that this looks almost identical to the example in the previous section using current script.  There's a parenthetical here saying the pubkey is the taproot address and a linked footnote about why you shouldn't assume taproot addresses are necessarily pubkeys.

> The advantage of having P as a musig point instead is the recipients can then change the arrangement later if they want.

Since we're going to use that feature in later subsections, I think it's best that we introduce Alice here as using musig to create the taproot internal key.  I think it's always better for her to do so anyway.

harding:
@JeremyRubin thanks for reviewing!  I'll be pushing edits soon that implement the change from a MAST tree to an output tree; I think your design there is not only better but it also makes that section of the article much easier to read.

I'm unable to see where the malleability concerns come from, unless it's some sort of future-proof thing.  It seems to me that the only thing you can do with tx version now is disable BIP68 consensus-enforced sequence, but if a tapscript is using BIP112 OP_CSV, it should fail if tx.version <2, right?  For BIP68 sequence and nLockTime in other contexts, if anyone creates a transaction that can't be included in a block, that doesn't seem like an issue for COSHV because one of the other people with a copy of the transaction tree can just create an alternative transaction that can be included.

I'm sure I'm missing something obvious.  If you could give me a hint, that'd be appreciated.  We're pretty close to our publication deadline now and I don't want to try to describe something that I don't understand, so I think I'm going to omit that for now.  We'll be able to cover it in a later newsletter when you make your proposal for it.

harding:
I've gone through all the feedback (thanks!) and fixed all the nits and addressed most of the other comments.  I've also fixed all the fixmes.  That's all been pushed.  If I get hit by a bus in the next couple hours (while sitting in my apartment), I think this should be publishable.

I'm going to keep working on some final edits and maybe image optimizations.  I'll try to have that all done by 10:00 EDT (which is late, I know, sorry).

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,631 @@
...
+
+Alice wants to pay a set of ten people but transaction fees are
+currently high so that she doesn't want to send ten separate
+transactions or even use payment batching to send one transaction
+that includes an output for each of the receivers.  Instead, she wants
+to trustlessly commit to paying them in the future without having to pay
+onchain fees for ten outputs now.  So Alice asks each of the receivers
+for one of their public keys and creates an unsigned and unbroadcast *setup* transaction
+that pays those keys using a 10-of-10 multisig script.  Then she creates
+a child transaction transaction that spends from the multisig output to the
```
jnewbery: "transaction transaction"

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,631 @@
...
+![A setup transaction paying a pre-signed distribution transaction](/img/posts/2019-05-tx-tree-1.png)
+
+When the setup transaction receives a reasonable number of
+confirmations, there's no way for Alice to take back her payment to the
+10 receivers.  As long as each of the receivers has a copy of the
+distribution transaction and all the others' signatures, there's also no
+way for any receivers to cheat any other receiver out of a payment.  So
+even though the distribution transaction that actually pays the
+receivers hasn't been broadcast or confirmed, the payments have the same
+amount of security as the confirmed setup transaction.  At any time, any
```
jnewbery: 'the same amount of security' seems difficult to quantify. Yes, it's the same number of confirmations that need to be unwound to double spend, but the security model does seem different somehow, in that the funds can be held up until one of the recipients is prepared to pay the fee.

I'd consider rewording "the same amount of security" since it's a bit vague.

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,631 @@
...
+of the receivers who wants to spend their money can broadcast
+the distribution transaction and wait for it to confirm (perhaps helping
+to pay its fee using child-pays-for-parent fee bumping).
```
jnewbery: This is a bit tricky. We don't have package relay yet, so if the distribution transaction doesn't have enough fees to get into the mempool, then there's no way to bump it. That's an extreme case - most of the time I expect the distribution transaction's fee to be somewhere between the mempool minimum and the rate required to be mined. It might be worth adding that in the case of extreme fee spikes, the distribution transaction would be stranded until mempools emptied enough to accept it.

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,631 @@
...
+signature aggregation is being used, such as [MuSig][] or (in theory)
+multiparty ECDSA (see [Newsletter #18][]).
+
+{% assign p2wpkh = "A transaction spending one P2WPKH input to two P2WPKH outputs" %}
+{% assign batched11 = "A transaction spending one P2WPKH input to eleven P2WPKH outputs, or 9*(8+1+22) more bytes than the two-output P2WPKH transaction" %}
+{% assign batched10 = "A transaction spending one P2WPKH input to ten P2WPKH outputs, or 8*(8+1+22) more bytes than the two-output P2WPKH transaction" %}
+
+| | Individual Payments | Batched Payment | Commit now, distribute later |
+|-|-|-|-|
+| Immediate (high fee) transactions | 10x<abbr title="{{p2wpkh}}">141 vbytes</abbr> | 1x<abbr title="{{batched11}}">420 vbytes</abbr> | 1x<abbr title="{{p2wpkh}}">141 vbytes</abbr> |
```
jnewbery: The tooltips are great, but it's not immediately intuitive that they exist. Consider adding a ℹ️symbol or a note at the bottom _hover over fees in table for calculations_

_posts/en/newsletters/2019-05-28-newsletter.md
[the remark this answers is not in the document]
ajtowns: For key path spends, you just always treat the key as if it were even (so imply an 0x02 prefix). If your secret key generated an odd key (0x03 prefix), just negate it to get the secret key for the even variant.

For script path spends you need the oddness in order to batch verification, because you need to verify Q = P+H(P,S)*G but if you're guessing Q's parity, then you might actually be plugging (-Q) = P+H(P,S)*G into the verifier, in which case you'll get a failure when you should be getting a success.

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -0,0 +1,597 @@
...
+was an all-or-nothing endeavor.  Either all ten receivers finally
+received their money at the same time, or all of them were still waiting
+for that confirmation to occur.  However, with a MAST tree and COSHV,
+it's possible to allow each receiver to individually withdraw their
+money from the covenant while also allowing all the other receivers to
+keep their money offchain.  To avoid combinatorial blow up, we'll
+explain this using an example of just four receivers instead of ten.
+
+Bob, Charlie, and Danielle, and Edmond receive a payment to a Taproot
+address that contains multiple tapleafs using the COSHV opcode.  Each
```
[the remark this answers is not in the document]
ajtowns: btw, what happened to Carol? did she retire? :(

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,631 @@
...
+amount of security as the confirmed setup transaction.  At any time, any
+of the receivers who wants to spend their money can broadcast
+the distribution transaction and wait for it to confirm (perhaps helping
+to pay its fee using child-pays-for-parent fee bumping).
+
+This technique allows spenders and receivers to lock in a set of
+payments during high fees and then only distribute the actual payments when
+fees are lower.  According to Bitcoin Core fee estimates at the time of
+writing, anyone patient enough to wait a week for a transaction to
+confirm (like the distribution transaction above) can save 99% on fees.
```
jnewbery: It's slightly confusing that this says "save 99% on fees" but the paragraph below says "save 90%". Perhaps rephrase this to say that "anyone patient enough to wait a week for a transaction to
confirm (like the distribution transaction above) can set the feerate to 1% of that required for a fast confirmation" or similar, to make it clear that you're talking about feerate estimates here, and total fees saved in the paragraph below.

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,631 @@
...
+one of the taproot leaves at the top level also allowed just sending the
+regular batched payment transaction---allowing everyone to exit together
+efficiently.
+
+This balanced tree construction provides a good generic starting part for
+allowing individuals to exit from the covenant without having to pay all
+the exit costs themselves.  When applying COSHV to specific cases where
+it's known some exits are more likely than others, it's possible to
+use unbalanced trees to make the more likely outcomes less expensive.
+We expect that is COSHV is adopted and sees regular use, tools such
```
jnewbery: s/is COSHV is/if COSHV is/

jnewbery: also consider changing this to say "if COSHV becomes part of the Bitcoin protocol and sees regular use" or similar to emphasise that this is still a proposal

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,631 @@
...
+as [miniscript][] will be adopted so that they take a set of desired
```
jnewbery: s/adopted/adapted/ ?

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,631 @@
...
+outcomes plus likeliness metadata and use that to produce optimized
+covenants.
+
+However, even with the simple example above we again see the key benefit
+of COSHV is its ability to allow Alice to construct of this complex
```
jnewbery: s/to construct of/to construct/

_posts/en/newsletters/2019-05-28-newsletter.md
```diff
@@ -229,73 +229,69 @@ would only require them to submit a Bitcoin address and an email address
...
+would be about double the number of participants.  However, this could
+be just the one output per participant from the previous subsection if
+one of the taproot leaves at the top level also allowed just sending the
+regular batched payment transaction---allowing everyone to exit together
+efficiently.
+
+This balanced tree construction provides a good generic starting part for
+allowing individuals to exit from the covenant without having to pay all
+the exit costs themselves.  When applying COSHV to specific cases where
+it's known some exits are more likely than others, it's possible to
```
ajtowns: With COSHV all exits are certain (p=1) so none are more likely than others; I think the difference is some are more _urgent_ than others. The outcome's the same -- since they're more urgent they're more likely to hit during high fees, so a shorter path saves fees.

_posts/en/newsletters/2019-05-28-newsletter.md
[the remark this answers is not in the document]
harding: Is there an semi-official list of names to use?  I always use the same names out to E, but after that I make it up ad hoc (excluding special cases like Eve for eavesdroppers and Mallory for malicious actors)

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,631 @@
...
+Starting this week, the Optech newsletter will be published every
+Wednesday instead of every Tuesday.  This will give us an extra weekday
+to review and edit newsletter drafts before we publish.
+
+## Special thanks
+
+We thank Jeremy Rubin and Anthony Towns for their reviews of a draft of
+this newsletter, including describing to us the tree of outputs idea.
+Any errors in the published version of this newsletter are the fault of
+the author.
```
jnewbery: Also thank sipa for his input on key aggregation?

_posts/en/newsletters/2019-05-29-newsletter.md
harding: > Consider adding a information_sourcesymbol or a note at the bottom hover over fees in table for calculations

I agree it's hard to see that the tooltips exist, but I don't really want to draw attention to them.  This is supposed to be a rough comparison table that covers just a single example to illustrate that there's significant savings to be had.  Like I said to AJ, if we formally write up this idea as a way to save money, that would be the place to focus on providing formulas and stuff.  Additionally, tooltips are not visible on touchscreen devices, so adding an icon here would be confusing for anyone reading on mobile (yeah, I could hide the icon on mobile or use JS to make the tooltip visible there or something).

jnewbery: ok. Sounds good!

jnewbery:
I've reviewed all the recent commits and it looks good to me.

ACK 9a47fc56eddc107ec3de80bbe4b521336f88175b

jnewbery:
ACK 84a3d966612b3ec42ba8f08a1a500ab8b67a0c5a

bitschmidty:
tACK 84a3d96

JeremyRubin:
I'm going to be annoying and leave a few more comments; overall it looks like an excellent presentation of the concept but a could have a few things a bit better; take em or leave em.

harding:
@JeremyRubin please do!

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,566 @@
...
+
+This technique allows spenders and receivers to lock in a set of
+payments during high fees and then only distribute the actual payments when
+fees are lower.  According to Bitcoin Core fee estimates at the time of
+writing, anyone patient enough to wait a week for a transaction to
+confirm (like the distribution transaction above) can save significantly on fees.
+Let's look at the example above in that context.  To make later
+comparisons to Taproot more fair, we'll assume some form of key and
+signature aggregation is being used, such as [MuSig][] or (in theory)
+multiparty ECDSA (see [Newsletter #18][]).
```
JeremyRubin: https://github.com/JeremyRubin/lazuli/blob/master/doc/protocol.pdf is my earlier citation on doing it with ECDSA and shows how messy it is

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,566 @@
...
+it's possible to allow each receiver to individually withdraw their
+money from the covenant while also allowing most the other receivers to
+keep their money offchain for the moment.
+
+For example, let's imagine Alice paid each of our ten receivers 1 BTC
+using a tree of outputs where each output paid two or three other outputs until
+an individual's money was withdrawn from the covenant.  A complete tree
+might for ten participants might look like this:
+
+![Example output tree for ten participants](/img/posts/2019-05-tx-tree-5.png)
```
JeremyRubin: It's not clear to me that you want to have a 'bushy tree' which pays out along a path rather than a simplest possible 8 or 16 node binary tree where payouts only happen on the leaf nodes. But you can, so it's not inaccurate. There's a trade off around free-riding and other things.

But I'm OK with is as the example of how to do it.

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,566 @@
...
+
+Since Bob is the first participant to withdraw, he has to create more
+outputs than everyone else.
+Anyone else exiting after Bob would need to create even fewer outputs as
+they could just start spending from one of the intermediate COSHV
+outputs that Bob had caused to be confirmed onchain.  Some people (like Charlie in the above
+example) would get their exit outputs as a byproduct of another person's
+exit.  Overall, the total number of outputs that need to be created
+would be greater than the number of participants.  However, if nobody
+tried to exit early, a Taproot leaf at the top level could optionally
```
JeremyRubin: At every subtree

JeremyRubin: or every so often.

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,566 @@
...
+of COSHV is its ability to allow Alice to construct this complex
+multi-transaction tree without any interaction from the receivers after
+she has each participant's own desired withdrawal address.  It's
+possible to construct the same tree using just the pre-signed
+transactions we saw in the first sub-section above, but each of the
+possible output sets would be a separate transaction that would need to
+be interactively signed by all participants before the setup transaction
+could be broadcast.  For larger numbers of participants, this could
+easily require the signing of many thousands of transactions by each
+participant.
```
JeremyRubin: Also each participant only needs to verify their own path's commitments, other participants subtrees are not needed, which should improve privacy.

_posts/en/newsletters/2019-05-29-newsletter.md
```diff
@@ -0,0 +1,566 @@
...
+that look identical to single-user, single-sig transactions and so would
+not directly reveal that they belonged to the pool.  Even if it was
+known that a joinpool controlled those funds, third parties would not
+know who within the group had spent or received money.  Exit
+transactions might reveal that someone was leaving a joinpool but, if
+the group was involved in enough spending and receiving, it could be
+very difficult for a third party to use the transaction history graph to
+determine which member of the pool was exiting.
+
+Within the pool, it would unfortunately be the case that all members would know each others' pool
```
JeremyRubin: This isn't a necessity, you don't need to know the subtrees!

harding: Ah, that's fair.  With the prior MAST construction, you did need to know; with output trees, you only need to know the path to your output (which leaks some data from other people's values in aggregate, but not necessarily anything directly).

JeremyRubin:
Also the channel factories part was good before, but maybe a bit too much content :)

jnewbery:
ACK 8c3967226d0fdab46bc69bb993dcef6ef1da2e47

bitschmidty:
tACK https://github.com/bitcoinops/bitcoinops.github.io/commit/8c3967226d0fdab46bc69bb993dcef6ef1da2e47

bitschmidty:
Big pat on the back for @harding for authoring this beast of a newsletter and for reviews from @JeremyRubin @jonatack @ajtowns @jnewbery and @moneyball w/ input from @sipa

JeremyRubin:
> I'm unable to see where the malleability concerns come from, unless it's some sort of future-proof thing. It seems to me that the only thing you can do with tx version now is disable BIP68 consensus-enforced sequence, but if a tapscript is using BIP112 OP_CSV, it should fail if tx.version <2, right? For BIP68 sequence and nLockTime in other contexts, if anyone creates a transaction that can't be included in a block, that doesn't seem like an issue for COSHV because one of the other people with a copy of the transaction tree can just create an alternative transaction that can be included.

If only I understood it better myself I could explain it -- I find OP_CSV and OP_CLTV to be something I can never load into my head how they work (appreciate any references).


Essentially the issue is best thought of in terms of channel factories. I have a txn A which has an output A.vout[0] which is OP_COSHV to create B.vout[0], which is a payment channel. 

When I create txn B, neither the version nor the locktime is committed to in A.vout[0]'s OP_COSHV, so they can be set to any value. Given that, setting them can change the TXID which can break the payments made in the channel.

Does that make sense?

harding:
@JeremyRubin thanks!  That payment channels problem does make sense.  I'll try to work in a note about that when we cover the other COSHV usages in next week's newsletter.

JeremyRubin:
Yeah I'll try to put out an amended BIP by then

PR #156, opened 2019-06-13
Scripted diff: add 'slug:' front matter to all posts

### LINKED ISSUE #155: RSS Feed duplicate "id" fields
The RSS feed being generated is not generating unique id fields for each post entry. Each month has its own "id"s which each post of that month seems to get, instead of a unique id.

For example, for June 2019, all posts are getting this id:

`<id>/en/newsletters/2019/06/newsletter</id>`

Instead of the expected:

`<id>/en/newsletters/2019/06/05</id>`
`<id>/en/newsletters/2019/06/12</id>`

One example of how this manifests to a feed reader (newsblur), is that only one post per each month is shown in the feed:
Screen Shot 2019-06-13 at 12 44 22 PM

harding:
Closes #155 

It looks like Jekyll's slug routine isn't handling our URLs well, I don't know why.  As seen [here](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/document.rb#L385), the slug is used for part of the supposed-to-be unique id given to each page for things like Atom feeds, leading to the problem described in #155.

We aren't currently using page slugs for anything else and it's possible to set them manually, so this one line scripted diff just gives each page its own unique slug to ensure our feed contains entirely unique IDs.  There are probably better ways to do this, but I think this is acceptable for now.

### DIFF
--- a/_posts/en/2018-07-20-announcing-bitcoin-optech.md
+++ b/_posts/en/2018-07-20-announcing-bitcoin-optech.md
@@ -2,6 +2,7 @@
 title: 'Announcing Bitcoin Optech'
 permalink: /en/announcing-bitcoin-optech/
 name: 2018-07-20-announcing-bitcoin-optech
+slug: 2018-07-20-announcing-bitcoin-optech
 type: posts
 layout: post
 lang: en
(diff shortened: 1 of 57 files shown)


jnewbery:
ACK 37adedb6bb664dd801fe88a1f8743fe5dc881571

Thanks @harding

bitschmidty:
tACK 37adedb6bb664dd801fe88a1f8743fe5dc881571

Nice workaround @harding

PR #162, opened 2019-07-01
Posts: add missing slugs and test for their prescence

harding:
I've forgotten to add slugs to several posts, breaking the fix added in #156.  This PR adds the missing slugs plus a test so that it shouldn't be possible to omit them accidentally again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,15 @@ test:
 	## Check for Markdown formatting problems
 	@ ## - MD009: trailing spaces (can lead to extraneous <br> tags
 	bundle exec mdl -g -r MD009 .
+
 	## Check for broken Markdown reference-style links that are displayed in text unchanged, e.g. [broken][broken link]
 	! find _site/ -name '*.html' | xargs grep ']\[' | grep -v skip-test | grep .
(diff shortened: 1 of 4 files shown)


bitschmidty:
@harding 

> ! git grep -L "^slug: " _posts

opens an empty file in the case of all posts passing that validation. I wish I had better bash-fu to suggest a fix

jnewbery:
4770b1443882ee29c8e14dcccbe64c62fb126355 works for me.

https://github.com/bitcoinops/bitcoinops.github.io/pull/162#issuecomment-507380288 needs addressing before merge.

harding:
I forced pushed a commit that adjusts the git command line to eliminate the problem @bitschmidty found.

bitschmidty:
tACK c1fa11976730ec6f2509176d24dd1ed377e950e1

jnewbery:
tACK c1fa11976730ec6f2509176d24dd1ed377e950e1

Tested:
- `make test` fails when the slug is removed from a post
- `make test` fails when two posts have identical slugs

Thanks for this, @harding !

PR #163, opened 2019-07-06
Newsletters: add 54 (2019-07-10)

harding:


moneyball:
utACK

practicalswift:
utACK 31cd0694a76af224b8907355152a1bcfd30a6e5f

I like the full disclosure part about lack of message signing support. Integrity is good.

_posts/en/newsletters/2019-07-10-newsletter.md
```diff
@@ -0,0 +1,116 @@
...
+  routed; otherwise, it would be rejected like normal.  Because the
+  routed payment might fail for other reasons and prevent the routing
+  node from earning any fees, any JIT rebalance operations need to be
+  free or they could end up costing the node money in a way that
+  attackers could exploit.
+
+    In a new [post][zmn jit] to the Lightning-Dev mailing list,
+    pseudonymous LN developer ZmnSCPxj described two situations in which
+    other profit-maximizing nodes might allow free rebalances.  The
+    first case is the observation that the person at the other end of
```
jnewbery: Is "node being routed through" or "downstream routing node" clearer than "the person at the other end of the outbound routing channel"?

_posts/en/newsletters/2019-07-10-newsletter.md
```diff
@@ -0,0 +1,116 @@
...
+    the outbound routing channel will receive their own routing fee paid
+    by the spender if the payment succeeds.  ZmnSCPxj describes a method
+    by which they can make their part of the rebalance contingent on
+    them receiving their routing income, ensuring that they either get
+    paid or the rebalance doesn't happen.  This would require additional
+    communication between nodes and so it's a change that probably needs
+    further discussion in order to be considered for addition to the LN
+    specification.
+
+    The second case ZmnSCPxj describes is other people along the
```
jnewbery: nit: s/people/nodes

_posts/en/newsletters/2019-07-10-newsletter.md
```diff
@@ -0,0 +1,116 @@
...
+{% include specials/bech32/17-signmessage.md %}
+
+## Notable code and documentation changes
+
+*Notable changes this week in [Bitcoin Core][bitcoin core repo],
+[LND][lnd repo], [C-Lightning][c-lightning repo], [Eclair][eclair repo],
+[libsecp256k1][libsecp256k1 repo], and [Bitcoin Improvement Proposals
+(BIPs)][bips repo].*
+
+- Bitcoin Core [#15427][Bitcoin Core #15427] and [#16326][bitcoin core
```
jnewbery: I don't think you need to mention 16326 here. That was just tidying up some loose ends from the main PR.

_posts/en/newsletters/2019-07-10-newsletter.md
```diff
@@ -0,0 +1,116 @@
...
+  new feature is especially useful for hardware wallets and other paired
+  wallets as it makes it possible to add HD key-path information to the
+  PSBTs so that wallets asked to sign a PSBT can easily derive the keys
+  needed for signing or verify that a change output does indeed pay back
+  into the wallet.
+
+- [Bitcoin Core #16257][] aborts adding funds to an unsigned transaction
+  if its feerate is above the maximum amount set by the `maxtxfee`
+  configuration option (default: 0.1 BTC).  Previously, the funding code
+  would silently reduce the fee to the maximum, which merely limited the
```
jnewbery: I don't think this captures how dangerous the previous behaviour was. 'merely limiting' the fees to ~$1200 (at time of writing) could be pretty catastrophic. Suggest this is changed to "...reduce the fee to the maximum, which could result in extremely high fees being paid accidentally."

_includes/references.md
```diff
@@ -31,6 +31,7 @@
 {% comment %}<!-- TODO: switch miniscript link to some sort of overview page when available -->{% endcomment %}
 [miniscript]: http://bitcoin.sipa.be/miniscript/miniscript.html
 [musig]: https://eprint.iacr.org/2018/068
+[output script descriptor]: https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md
```
jnewbery: Is this alias reference required? Why not use the [output script descriptors] link below?

_posts/en/2019-03-19-bech32-sending-support.md
```diff
@@ -123,5 +123,11 @@ on bech32 sending support, from March 19th to August 28th, 2019.
 
 {% include specials/bech32/16-checklist.md %}
 
+## Message signing support
+
+*Originally published in [Newsletter #54][].*
+
+{% include specials/bech32/17-signmessage.md %}
```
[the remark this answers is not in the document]
harding: If there are significant changes to the status quo, we'll probably mention them in the newsletter (in the News section if the bech32 section is done), so we can just link to those in a note at the top of this section.  E.g. "Edit: subsequent to the writing of this section, blah blah was widely adopted by native segwit wallets.  See [Newsletter #123][] for details."

_posts/en/newsletters/2019-07-10-newsletter.md
```diff
@@ -0,0 +1,116 @@
...
+  sometimes LN nodes receive a routed payment that they reject because
+  their outbound channel for that payment doesn't currently have a high
+  enough balance to support that payment.  Rene Pickhardt previously
+  [proposed][pickhardt jit] Just-In-Time (JIT) routing where the node
+  would attempt to move funds into that channel from one or more of its
+  other channel balances.  If successful, the payment could then be
+  routed; otherwise, it would be rejected like normal.  Because the
+  routed payment might fail for other reasons and prevent the routing
+  node from earning any fees, any JIT rebalance operations need to be
+  free or they could end up costing the node money in a way that
```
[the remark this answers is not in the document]
harding: I'm not sure what the nit is here, are you saying that you understand the text but think it could be clearer or that you think it might be wrong but you're not certain?

_posts/en/newsletters/2019-07-10-newsletter.md
harding: I think there were more loose ends to 15427 than there should've been, so the mention of 16326 seems appropriate.  I prefer to keep it, but will respect your call if you want it dropped.

_includes/references.md
harding: Definitely not required, but we've had to use the other link several times already for the singular case so it seemed appropriate to just alias it:

```text
$ git grep 'output script descriptor]'
_posts/en/newsletters/2018-12-04-newsletter.md:  [output script descriptor][output script descriptors] for each address
_posts/en/newsletters/2019-02-12-newsletter.md:  specified using an [output script descriptor][output script
_posts/en/newsletters/2019-06-19-newsletter.md:  specific private keys using an [output script descriptor][output
```

I'm happy to add another case of `[output script descriptor][output script descriptors]` if you think that's cleaner.

_posts/en/newsletters/2019-07-10-newsletter.md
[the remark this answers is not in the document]
jnewbery: Sorry, the nit is that JIT rebalance operations do not need to be free. They just need to be lower fees than the original routing fees.

_posts/en/newsletters/2019-07-10-newsletter.md
jnewbery: I'm happy either way. I just think readers who are interested in learning more about the changes to `utxoupdatepsbt` won't gain much from reading 16326 since it was a bug fix for what should have been included in the original PR.

_includes/references.md
jnewbery: Got it. Let's keep the alias. Is there a way to define the alias in reference to the original [output script descriptors] reference, rather than duplicate the URL?

_posts/en/newsletters/2019-07-10-newsletter.md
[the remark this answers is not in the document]
harding: I believe that's incorrect, and I thought this sentence described why they need to be free.  I'll try to make it more clear, but the idea is that the rebalance operation needs to succeed before the routing operation can start, i.e. it's not one atomic operation.  So, if the payment being routed fails, the person routing ends up paying any rebalance fees without receiving any compensating routing income.

This can be exploited by one of your channel peers sending a payment through your node that they know will ultimately fail and that they strongly suspect will cause you to rebalance.  Someone who has two or more channels open to you can even use this to steal from you by sending a series of payments (that ultimately fail) that cause you to rebalance between your various channels with them, paying them some amount of fees each time.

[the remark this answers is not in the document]
jnewbery: You're correct. The sentence _is_ clear and I'm just being slow. Keep it as it is!

harding:
Pushed a commit that should address all of @jnewbery's feedback (thanks!).

_posts/en/newsletters/2019-07-10-newsletter.md
```diff
@@ -0,0 +1,116 @@
...
+{% include references.md %}
+
+## Action items
+
+*None this week.*
+
+## News
+
+- **Eclair 0.3.1 released:** the newest version of this LN software
+  includes new and improved API calls as well other changes that allow
```
bitschmidty: > as well other changes

to 

> as well as other changes

harding: Fixed, thanks!

jnewbery:
ACK efde31dad75b254a95dbcfd6489a7cff6cb7be68

Thanks @harding . Great newsletter as always!

_includes/specials/bech32/17-signmessage.md
```diff
@@ -0,0 +1,68 @@
...
+    error message:
+    Address does not refer to key
+
+Some wallets do support message signing for segwit addresses---but in a
+non-standardized way.  For example, the [Trezor][trezor segwit
+signmessage] and [Electrum][electrum segwit signmessage] wallets each
+provide message signing support for P2WPKH and P2SH-wrapped P2WPKH
+addresses.  Yet both implementations were made independently and use
+slightly different protocols, so they're [unable to verify][trezor
+electrum incompatible] signatures made by the other system.
```
jonatack: nit: perhaps s/made/generated/ here, or s/made/developed/ 2 lines above, to avoid using the same verb twice in the same sentence with somewhat different meanings.

harding: Good suggestion, thanks!  Pushed s/made/produced/ for the second occurrence of the verb.

_includes/specials/bech32/17-signmessage.md
```diff
@@ -0,0 +1,68 @@
...
+Additionally, the algorithms used by all wallets we're aware of can't be
+easily adapted to P2SH and P2WSH scripts used for multisig and other
+advanced encumbrances.  That means message signing today is universally
```
jonatack: nit: I'm not too sure what "encumbrances" means here. Could be wrong but I thought of it has having a negative connotation similar to "burden". Was the intent something more along the line of "accoutrements"...?  :D

_posts/en/newsletters/2019-07-10-newsletter.md
```diff
@@ -0,0 +1,116 @@
...
+  would attempt to move funds into that channel from one or more of its
+  other channel balances.  If successful, the payment could then be
+  routed; otherwise, it would be rejected like normal.  Because the
+  routed payment might fail for other reasons and prevent the routing
+  node from earning any fees, any JIT rebalance operations need to be
+  free or they could end up costing the node money in a way that
+  attackers could exploit.
+
+    In a new [post][zmn jit] to the Lightning-Dev mailing list,
+    pseudonymous LN developer ZmnSCPxj described two situations in which
```
jonatack: Here "described" is past tense but further in the same paragraph and the next one it is used in the present tense.

jonatack:
(Great newsletter as usual! Will try to review a bit earlier; I suspect you are about ready to publish)

_includes/specials/bech32/17-signmessage.md
harding: Yeah, it's not a well known word.  It does mean the same basic thing as burden, but it's Bitcoin usage comes from its legal usage as a the jargon used to describe what terms need to be fulfilled in the transfer of property.  For Bitcoin, the property being transferred is, of course, bitcoins and the conditions are the terms in the scriptPubKey (either directly or by hash reference).  E.g. see the opening paragraphs of my article here: https://bitcointechtalk.com/what-is-a-bitcoin-merklized-abstract-syntax-tree-mast-33fdf2da5e2f?gi=52eef0b5b732

I did consider explaining this more simply, but this section is already a bit hard to parse due to the number of concepts it needs to introduce, so I think it's better that we keep it short and let people who don't understand the term just guess at its meaning and continue reading to see if the rest of the info in the section matters to them.  If it does, they can go look up encumbrance.

_posts/en/newsletters/2019-07-10-newsletter.md
harding: Fixed and pushed.  Thanks again!

_includes/specials/bech32/17-signmessage.md
jonatack: Thanks for the contextual explanation! TIL :)

jonatack: (perhaps "conditions" or "features" would be more clear to laypeople but admittedly may be less precise or have somewhat different meanings)

bitschmidty:
tACK 3179e06

and squashed

bitschmidty:
Thanks @harding, you’ve done it again 
The next newsletter? Wednesdays when

Thank yous go out to John and Jon
@jnewbery and @jonatack for gettin their review on

PR #166, opened 2019-07-16
Docs: add style guide

### LINKED ISSUE #165: House style guide
There are certain conventions that we've adopted in our newsletters over the last year. For example:

- onchain over on-chain (https://github.com/bitcoinops/bitcoinops.github.io/pull/86#discussion_r234697525)
- light client over lite client (https://github.com/bitcoinops/bitcoinops.github.io/pull/105#discussion_r251551595)
- merkle tree over Merkle tree (https://github.com/bitcoinops/bitcoinops.github.io/pull/113#discussion_r261849651)
- k-of-n multisigs over m-of-n multisigs

We should document all of these somewhere.

The Optech newsletter is very highly regarded. I expect other authors would want to follow the style conventions we use if they were documented somewhere.

harding:
Closes #165 

This is all I can think of for the moment.  I assume that we'll add to it in subsequent weeks.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -1,0 +1,80 @@
+# Optech Style Guide
+
+## Vocabulary
+
+### Proper nouns
+
+The first letter of a proper noun is always capitalized.
+
+- Bitcoin (the network)
+
+### Common nouns
+
+The first letter of a common noun is only capitalized at the start of a
+sentence or heading.
+
+- bitcoin (the currency)
+- schnorr
+- taproot
+- merkle (tree|root|branch|node)
+- segwit
+
+### Forbidden abbreviations
+
+- Core (use Bitcoin Core)
+
+### Unintroduced abbreviations
+
+The first occurrence of an abbreviation in a document should be preceded
+by its full name (e.g. Foo Bar Baz (FBB)) with the exception of the
+following abbreviations we assume readers will already know.
+
+- BIP (Bitcoin Improvement Proposal)
+- BTC (Bitcoin)
+- ECDSA (Elliptic Curve Digital Signature Algorithm)
+- kB, MB, GB, TB, etc... (SI-prefixed byte sizes)
+- kiB, MiB, GiB, TiB, etc... (SI-prefixed binary byte sizes)
+- LN (Lightning Network)
+- P2P (Peer-to-Peer)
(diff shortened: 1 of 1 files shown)


jonatack:
LGTM. One point of debate that this style guide manages to avoid (with "BTC") is several bitcoin vs bitcoins. Do you sometimes state amounts in long form? IIRC, I've seen "a few bitcoin" in some Bitcoin books, and "a few bitcoins" in the Bitcoin Core codebase.

harding:
@jonatack good point, thanks!  I force pushed a commit requiring the fractional and plural forms append an `s` when written out.

STYLE.md
```diff
@@ -0,0 +1,73 @@
...
+
+| Use | Don't use |
+|-|-|
+| block chain | blockchain |
+| offchain | off-chain |
+| onchain | on-chain |
+
+### Spelling
+
+The author of the document gets to choose its flavor of English (e.g.
```
jnewbery: You've misspelled flavour :grin:

jnewbery:
Looks great. A couple of comments/additions:

- when you say 'append an `s` for fraction amounts', presumably you mean "0.2 bitcoins", but not "half a bitcoins".
- perhaps note our convention that a party making a payment is called the _spender_ (not _sender_, which could be ambiguously refer to a party sending bitcoin or a party sending data). The party receiving the payment is the _recipient_ (there's no word spendee in English, so we just have to live with the ambiguity between receiving payments and receiving data)

STYLE.md
harding: lolololol

harding:
Forced pushed edits per @jnewbery comments.  Note, I made the preferred word "receivers" not "recipients".  We (mostly me) have been inconsistent about this in the repository, but "receiver" is my preferred term and has many more `git grep` results in the repository than "recipient".  I personally think receiver sounds nicer when paired with spender, given that they both end with the same sound, but I don't have a strong opinion about this---so anyone let me know if you disagree.

jnewbery:
+1 for receiver over recipient.

ACK dde060a36dcb3a46242dc1325df8a3ac77141a1b . I'll merge now. We can append to it over the coming weeks.

Thanks Dave!

PR #169, opened 2019-07-17
Style additions

bitschmidty:
I wanted to throw in a few from my list.

A couple I was not sure about:

- P2SH-P2WSH (is this it or P2SH-wrapped P2WPKH or I have seen "nested" as well)
- P2SH-P2WPKH

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -18,6 +18,7 @@ sentence or heading.
 - taproot
 - merkle (tree|root|branch|node)
 - segwit
+- bech32
 
 ### Forbidden abbreviations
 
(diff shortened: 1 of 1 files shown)


harding:
I've probably used every possible combination for P2SH-wrapped segwit stuff depending on what I thought sounded better in context.  (E.g. sometimes I want to draw attention to how it works, sometimes I just assume people know what it is.)  I think it's good to put at least some of these on the abbreviation list but I'm not sure we want to restrict use of them to a single term.

Modulo the nit caught by the linter, this LGTM.  Thanks!

STYLE.md
```diff
@@ -40,6 +42,14 @@ following abbreviations we assume readers will already know.
 - RFC (Request For Comments, the IETF documents)
 - segwit (segregated witness)
 - vbyte (virtual byte)
+- UTXO (Unspent Transaction Output)
```
jnewbery: Please merge this into the existing list alphabetically!

STYLE.md
```diff
@@ -48,6 +58,10 @@ following abbreviations we assume readers will already know.
 | block chain | blockchain |
 | offchain | off-chain |
 | onchain | on-chain |
+| multisig | multi-sig |
```
jnewbery: Please merge this into the existing list alphabetically!

bitschmidty:
squashed and pushed a commit for @jnewbery 's feedback

jnewbery:
ACK 49da4f6d98cb951189160e5181c7228516111907

jnewbery:
Once everyone has their initial set of style rules merged, I suggest we keep a long-running issue open to collect new style rules, and only merge infrequently.

STYLE.md
```diff
@@ -18,6 +18,7 @@ sentence or heading.
 - taproot
 - merkle (tree|root|branch|node)
 - segwit
+- bech32
```
practicalswift: Should these be sorted as with "Forbidden abbreviations" and "Compound words" below?

practicalswift:
utACK 49da4f6d98cb951189160e5181c7228516111907 modulo sorting nit

harding:
If this gets modified again before merge, maybe it could also include `DoS` in the abbreviations, as that's a term I used in this upcoming week's newsletter.

STYLE.md
jnewbery: I didn't comment on this because the list wasn't sorted before this PR. @harding , was there a reason for that?

harding: No reason.  I must've forgotten to do it (I typed things as they occurred to me and then sorted by piping ranges from my editor to `sort`).

jnewbery:
I've sorted the _Common nouns_ section, added DoS to the _Unintroduced acronyms_ section, and merged. Please use #171 for any further suggestions for our style guide.

PR #167, opened 2019-07-16
Indexes: add topics index

harding:
This PR adds a page `/en/topics` that provides a hand-curated index to topics we frequently mention in our newsletters and other content.  To allow for incremental deployment of this new feature, similar to what we did with the compatibility matrix, the Topics entry only appears in in the page-top navigation if the site is built with `JEKYLL_ENV=future make preview` (or another make target using the same environment variable).

This PR currently includes 40 topics and some edits to add anchor points for them in the final two commits ("add topics" and "edit content for more anchors").  Those don't need to be reviewed in this PR, they're only attached to make it possible to see what the index pages will look like and to allow exploring the different features of the templates.  When this PR is nearing completion, I'll drop those two commits here and move them to another PR.  (I've confirmed that the site builds and passes tests without those commits.)

Issue #45, opened 2018-07-27
Use explicit license


harding:
The homepage says, "all materials and documentation produced are placed in the public domain".  This is fine with me, but my non-lawyer [understanding](https://en.wikipedia.org/wiki/Public_domain#Dedicating_works_to_the_public_domain) is that public domain grants are not valid in all jurisdictions (particularly in the EU), so it's better to use a public domain grant with an explicit backup license like [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

Although CC0 has a no-warranty clause, my (again non-lawyer) interpretation is that it doesn't necessarily apply if the court decides to enforce the public domain grant, possibly opening dedicators to lawsuits for inaccurate documentation.  For that reason, I slightly prefer MIT, which is almost as liberal as public domain but requires the no-warranty statement be retained with all copies.

For comparison, Bitcoin Core, Bitcoin.org, and BitcoinCore.org all use the MIT license for their documentation, and the Bitcoin Wiki uses CC-SA for most content.  My personal preferred license is CC-SA, but I think I'm probably in the minority there in the Bitcoin community and any [FSF-vetted free license](https://www.gnu.org/licenses/license-list.en.html) is ultimately fine by me.

(This issue arose in my mind as I'm about to reuse a couple sentences I wrote for previous newsletters on BitcoinCore.org.)

jnewbery:
Sounds good. I have a slight preference for the MIT licence since I'm more familiar with it. I'll open a PR.

PR #172, opened 2019-07-23
Plugins: new auto-anchor plugin for to add ids to list items

harding:
For #167, @bitschmidty suggested to me that we could automatically generate ids for all list items.  This plugin implements that idea by using the summary text we have for most list items to generate the automatic identifier (slug).  The addition of automatic ids can be disabled on a page-by-page basis (or more broadly using the Jekyll config file with path globs) using a `auto_id: false` page config option.

The html proofer called from the Makefile already errors if we accidentally use the same id twice on the same page (overloading it; most browsers will only go to the first occurrence) and this PR includes a few minor content changes to deal with situations where that occurs.

One way to test this PR is to build the site with and without the plugin and then diff the HTML.  E.g.:

```bash
mv _plugins/auto-anchor.rb /tmp
make
cp -a _site _old_site
mv /tmp/auto-anchor.rb  _plugins/
make
diff -ru _old_site _site | colordiff | less -R
```

That's a rather large diff.  If that's too many changes for reviewers to feel comfortable with, one option is to set `auto_id: false` as a site-wide page default and then enable it page-by-page as we need it or with new posts that are well-reviewed.

### DIFF
--- a/_plugins/auto-anchor.rb
+++ b/_plugins/auto-anchor.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+## Automatically add id tags to list items that are formatted like one
+## of the following:
+# - **<Summary:** Details
+# - *<Summary:* Details
+# - [Summary][]: Details
+# - [Summary](URL): Details
+
+Jekyll::Hooks.register :documents, :pre_render do |post|
+  ## Don't process documents if YAML headers say: "auto_id: false"
+  if post.data["auto_id"] != false
+    post.content.gsub!(/^ *- .*/) do |string|
+      ## List items surrounded in bold
+      title = string.match(/\*\*.*\*\*/).to_s
+      ## List items surrounded in italics
+      title.empty? && title = string.match(/\*.*\*/).to_s
+      ## List items that are hyperlinks
+      title.empty? && title = string.match(/\[.*\][(\[]/).to_s
+
+      if title.empty?
+        ## No match, pass item through unchanged
+        string
+      else
+        ## Remove double-quotes from titles before attempting to slugify
+        title.gsub!('"', '')
+        ## Use Liquid/Jekyll slugify filter to choose our id
+        id_prefix = '- {:#{{ "' + title + '" | slugify }}} '
+        string.sub!(/-/, id_prefix)
+      end
+    end
+  end
+end
--- a/_posts/en/2019-06-14-exec-briefing.md
+++ b/_posts/en/2019-06-14-exec-briefing.md
@@ -11,6 +11,7 @@ excerpt: >
   Videos and slides from the Bitcoin Optech Executive Briefing
   on May 14th, 2019.
 
+auto_id: false
 ---
 {% include references.md %}
 
--- a/_posts/en/newsletters/2018-09-25-newsletter.md
+++ b/_posts/en/newsletters/2018-09-25-newsletter.md
@@ -58,7 +58,7 @@ infrastructure projects.
 
     - [Additional technical information][bse 79484] by Andrew Chow (also described below)
 
-    - [CVE-2018-17144][], National Vulnerability Database (NVE) entry
+    - [CVE-2018-17144 entry][cve-2018-17144], National Vulnerability Database (NVE) entry
       being updated by Luke Dashjr
 
     We're aware of several very insightful people currently reflecting
--- a/_posts/en/newsletters/2018-12-28-newsletter.md
+++ b/_posts/en/newsletters/2018-12-28-newsletter.md
@@ -204,7 +204,7 @@ the node's bandwidth and memory---problems which developers are still
 ### 2018 summary<br>Notable technical conferences and other events
 
 - [BPASE][bpase], January, Stanford University
-- [Bitcoin Core developers meetup][coredevtech nyc], March, New York
+- [Bitcoin Core developers meetup NYC][coredevtech nyc], March, New York
   City ([transcripts][coredevtech ts])
 - [L2 Summit][], May, Boston
 - [Building on Bitcoin][], July, Lisbon ([transcripts][bob ts])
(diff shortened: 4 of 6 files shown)


harding:
@bitschmidty 

> While testing I ran into 2 scenarios that were not autogenerated and perhaps could be:
>
> 1. Newsletter #45 has a nested list with emphasis instead of bold
> 2. bech32 blog post section "Top bech32 questions from the Bitcoin StackExchange" has links leading the list items.

I've updated the plugin to deal with the first case (nested lists, and lists using emphasis instead of bold).

The second case is because the summary/title part runs off the end of a line, e.g. the part between `[` and `]`:

```
- [Will a Schnorr soft fork introduce a new address
  format?]({{bse}}82952)  Although upgrading to bech32 sending support
  should be easy, you probably don't want to repeat that work for
```

To fix that, we'd have to use multi-line substitutions which both increases the chance of a mismatch that alters content in a way we don't want and also significantly increases the CPU requirements for site previewing and building.  Since the manual fix in that case is so simple---just keep the summary/title element all on one line, I haven't addressed the second case in the plugin.

> Both also seem to have a paragraph tag as the first element within the list item. Not sure why that sometimes gets added or if it is a consideration.

Not a consideration because the plugin modifies the text before it's processed into HTML by Liquid and Kramdown, so it never sees the `<p>` tags.  The reason for paragraph tags you see is because most Markdown parsers add paragraph tags if you put spaces between list items, e.g.:

```text
$ cat <<EOF | kramdown
- A
- B
- C

---

- A

- B

- C
EOF
<ul>
  <li>A</li>
  <li>B</li>
  <li>C</li>
</ul>

<hr />

<ul>
  <li>
    <p>A</p>
  </li>
  <li>
    <p>B</p>
  </li>
  <li>
    <p>C</p>
  </li>
</ul>
```

Thanks!

bitschmidty:
Thanks for the explanation about markdown parsing adding paragraphs.

For the second case, adding manual ids for any cases not handled here is not difficult, as you mentioned.

tACK 56c63bad5c1993501e4faa7a0a3be7442bf7ab94

jnewbery:
Concept ACK! Thanks @harding (and thanks @bitschmidty for reviewing)

I've tested this with a couple of random list elements and it looks good so far.

I'll have a quick skim over the ruby code and the _site diff and then merge if I don't see any problems.

_plugins/auto-anchor.rb
```diff
@@ -0,0 +1,29 @@
+## Automatically add id tags to list items that are formatted like one
```
jonatack: - use snakecase for the names of Ruby files
- add `# frozen_string_literal: true` magic comment to top of Ruby files (same for the Gemfile)

_plugins/auto-anchor.rb
```diff
@@ -0,0 +1,29 @@
...
+## of the following:
```
jonatack: - Ruby comments usually begin with a single #

_plugins/auto-anchor.rb
```diff
@@ -0,0 +1,29 @@
...
+# - **<Summary:** Details
+# - **[Summary][]: Details
+# - **[Summary](URL): Details
+Jekyll::Hooks.register :documents, :pre_render do |post|
+  ## Don't process documents if YAML headers say: "auto_id: false"
+  if post.data["auto_id"] != false
+    post.content.gsub!(/^ *- .*/) { |string|
```
jonatack: Block delimiters: use do...end rather than {...} for multi-line blocks

_plugins/auto-anchor.rb
```diff
@@ -0,0 +1,29 @@
...
+      ## List items surrounded in bold
+      title = string.match(/\*\*.*\*\*/).to_s
+      ## List items surrounded in italics
+      title.empty? and title = string.match(/\*.*\*/).to_s
```
jonatack: s/and/&&/

_plugins/auto-anchor.rb
```diff
@@ -0,0 +1,29 @@
...
+      ## List items that are hyperlinks
+      title.empty? and title = string.match(/\[.*\][(\[]/).to_s
```
jonatack: s/and/&&/

_plugins/auto-anchor.rb
```diff
@@ -0,0 +1,29 @@
+## Automatically add id tags to list items that are formatted like one
+## of the following:
+# - **<Summary:** Details
+# - **[Summary][]: Details
+# - **[Summary](URL): Details
+Jekyll::Hooks.register :documents, :pre_render do |post|
+  ## Don't process documents if YAML headers say: "auto_id: false"
+  if post.data["auto_id"] != false
```
jonatack: Since this file is generally using single-quotes, which can make sense when you don't need string interpolation or special symbols, single quotes can be used here.

_plugins/auto-anchor.rb
```diff
@@ -0,0 +1,29 @@
...
+    post.content.gsub!(/^ *- .*/) { |string|
+      ## List items surrounded in bold
+      title = string.match(/\*\*.*\*\*/).to_s
+      ## List items surrounded in italics
+      title.empty? and title = string.match(/\*.*\*/).to_s
+      ## List items that are hyperlinks
+      title.empty? and title = string.match(/\[.*\][(\[]/).to_s
+
+      if title != ''
```
jonatack: `unless title.empty?`

harding:
I implemented a few of @jonatack's suggestions (thanks!).  To test that these created no actual change in the site, I diff'd the HTML output from before and after the edits and the diffs were equivalent (excepting expected changes in build timestamps).

_plugins/auto-anchor.rb
```diff
@@ -0,0 +1,33 @@
...
+  if post.data["auto_id"] != false
+    post.content.gsub!(/^ *- .*/) do |string|
+      ## List items surrounded in bold
+      title = string.match(/\*\*.*\*\*/).to_s
+      ## List items surrounded in italics
+      title.empty? && title = string.match(/\*.*\*/).to_s
+      ## List items that are hyperlinks
+      title.empty? && title = string.match(/\[.*\][(\[]/).to_s
+
+      if title.empty?
```
jonatack: If the original `if title != ''` was correct, the equivalent would be either `if !title.empty?` or `unless title.empty?` (or if ActiveSupport is loaded, `if title.present?`)

ah Ruby, so many ways to do the same thing

harding: I also reversed the branches.  E.g. before it was: `if title != '' { a } else { b } end` and now it's `if title.empty? { b } else { a } end`

jonatack: My bad. Even better.

bitschmidty:
tACK c9667d6d6085b10d5ea08b09b318aba1ab75804a

jnewbery:
lightly reviewed ACK 073beb87696ad5a174e3403cc3b0d0313be1a766

PR #181, opened 2019-08-06
Compatibility data: electrum greenaddress jaxx ledger-live mycelium

bitschmidty:


jnewbery:
rebased

jnewbery:
Added mycelium

jnewbery:
@moneyball - please review for typos/formatting/etc. Let me know if you can't review this week. Thanks!

bitschmidty:
pushed commits with: 

- changes to the captions prefixes made per #183 (comment)
- changes to caption wording based on Bitcoin Core wallet example
- moved images into /rbf/ sub folder to be consistent with /segwit/ usability images (and more clear)
- changes to 'na' vs 'untested' for rbf sending tests when service doesn’t send via rbf

bitschmidty:
rebased on master to get the latest compatibility styling and templates

moneyball:
Screen Shot 2019-08-13 at 4 15 27 PM

i noticed that whichever section was clicked through to gets this blue background. it looks a bit out of place when scrolling around the page. is it intentional?

moneyball:
Screen Shot 2019-08-13 at 4 16 54 PM

i'm wondering if we should have X instead of ? for wallets that don't support BIP125...the remaining portions of sending RBF are known to not work, right, so should be X not ?

moneyball:
Screen Shot 2019-08-13 at 4 18 42 PM

is the checkmark the same size as the X? my eyes tell me the X is bigger, which makes it more prominent, and almost fooled me into thinking that it was affirmative that the feature was supported.

moneyball:
Should we put Segwit addresses section above RBF? I realize that isn't the order the research was conducted, but I would think we should order it based on what we believe will be of interest to our audience. I suspect there is more interest in Segwit support than RBF.

_data/compatibility/mycelium.yaml
```diff
@@ -0,0 +1,66 @@
...
+  examples:
+    - image: /img/compatibility/mycelium/segwit/receive-screen.png
+      caption: >
+        Mycelium only generates P2PKH receive addresses.
+    - image: /img/compatibility/mycelium/segwit/send-screen.png
+      caption: >
+        Mycelium cannot send to any bech32 addresses.
+    - image: /img/compatibility/mycelium/segwit/send-screen-segwit-v1-error.png
+      caption: >
+        Jaxx displays a validation error when trying to send to a segwit v1
```
moneyball: Mycelium not Jaxx

bitschmidty:
> i noticed that whichever section was clicked through to gets this blue background. it looks a bit out of place when scrolling around the page. is it intentional?

Yep. We wanted to anchor link the user to the applicable section (based on the link they clicked on) and also visually indicate the section of the page that was referenced from the preceding page. Do you think there is something more we can do to clarify? 

> i'm wondering if we should have X instead of ? for wallets that don't support BIP125...the remaining portions of sending RBF are known to not work, right, so should be X not ?

I agree the ? might be confusing for the status of "untested" which is what those tests are. I think it is best to have a dash '-'. Which is actually the character that we use for "N/A" status in the table as well. The dash icon is still clickable through to the details of the test. I dont like the X because we really were not able to test it, so "untested" is correct, but I agree ? might not be best.

> Should we put Segwit addresses section above RBF? I realize that isn't the order the research was conducted, but I would think we should order it based on what we believe will be of interest to our audience. I suspect there is more interest in Segwit support than RBF.

I dont have a strong preference on this. @jnewbery ?

harding:
I'd also slightly prefer Segwit before RBF.  Changing that only requires moving stuff around in one file, so it shouldn't be much work (I'm happy to do it).

_data/compatibility/mycelium.yaml
bitschmidty: Good catch, thanks @moneyball

jnewbery:
> Should we put Segwit addresses section above RBF?

> I'd also slightly prefer Segwit before RBF. Changing that only requires moving stuff around in one file, so it shouldn't be much work (I'm happy to do it).

ACK. Let's do that.

I also find it slightly odd that now that the RBF and segwit tables link to `#rbf` and `#segwit`, there isn't a direct link to the top of the details page. One thing we could potentially do is change rbf and segwit to be tabs or collapsible sections.

harding:
@jnewbery 

> One thing we could potentially do is change rbf and segwit to be tabs or collapsible sections.

Putting the sections inside `<details>` tags would be easy for making them collapsible, but if they're collapsed by default then the links to the anchor targets couldn't automatically expand them without JS.  So far we've avoided any JS dependencies on the site except for special features like particular bech32 sending support sections.  (I don't know how to do tabs at all without JS, but even if there's a pure-CSS way, I suspect it would have the same problem.)

bitschmidty:
squashed commits to one per evaluated service

bitschmidty:
pushed a commit for the updated Blockstream Green logo vs older GreenAddress logo

moneyball:
I still see ? instead of dash.

My understanding is that the checkmark size will not be changed.

bitschmidty:
> I still see ? instead of dash.

I think we can leave this as untested. If the icons become confusing for users (for this reason or others we have not thought of) we can revisit.

> My understanding is that the checkmark size will not be changed.

Yep!

bitschmidty:
pushed a commit which removed segwitv1 usability examples, since we are not including segwitv1 data for now

PR #182, opened 2019-08-06
Compatibility data: opendime samourai trezor wasabi xapo

bitschmidty:


jnewbery:
rebased

jnewbery:
removed mycelium

jnewbery:
I'll review this for typos/formatting/etc this week.

harding:
I was skimming this in order to maybe steal some of @bitschmidty's research and I wanted to suggest, for consideration, that Trezor might be better renamed as Wallet.Trezor.Com or something because the review is of the web frontend rather than the Trezor device.

bitschmidty:
@harding Similar to the Ledger Live (vs Ledger), right? 

It seems that Trezor's wallet is called Trezor Wallet (https://wallet.trezor.io/#/)

There is a similar consideration for Opendime.

harding:
@bitschmidty yeah, Ledger Live vs Ledger too.  For OpenDime, I only skimmed the listing here (I was looking for screenshots of Send screens) but it looked to me like you reviewed the hardware independent of a software wallet, so I think it's probably appropriately named.

Bitcoin.org wallet maintainer @crwatkins has previously told me that he'd have preferred it if Bitcoin.org, instead of evaluating and listing hardware wallets by themselves, instead did them in pairs with compatible software wallets.  E.g. instead of listing "Trezor" they'd list "Trezor+Trezor.com", "Trezor+Electrum", "Trezor+HWI+Core", etc...  That way if the scoring said the wallet did X, the reader could be sure that it did X when used with the indicated software.

crwatkins:
At the risk of jumping in the middle of a PR that I haven't fully reviewed and carrying on a personal chat with @harding in the middle of it, I'll agree with what [he said](https://github.com/bitcoinops/bitcoinops.github.io/pull/182#issuecomment-519605583).  As we know, most "hardware wallets" are really just key stores with signing engines (with the notable exception of the discontinued Case wallet, and the non-notable exceptions which are just software running on top of Android on some mobile-phone-like general purpose hardware). They are not fully functioning wallets by themselves.

The issue at bitcoin.org currently is that most hardware cannot be reviewed as a wallet in isolation (perhaps as an engine, but not as a wallet) and we review "wallets".   Another complication is that historically when a new "hardware wallet" comes out from a new vendor, their companion software is often fairly immature and does not pass recommended bitcoin.org wallet criteria, requiring the use of a third party app to meet all the criteria.

jnewbery:
I've pushed a commit that updates the image files. Ideally we want images that:

- Are an image on a neutral background
- that background is transparent (not white)
- contain the name of the service/wallet

_data/compatibility/opendime.yaml
```diff
@@ -0,0 +1,58 @@
...
+    send:
+      signals_bip125: "false"
+      list: "false"
+      details: "na"
+      shows_replaced_version: "na"
+      shows_original_version: "na"
+  examples:
+    - image: /img/compatibility/opendime/sending-instructions.png
+      caption: >
+        Sending RBF Transaction - Opendime does not have send transaction capabilities. You need to sweep the private key to use in a regular wallet.
```
jnewbery: Suggest removing the 'Sending RBF Transaction' since the software does not allow sending an RBF-signaling tx

_data/compatibility/opendime.yaml
```diff
@@ -0,0 +1,58 @@
...
+    date: "2019-07-24"
+    platforms:
+      - macOS
+    version: "2.2.0"
+  features:
+    receive:
+      p2sh_wrapped: "false"
+      bech32: "false"
+      default: "p2pkh"
+    send:
```
jnewbery: Is it possible to just hide this section rather than have the repeated "No sending support"?

_data/compatibility/opendime.yaml
```diff
@@ -0,0 +1,58 @@
...
+      bech32: "na"
+      change_bech32: "na"
+      segwit_v1: "Sending transactions via Opendime software is not supported.
+      Private keys are used by user-selected wallet software to create transactions."
+      bech32_p2wsh: "na"
+  examples:
+    - image: /img/compatibility/opendime/segwit/receive-screen.png
+      caption: >
+        Opendime has a single P2PKH receive address. It is not capable of
```
jnewbery: Can the balance.py program be used to send funds off the opendime?

_data/compatibility/samourai.yaml
```diff
@@ -0,0 +1,97 @@
...
+        Sending RBF Transaction - Send RBF bumped transaction replaces original RBF enabled transaction. No flag. Confirmation message does say “RBF transaction sent”
+    - image: /img/compatibility/samourai/transaction-list-incoming-rbf.png
+      caption: >
+        Receiving RBF Transaction - Transaction list screen for receiving RBF enabled transaction. No RBF flag.
+    - image: /img/compatibility/samourai/transaction-details-incoming-rbf.png
+      caption: >
+        Receiving RBF Transaction - Transaction details prompt for received transaction. “View transaction” defaults to Smartbit explorer (explorer configurable in settings).
+    - image: /img/compatibility/samourai/transaction-details-incoming-rbf-increase-fee.png
+      caption: >
+        Receiving RBF Transaction - Transaction details “Increase transaction fee” dialog.
```
jnewbery: I don't understand this. How can a received transaction be bumped using RBF?

_data/compatibility/samourai.yaml
```diff
@@ -0,0 +1,97 @@
...
+        Receiving Bumped RBF Transaction - Transaction list screen showing both original and bumped transactions. Total reflecting sum of both transactions. No RBF flag or warnings.
+    - image: /img/compatibility/samourai/transaction-list-replacement-confirmed.png
+      caption: >
+        Receiving Bumped RBF Transaction - Transaction list screen updated to just show replacement transaction after a period of time. Even before either RBF transaction was confirmed.
+segwit:
+  tested:
+    date: "2019-07-25"
+    platforms:
+      - Android
+    version: "0.99.82"
```
jnewbery: I don't understand this versioning. Version tested for RBF was 3.3.2, and this is 0.99.82?

_data/compatibility/trezor.yaml
```diff
@@ -0,0 +1,68 @@
...
+  features:
+    receive:
+      notification: "na"
+      list: "false"
+      details: "false"
+      shows_replaced_version: "true"
+      shows_original_version: "false"
+    send:
+      signals_bip125: "false"
+      list: "false"
```
jnewbery: how are these things testable (sent transaction labeled as RBF in summary and details screen), when the service doesn't support sending with RBF?

_data/compatibility/trezor.yaml
```diff
@@ -0,0 +1,68 @@
...
+      details: "false"
+      shows_replaced_version: "na"
+      shows_original_version: "na"
+  examples:
+    - image: /img/compatibility/trezor/send-screen.png
+      caption: >
+        Sending RBF Transaction - Send transaction screen. Fee options. No RBF options. Transaction not sent RBF.
+    - image: /img/compatibility/trezor/transaction-list-sent.png
+      caption: >
+        Bumping RBF Transaction - Transaction list screen showing sent transaction. No bumping options.
```
jnewbery: Remove words 'Bumping RBF Transaction' since the service doesn't allow bumping.

_data/compatibility/trezor.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - image: /img/compatibility/trezor/transaction-list-incoming-rbf.png
+      caption: >
+        Receiving RBF Transaction - Incoming transaction list. No RBF flag.
+    - image: /img/compatibility/trezor/transaction-details-incoming-rbf.png
+      caption: >
+        Receiving RBF Transaction - Trezor block explorer for incoming RBF transaction.
```
jnewbery: Add note that it doesn't display that the transaction is RBF.

_data/compatibility/trezor.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - image: /img/compatibility/trezor/transaction-list-incoming-replacement.png
+      caption: >
+        Receiving Bumped RBF Transaction - Bumped transaction replaces original transaction. Block explorer no longer finds original transaction either.
```
jnewbery: remove word 'either'

_data/compatibility/xapo.yaml
```diff
@@ -0,0 +1,82 @@
...
+        Receiving Bumped RBF Transaction - Even after bumped transaction has 100 confirms, original transaction stays pending.
+segwit:
+  tested:
+    date: "2019-07-27"
+    platforms:
+      - iOS
+    version: "5.0.3"
+  features:
+    receive:
+      p2sh_wrapped: "false"
```
jnewbery: This seems inconsistent. It says xapo does not support receiving the wrapped-segwit, but its default receiving address is wrapped-segwit.

_data/compatibility/opendime.yaml
bitschmidty: Agreed here. Brought it up in a discussion here for consensus: https://github.com/bitcoinops/bitcoinops.github.io/pull/183#issuecomment-519931177

_data/compatibility/opendime.yaml
bitschmidty: Yep, balance.py can spend. And can send to bech32 p2wsh and p2wpkh. Im going to change the comment here as well as the results, including "balance.py" as the wallet in this scenario.

_data/compatibility/opendime.yaml
bitschmidty: Per https://github.com/bitcoinops/bitcoinops.github.io/pull/182#r312521548 there are some sending capabilities if we include the balance.py file (which I think is reasonable).

_data/compatibility/samourai.yaml
bitschmidty: @jnewbery This screenshot is illustrating that an incoming transaction (including incoming rbf signaling transactions) can have its transaction fee increased (via cpfp). I thought it was interesting to include even if it was not using rbf to do so.

_data/compatibility/samourai.yaml
bitschmidty: This version is wrong. It is 0.81. Good catch.

_data/compatibility/trezor.yaml
bitschmidty: Will resolve this as part of the large "prefix" discussion.

_data/compatibility/trezor.yaml
bitschmidty: You’re right. "Doesn’t flag RBF transactions (incoming)" doesn’t mean "Doesn’t flag RBF transactions outgoing". 

I will make sure this is consistent across other data files as well.

_data/compatibility/xapo.yaml
bitschmidty: Inconsistent and just WRONG. Corrected.

bitschmidty:
Pushed a commit to address (most of) @jnewbery 's great feedback.

_data/compatibility/samourai.yaml
jnewbery: That is interesting. Can you make that explicit in the caption?

jnewbery:
Thanks @bitschmidty . Let me know when you've updated the captions (https://github.com/bitcoinops/bitcoinops.github.io/pull/183#issuecomment-519931177) and I'll rereview.

_data/compatibility/samourai.yaml
bitschmidty: Added!

bitschmidty:
@jnewbery 

1. changes to the captions prefixes made per https://github.com/bitcoinops/bitcoinops.github.io/pull/183#issuecomment-519931177
2. changes to caption wording based on Bitcoin Core wallet example
3. moved images into /rbf/ sub folder to be consistent with /segwit/ usability images (and more clear)
4. changes to 'na' vs 'untested' for rbf sending tests

jnewbery:
Thanks Mike. Will rereview on Monday.

_data/compatibility/opendime.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - macOS
+    version: "2.2.0"
+  features:
+    receive:
+      p2sh_wrapped: "false"
+      bech32: "false"
+      default: "p2pkh"
+    send:
+      bech32: "true"
+      change_bech32: "na"
```
jnewbery: This looks a little weird:

image

I suggest we update the template text from **No sending capabilities** to **No sending or change capabilities**

jnewbery:
Looks great @bitschmidty . I've pushed a branch that:
- rebases on master
- squashes commits so there's one for each wallet
- adds a commit to change the wording for the template as suggested here: https://github.com/bitcoinops/bitcoinops.github.io/pull/182/files/c700ffe5fdbb8798f6e22c72d750c6f77cb787dc#diff-e907fc5ba91b63ae75420a185d17829b

If you're happy with those changes, please go ahead and merge.

bitschmidty:
@jnewbery looks great, thank you.

PR #185, opened 2019-08-07
Add 'make production' to check for unadressed FIXMEs

harding:
This adds an extra makefile target that can be run manually before merging to check for any issues mentioned in the source code that haven't been addressed yet.  I don't know how the Netlify integration works, but if it builds by calling `make` or `make all` and fails to upload the built site if `make` returns false, then changing it to use the new `make production` would prevent it from uploading any changes adding FIXMEs to the site.

A second commit here also re-arranges the existing tests so that source-code tests are run before the site build.  As we've been adding more pages to the site, the build has increased to more than a trivial amount of time.  This allows some tests to fail without having to wait for building.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 all: build test
+production: all production-test
 
 preview:
 	bundle exec jekyll clean
@@ -26,3 +27,9 @@ test:
 
 	## Check for broken links
 	bundle exec htmlproofer --check-html --disable-external --url-ignore '/^\/bin/.*/' ./_site
+
+## Tests to run last because they identify problems that may not be fixable during initial commit review.
+## However, these should not be still failing when the site goes to production
+production-test:
+	## Fail if there are any FIXMEs in site source code; add "skip-test" to the same line of source code to skip
+	! git --no-pager grep FIXME | grep -v skip-test | grep .
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-all: build test
+all: test-before-build build test-after-build
 production: all production-test
 
 preview:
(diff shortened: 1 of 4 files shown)


jnewbery:
Thanks @harding! 

@adamjonas - do you know how netlify builds the site?

adamjonas:
https://5d4afce478545e00084ee37b--bitcoinops.netlify.com/

jnewbery:
@adamjonas - specifically, do you know whether netlify calls `make` or `make all` and fails to upload the built site if `make` returns false? Is it possible to change it to use `make production`? (we'd still want it to use `make` for PR previews.)

adamjonas:
tACK e36b4f3

Was building with `make`:
Screen Shot 2019-08-07 at 3 06 07 PM

Built with `make production` for this branch:
Screen Shot 2019-08-07 at 3 15 52 PM

bitschmidty:
tACK e36b4f33f2db4f6c5205d257012e0b8015e2530f

Great ideas @harding 

I will add running `make production` as part of the newsletter publishing process docs as well.

Makefile
```diff
@@ -8,21 +9,29 @@ build:
...
 
+test-after-build:
+	## Check for broken Markdown reference-style links that are displayed in text unchanged, e.g. [broken][broken link]
+	! find _site/ -name '*.html' | xargs grep ']\[' | grep -v skip-test | grep .
+	! find _site/ -name '*.html' | xargs grep '\[^' | grep .
+
 	## Check for broken links
 	bundle exec htmlproofer --check-html --disable-external --url-ignore '/^\/bin/.*/' ./_site
+
+## Tests not run by CI because they might by dependent on resources that aren't available yet.
```
jnewbery:
```suggestion
## Tests not run by CI because they might be dependent on resources that aren't available yet.
```

jnewbery:
One typo correction inline.

I don't understand the comment "Tests not run by CI...". Don't we want the production tests to be run by CI so we know that they're failing before merge?

Perhaps we can run `make production` on travis, but `make all` on netlify, so github will show checks as failing, but we can still view the preview in netlify?

harding:
> Don't we want the production tests to be run by CI so we know that they're failing before merge?

I think we want the preview to work, which I thought was tied to the CI (but they actually seem to be separate).  I'll change the travis.yml to run `make production`.

jnewbery:
rebased on master. I think this is ready for merge once the travis config has been updated

harding:
Makefile comment and travis config updated, and the log from the latest build confirms `make production` is being used: https://travis-ci.org/bitcoinops/bitcoinops.github.io/builds/571598720#L1024

jnewbery:
Splendid. Thanks Dave!

PR #168, opened 2019-07-16
Add Compatibility Section to the site

bitschmidty:
This PR adds:

1. A new navigation item on the site for Compatibility
2. A Compatibility page summarizing, in a matrix, RBF-capability tested Bitcoin services/wallets
3. Individual Bitcoin service/wallet detail pages with more RBF testing details

PR #190, opened 2019-08-09
Compat: reactivate schema checker

harding:
Based on top of #185, ignore first two commits..

This was previously included as part of #168 but it seems to prevent Netlify from successfully building both previews and production deployments of the site so it was disabled in #174.  To debug, it'd be useful to get a copy of the Netlify build log for the preview from @adamjonas 

This runs for me locally and it previously ran successfully on Travis.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,4 +14,6 @@ end
 group :testing do
   gem 'html-proofer'
   gem 'mdl'
+  gem 'json-schema'
+  gem 'toml'
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -52,6 +52,8 @@ GEM
       jekyll (~> 3.3)
     jekyll-watch (2.0.0)
       listen (~> 3.0)
+    json-schema (2.8.1)
+      addressable (>= 2.4)
     kramdown (1.17.0)
     liquid (4.0.0)
     listen (3.1.1)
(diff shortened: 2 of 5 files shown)


bitschmidty:
@harding 

I pushed this to my netlify instance as is and it failed to build. (log attached)

I [took out the "BUNDLED WITH" directive in Gemfile.lock](https://github.com/bitschmidty/bitcoinops.github.io/commit/87d3a2a3d6a9d9e61f953f22cea0c725660e33ed) and netlify built correctly. (log attached)

[netlify-successful-build.txt](https://github.com/bitcoinops/bitcoinops.github.io/files/3486595/netlify-successful-build.txt)
[netlify-failed-build.txt](https://github.com/bitcoinops/bitcoinops.github.io/files/3486596/netlify-failed-build.txt)

harding:
@bitschmidty I cherry picked your commit but it doesn't seem to have changed things for the PR preview here.  I think we still need logs from the bitcoinops netlify build.

adamjonas:
@harding 

```

harding:
I give up.  It appears to be something specific to our Netlify instance (rather than their system in general) that I can't debug myself.  Until this is resolved, it appears that we're stuck with the ruby gem dependencies we have cached on Netlify and can't add or upgrade any.  For now, we can just skip having this extra test.

For reference, we seem to be failing here: https://github.com/netlify/build-image/blob/trusty/run-build-functions.sh#L318

PR #194, opened 2019-08-12
Workshops: add NYC and SF taproot/schnorr workshops

bitschmidty:
This PR adds both NYC and SF taproot/schnorr workshops to the workshops page.

I have included them as separate workshops even though the topics are the same. That way we can separately detail the attendees as well as thank associated companies for hosting in the respective venues.

### DIFF
--- a/workshops.md
+++ b/workshops.md
@@ -9,26 +9,40 @@ together to discuss approaches and challenges in implementing scaling
 technologies. Each workshop will be tailored to the member companies attending
 and the specific scaling challenges that they are facing.
 
-- Topics are discussed in a roundtable format in which every participant has an
-  equal opportunity to engage.
-
-- Each topic has a moderator and notetaker. The moderator is responsible
-  for a brief introduction of a topic and keeping discussion on track and on
-  time.
-
-- We want participants to be comfortable to speak freely. Notes and action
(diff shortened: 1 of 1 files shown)


workshops.md
```diff
@@ -9,26 +9,40 @@ together to discuss approaches and challenges in implementing scaling
...
 
-- Topics are discussed in a roundtable format in which every participant has an
-  equal opportunity to engage.
+If you have any requests or suggestions for future workshop events, please
+[contact us][optech email].
 
-- Each topic has a moderator and notetaker. The moderator is responsible
-  for a brief introduction of a topic and keeping discussion on track and on
-  time.
+## <span id="taproot-workshop">Workshops #3 & #4 - Schnorr and Taproot Seminars</span>
```
harding: Instead of the span, add `{#id-you-want}` after the header definition (this is a Kramdown extension not supported in all Markdown flavors).  E.g.:

    ## Workshops #3 & #4 - Schnorr and Taproot Seminars {#taproot-workshop}

jnewbery:
I've made some suggested changes in a commit. @bitschmidty - let me know what you think.

workshops.md
bitschmidty: Added

PR #197, opened 2019-08-14
CSS: make blue :target highlight fade over 5 seconds

harding:
In https://github.com/bitcoinops/bitcoinops.github.io/pull/181#issuecomment-521044085 , @moneyball mentioned that he found the `:target` highlighting a bit confusing.  This maybe makes it more clear that it's just highlighting to help the user identify position by having it fade over time (currently 5 seconds, we could make it longer or shorter).

### DIFF
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -179,8 +179,14 @@ div.xoverflow {
   padding:20px
 }
 
+@keyframes highlightfade {
+    from { background: aqua; }
+    to { background: transparent; }
+}
+
 :target {
-  background-color: #daf8fa;
+    animation-name: highlightfade;
+    animation-duration: 7s;
 }
 
 td.compat_yes {


bitschmidty:
Funny enough I had this same idea. Looks great! 

tACK 19c8838538d343403dbc7a0bf61ef0d820afded2

assets/css/main.scss
```diff
@@ -179,8 +179,14 @@ div.xoverflow {
...
 
+@keyframes highlightfade {
+    from { background: #daf8fa; }
+    to { background: transparent; }
+}
+
 :target {
-  background-color: #daf8fa;
+    animation-name: highlightfade;
+    animation-duration: 5s;
```
bitschmidty: nit: I think we could go a bit longer.

jnewbery: I think we can also afford to make the highlighting a bit darker if it's going to fade. i tried `#8c80ff` and it looked ok to me.

jnewbery: awww, that's cute. Thanks for the color swatch, github! `#ff0000` `#00ff00` `#0000ff`

harding:
Changed the color slightly to a nearby named color and bumped up the time from 5 seconds to 7 seconds.  I tried longer times, but around 10 seconds, the fade is so slow that it seems choppy to me (I think this an optical illusion where my eye/brain *assumes* the color isn't changing because it hasn't changed enough in a certain amount of time, then it passes my threshold and I suddenly see the change).

jnewbery:
:ocean:

PR #198, opened 2019-08-14
Compat: re-order RBF and segwit

harding:
As requested by @moneyball here: https://github.com/bitcoinops/bitcoinops.github.io/pull/181#issuecomment-521045558 and commented in in that PR by @bitschmidty, @jnewbery, and myself.


bitschmidty:
tACK 3f97a10ce8cfe5f305c48e3086706994990261fd

Looks good. Thanks for getting this out so quickly @harding

jnewbery:
looks good. Confirmed that this is almost entirely move-only with `git diff --color-moved=dimmed-zebra`

Issue #195, opened 2019-08-13
PNG optimization


0xB10C:
As suggested in https://github.com/bitcoinops/bitcoinops.github.io/pull/183#issuecomment-520550392 (and following comments) PNG's displayed on the site can and should be optimized. @harding and I would opt to use the tool [`optipng`](http://optipng.sourceforge.net/). The tool offers different optimization levels. Using the level `-o7` can be slow, but results in presumably the best optimization. 

Based on the previous comments I propose following next steps:

- [x] PNG files currently included or added in open PRs should be manually optimized with `-o7` in one or more PRs. (addressed in #199)
- [x] A Travis test should be introduced that checks PNG's. The test should fail if a PNG is not optimized.  (addressed in  #200)

PR #200, opened 2019-08-14
test: Add travis check ensuring PNGs are optimized

0xB10C:
This adds a check to the `test-before-build` section of the Makefile that tests if a **newly added or changed** PNG is optimized.  <s>The check only runs (i.e. takes time) for PRs on Travis and not locally.</s> Existing PNGs are not checked to keep Travis build time as short as possible.

<s>Thus reviewing this PR could be a bit tricky.</s> To showcase functionality I've created a PR (https://github.com/0xB10C/bitcoinops.github.io/pull/2) on my travis-enabled fork that adds two PNGs. One optimized and one unoptimized. The Travis [build fails](https://travis-ci.com/0xB10C/bitcoinops.github.io/builds/123239599) on the unoptimized PNG. <s> A diligent reviewer could use a similar setup. </s>

```
## Check that newly added or modifyed PNGs are optimized
sh _contrib/travis-check-png-optimized.sh
checking img/optimized.png
checking img/unoptimized.png
The file img/unoptimized.png is not optimized. To optimize use 'optipng -o7 img/unoptimized.png'.
```

edit: This should only be merged after #195

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,3 +28,4 @@ branches:
 before_install:
   - gem update --system
   - gem install bundler
+  - sudo apt install optipng
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,9 @@ test-before-build:
 	## Check that all slugs are unique
 	! git --no-pager grep -h "^slug: " _posts | sort | uniq -d | grep .
 
+	## Check that newly added or modifyed PNGs are optimized
+	_contrib/travis-check-png-optimized.sh
+
 test-after-build:
 	## Check for broken Markdown reference-style links that are displayed in text unchanged, e.g. [broken][broken link]
 	! find _site/ -name '*.html' | xargs grep ']\[' | grep -v skip-test | grep .
--- a/_contrib/travis-check-png-optimized.sh
+++ b/_contrib/travis-check-png-optimized.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+## Finds newly added or modified PNG files in the branch and checks if each file is optimized.
+## If not optimized, it fails with an error code of 1.
+
+## This check might be run on a system where optipng is not installed (e.g. Netlify). If so, the check should not fail.
+if ! which optipng > /dev/null; then
+	echo "Skipping check since optipng is not istalled."
+	exit 0
+fi
+
+## Travis is not aware of the origin/master branch. Needs to be added and fetched first.
+## As per https://github.com/travis-ci/travis-ci/issues/6069#issuecomment-266546552
+if [ "$TRAVIS" = "true" ]; then
+	git remote set-branches --add origin master
+	git fetch
+fi
+
+EXITCODE=0
+CHANGEDFILES=$(git diff --name-only --diff-filter=AM origin/master HEAD)
+
+for FILE in $CHANGEDFILES; do
+	if echo "$FILE" | grep -q ".png$"; then 
+		echo checking "$FILE"
+		OPTIPNGOUTPUT=$(optipng -simulate "$FILE" 2>&1)
+		if ! echo "$OPTIPNGOUTPUT" | grep -q "is already optimized"; then
+			echo "The file $FILE is not optimized. To optimize use 'optipng -o7 $FILE'."
+			EXITCODE=1
+		fi
+	fi
+done
+
+exit $EXITCODE


Makefile
```diff
@@ -22,6 +22,9 @@ test-before-build:
 	## Check that all slugs are unique
 	! git --no-pager grep -h "^slug: " _posts | sort | uniq -d | grep .
 
+	## Check that newly added or modifyed PNGs are optimized
+	sh _contrib/travis-check-png-optimized.sh
```
practicalswift: `sh` needed? Could do `_contrib/travis-check-png-optimized.sh` assuming it is executable?

_contrib/travis-check-png-optimized.sh
```diff
@@ -0,0 +1,17 @@
...
+CHANGEDFILES=$(git diff --name-only --diff-filter=AM $TRAVIS_COMMIT_RANGE)
+
+for FILE in $CHANGEDFILES; do
+	if echo $FILE | grep -q ".png$"; then 
+		echo checking $FILE
+		OPTIPNGOUTPUT=$(optipng -simulate "$FILE" 2>&1)
+		if ! echo $OPTIPNGOUTPUT | grep -q "is already optimized"; then
+			echo "The file $FILE is not optimized. To optimize use 'optipng -o7 $FILE'."
+			exit 1
+	fi
```
practicalswift: Indentation seems off here. Indentation should be a multiple of two or four spaces?

_contrib/travis-check-png-optimized.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+## Finds newly added or modifyed PNG files in the repository and checks if each file is optimized.
+## If not optimized it fails with a error code 1
+
+CHANGEDFILES=$(git diff --name-only --diff-filter=AM $TRAVIS_COMMIT_RANGE)
```
practicalswift: Nit: Consider running `shellcheck` on this script and address nits :-)

harding: Maybe instead of `$TRAVIS_COMMIT_RANGE`, use `$(git merge-base master HEAD)` as something that's likely to be available to any of us?

0xB10C: Thanks! Didn't know about `shellcheck`. Ran `spellcheck` on it as well :-)

0xB10C:
Thanks for the review! Addressed nits.

_contrib/travis-check-png-optimized.sh
```diff
@@ -0,0 +1,17 @@
...
+
+CHANGEDFILES=$(git diff --name-only --diff-filter=AM "$TRAVIS_COMMIT_RANGE")
+
+for FILE in $CHANGEDFILES; do
+	if echo "$FILE" | grep -q ".png$"; then 
+		echo checking "$FILE"
+		OPTIPNGOUTPUT=$(optipng -simulate "$FILE" 2>&1)
+		if ! echo "$OPTIPNGOUTPUT" | grep -q "is already optimized"; then
+			echo "The file $FILE is not optimized. To optimize use 'optipng -o7 $FILE'."
+			exit 1
```
harding: I'd prefer the test identify and print a listing of all failed files instead of just the first failure, that way they can all be fixed in one go (especially useful since this could be a slow test if there are lots of files being added).  E.g. `for FILE in $CHANGEDFILES; do [...] ; done | grep -q 'is not optimized' ; if [ $? -eq 0 ]; then exit 1 ; fi`

_contrib/travis-check-png-optimized.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+## Finds newly added or modified PNG files in the repository and checks if each file is optimized.
+## If not optimized, it fails with an error code of 1
+
+CHANGEDFILES=$(git diff --name-only --diff-filter=AM "$TRAVIS_COMMIT_RANGE")
```
harding: Maybe instead of `$TRAVIS_COMMIT_RANGE` use `$(git merge-base master HEAD)` so this is easier to run on our own systems.

.travis.yml
```diff
@@ -28,3 +28,4 @@ branches:
 before_install:
   - gem update --system
   - gem install bundler
+  - sudo apt install optipng
```
harding: The make target you've added this to is also run on Netlify's container before deployment, where I expect it'll fail due to the lack of the special Travis variable and optipng.  My suggestion is that the Makefile be edited to create a separate `travis` build target that depends on `production` and also includes this test, e.g.:

```makefile
travis: production
  	## Check that newly added or modifyed PNGs are optimized
	_contrib/travis-check-png-optimized.sh
```

Then update the `.travis.yml` to call `make travis` instead of the current `make production`.

harding: Oh, it succeeds because there's no diff when that travis variable is unavailable, so maybe this is unnecessary.

_contrib/travis-check-png-optimized.sh
0xB10C: Good point. I'm addressing this in an upcoming push.

.travis.yml
0xB10C: > Oh, it succeeds because there's no diff when that travis variable is unavailable, so maybe this is unnecessary.

When not using `$TRAVIS_COMMIT_RANGE` this becomes a issue again though. But I'd rather not name the build target `travis`. A developer would need to run `make travis` locally, which could be a bit confusing.

0xB10C: Do you have ideas for a different name? I thought maybe `make check`...

harding: Maybe you could leave it where it is and make one of the first lines of the script something like `if ! which optipng >/dev/null ; then echo "optipng not installed; skipping test..."; exit ; fi`

0xB10C:
Thanks @harding, I've addressed your comments. The check now runs for all added/changed images and lists them before failing if one of them is not optimized (see travis log of [Add a third unoptimized.png](https://travis-ci.com/0xB10C/bitcoinops.github.io/builds/123273022)). I don't use `$TRAVIS_COMMIT_RANGE` anymore, so the checks should run locally. And the check is skipped if `optipng` is not installed.

_contrib/travis-check-png-optimized.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+
+## Finds newly added or modified PNG files in the branch and checks if each file is optimized.
+## If not optimized, it fails with an error code of 1.
+
+## This check might be run on a system where optipng is not istalled (e.g. Netlify). If so, the check should not fail.
+if ! which optipng > /dev/null; then
+	echo "Skipping check since optipng is not istalled."
```
harding: Spelling nit for "istalled"

harding:
Tested ACK for 84e48dc6932ff2137f2a976900d2bfc08aa0f763 (with a spelling nit).  Locally tested with and without optipng installed and with and without unoptimized images in git history; worked as expected.

0xB10C:
Corrected spelling.

jnewbery:
Thanks @0xB10C (and @harding for review)

jnewbery:
#200 is merged. Can this be closed?

PR #201, opened 2019-08-14
Compatibility data: bitcoin core wallet text standardization

bitschmidty:
This PR updates the Bitcoin Core wallet data files with some of the standard prefixes that we agreed upon since initial data file commits.


harding:
Untested ACK.

jnewbery:
ACK. Thanks @bitschmidty

PR #206, opened 2019-08-19
Compat matrix: add links to subsections

harding:
We previously added links to the segwit and RBF subsections on the individual details page, but now that the number of rows in the matrix is growing, I think we should have links to the subsections on the main matrix page too:


bitschmidty:
tACK. Good idea again @harding !

jnewbery:
ACK

PR #211, opened 2019-08-20
Fill out untested gaps for Wasabi

nopara73:


harding:
@MaxHillebrand thank you for the review and especially for the link to the issue with the Wasabi RBF discussion---it's valuable to us RBF advocates to learn why people are reluctant to use RBF and there are several points in that discussion that I think are outdated or which could be addressed by data we can fairly easily collect.  I'll comment over on that issue later today.  Thanks!

_data/compatibility/wasabi.yaml
```diff
@@ -10,17 +10,17 @@ rbf:
     version: "1.0"
   features:
     receive:
-      notification: "na"
+      notification: "false"
```
bitschmidty: Notification field being `na` means that the wallet/service does not have notification for incoming transactions.
Notification field being `false` means that the wallet/service does have a notification for incoming transactions but that the notification doesn’t show if the tx is signaling RBF.

That said, I just tested again and I did not get a notification. So I believe `na` is correct for this field.

_data/compatibility/wasabi.yaml
```diff
@@ -10,17 +10,17 @@ rbf:
...
       details: "false"
       shows_replaced_version: "true"
       shows_original_version: "false"
     send:
       signals_bip125: "false"
-      list: "untested"
-      details: "untested"
-      shows_replaced_version: "untested"
-      shows_original_version: "untested"
+      list: "false"
```
bitschmidty: If sending an RBF signaling transaction isnt possible from Wasabi, there is no way to tell what a "rbf-signaling-sent-from-wasabi" transaction would look like in the list/details pages. That is why, in these scenarios we have the fields as untested. Let me know if that is unclear.

nopara73: One can test this by importing the seed to Electrum and sending RBFs and see that the tx is properly replaced.

_data/compatibility/wasabi.yaml
nopara73: > Notification field being na means that the wallet/service does not have notification for incoming transactions.
> Notification field being false means that the wallet/service does have a notification for incoming transactions but that the notification doesn’t show if the tx is signaling RBF.

The wallet have notifications on OSX and Linux, but not on Windows. Anyhow, it's going to be refactored to in-wallet (non-os-native) notifications for all platforms. At this point however we can think about it as a partially implemented feature and such feature should not be considered implemented, so `na` should work, so I misunderstood it. I change it back. Will update to `false` after the native notifications are implemented.  https://github.com/zkSNACKs/WalletWasabi/pull/2119#issuecomment-521191190

bitschmidty:
@nopara73 @MaxHillebrand I pushed a commit to explain how the testing was done. Also updated the tested on date. Let me know if it looks good to you guys!

nopara73:
> Let me know if it looks good to you guys!

LGTM

jnewbery:
I've squashed commits. ACK 5416eb9c1ecdfcc57a260cb3dadc794b51d2d229

PR #224, opened 2019-09-10
Update ruby version & re-activate schema checker

harding:
This upgrades the version of Ruby used by the site-building tools to 2.6.4, which is listed as the latest stable version on https://www.ruby-lang.org/en/downloads/ .  It also re-enables the schema checker that was part of #168 but removed in #174 due to an incompatibility between the Gemfile and the Netlify preview and deployment scripts.  (I still don't know what the specific problem was, but upgrading Ruby fixes it in my tests.)

If you've previously installed Ruby (e.g. per our README.md), you will need to install the latest version of Ruby.  If you followed the README's previous install instructions to use rvm, you can do this with a single command:

    rvm install ruby-2.6.4

After it's installed, rvm will automatically select the correct ruby when you enter the repository top-level directory (if you're already in the directory when the ruby version changes, you can use `cd .` to switch).

This change in ruby versions should not produce any meaningful content differences in the site, so the testing procedure I recommend is:

```bash
rvm install ruby-2.6.4
git blah blah ## checkout this branch
cd . ## use correct ruby (new ruby version)
make  ## build site and run tests
cp -a _site _new_site ## backup new site content
git reset --hard HEAD^^ ## throw away this PR's changes
cd . ## use correct ruby (old ruby version)
make
diff -ru _site _new_site | colordiff | less -R  ## see what's changed
```

For me, the new ruby caused a bunch of JSON associative arrays to have a different sort order, which is annoying to scan through but not a functional change.  All other content was unaffected.

Note: normally I'd upgrade Ruby in a separate PR from adding dependencies (the revert commit) but I think it's better to do both together here to ensure upgrading Ruby truly eliminates the dependency problem.

My thanks to @schmidty for making the Netlify build logs public.


Gemfile.lock
```diff
@@ -101,11 +106,13 @@ PLATFORMS
...
   minima (~> 2.0)
+  toml
 
 RUBY VERSION
-   ruby 2.5.1p57
+   ruby 2.6.4p104
 
 BUNDLED WITH
-   1.16.6
+   1.17.3
```
jonatack: I'd suggest upgrading bundler as well:
```
~/projects/bitcoin/bitcoinops.github.io (pr/224)$ gem install bundler
Fetching bundler-2.0.2.gem
Successfully installed bundler-2.0.2
Parsing documentation for bundler-2.0.2
Installing ri documentation for bundler-2.0.2
Done installing documentation for bundler after 2 seconds
1 gem installed
```

jonatack:
For info, Ruby 2.7.0 will be released around Christmas time as per the Ruby major release tradition.

harding:
Removed `make test` instructions from the README.  (The tests were recently split into two parts, those that can be run before site build and those that can be run after.  Both parts are automatically run in the correct sequence by `make` without a specific target.)

Gemfile.lock
harding: I'd prefer to leave that for later.  This PR resolves a problem we had, but subsequent to it getting merged, I plan to PR an update to the latest version of Jekyll that will update almost all of our dependencies.  To me, that seems like the correct time to update bundler.

jonatack: Bundler is the dependencies resolver/package manager, so I tend to update it before changing/upgrading dependencies. If helpful, I'm on the latest and had no issues with the deps when bundle installing, but as long as you're not seeing any issues then either way should be fine.

jnewbery:
tACK f8eff5a3aaacfbb14be921de090b1f4419e05d38

PR #228, opened 2019-09-23
Fix NODE_ prefix in 2019-09-18-newsletter.md

MarcoFalke:
Remove NODE_ prefix from network services in RPC response


harding:
ACK per https://github.com/bitcoin/bitcoin/pull/16787/files#diff-a2c210f0004fc59cf7e73a29076dd7a8R743

Thanks!

PR #232, opened 2019-10-05
Newsletters: add 67 (2019-10-09)

harding:
- ~~[ ] The paragraph starting "We at Optech are saddened to hear these sentiments" is meant to be presented as an Optech group opinion, so please mention if you don't think it accurately reflects your thinking about building on schnorr and taproot, or you think it's otherwise inappropriate for the newsletter.~~  Edit: removed
- [x] Update RC links (FIXMEs) near merge time

### DIFF
--- a/_posts/en/newsletters/2019-02-19-newsletter.md
+++ b/_posts/en/newsletters/2019-02-19-newsletter.md
@@ -21,8 +21,8 @@ None this week.
 
 ## News
 
-- **Discussion about tagging outputs to enable restricted features on
-  spending:** The [BIP118][] SIGHASH_NOINPUT_UNSAFE (noinput) proposal
+- **Discussion about tagging outputs to enable restricted features on spending:**
+  The [BIP118][] SIGHASH_NOINPUT_UNSAFE (noinput) proposal
   allows the person generating a signature that authorizes the spend of
   one UTXO to optionally allow that signature to be reused ("replayed")
   for spending other UTXOs sent to the same public key.  This enables
--- a/_posts/en/newsletters/2019-10-09-newsletter.md
+++ b/_posts/en/newsletters/2019-10-09-newsletter.md
@@ -1,0 +1,172 @@
+---
+title: 'Bitcoin Optech Newsletter #67'
+permalink: /en/newsletters/2019/10/09/
+name: 2019-10-09-newsletter
+slug: 2019-10-09-newsletter
+type: newsletter
+layout: newsletter
+lang: en
+---
+This week's newsletter requests help testing release candidates for
+Bitcoin Core and LND, tracks continued discussion about the proposed
+noinput and anyprevout sighash flags, and describes several notable
+changes to popular Bitcoin infrastructure projects.
+
+{% comment %}<!-- include references.md below the fold but above any Jekyll/Liquid variables-->{% endcomment %}
+{% include references.md %}
+
+## Action items
+
+- **Help test Bitcoin Core 0.19.0rc1:** production users of Bitcoin Core
+  are especially encouraged to test this [latest release candidate][core
+  0.19.0] to ensure that it fulfills all of your organization's needs.
+  Experienced users who plan to test are also asked to take a few
+  moments to test the GUI and look for problems that might
+  disproportionately affect less-experienced users who don't normally
+  participate in RC testing.
+
+- **Help test LND 0.8.0-beta-rc2:** experienced users of LND are
+  encouraged to [help test][lnd 0.8.0-beta] the next release.  For the
+  first time ever, this testing can include creating a [reproducible
+  build][lnd repo build] of LND and verifying that it has the same hash
+  as the binaries distributed by the LND developers.
+
+
+## News
+
+- **Continued discussion about noinput/anyprevout:** this proposed
+  sighash flag that would allow LN implementations to use [eltoo][] was
+  [discussed][noinput thread] again on the Bitcoin-dev and Lightning-dev mailing lists.
+  After summarizing previous discussions, Christian Decker asked several
+  questions: is the idea behind the proposal useful?  (Respondents seemed
+  to agree that it was.)  Do people want mandatory chaperon signatures?
+  (Respondents seemed moderately opposed.)  Do people want mandatory output
+  tagging?  (Respondents seemed opposed, some strongly.)
+
+    In response to the question about output tagging, C-Lightning
+    contributor ZmnSCPxj [proposed][zmn internal tagging] an alternative
+    tagging mechanism that would put the tag inside the taproot
+    commitment, making it invisible unless a script-path spend was used.
+    This could allow a spender who was worried about noinput to ensure
+    they didn't pay noinput-compatible scripts---the [original goal][orig
+    output tagging] behind output tagging---but without the decrease in
+    privacy and fungibility created by output tagging.  Several people
+    seemed to express interest in this idea, although it wasn't clear
+    whether they wanted to see it as part of a proposal or they just
+    preferred it to external output tagging (which, as noted above, was
+    generally opposed by respondents).
+
+    The entire thread is more than 20 messages at present and started a
+    [spin-off discussion about `OP_CAT`][cat spinoff].  Hopefully the
+    discussion will be able to settle the major unresolved issues
+    related to noinput and help get this proposal on track for
+    inclusion in a subsequent soft fork.
+
+## Notable code and documentation changes
+
+*Notable changes this week in [Bitcoin Core][bitcoin core repo],
+[LND][lnd repo], [C-Lightning][c-lightning repo], [Eclair][eclair repo],
+[libsecp256k1][libsecp256k1 repo], [Bitcoin Improvement Proposals
+(BIPs)][bips repo], and [Lightning BOLTs][bolts repo].*
+
+- [Bitcoin Core #13716][] adds `-stdinrpcpass` and
+  `-stdinwalletpassphrase` parameters to
+  `bitcoin-cli` that allow it to read either an RPC or wallet
+  passphrase from the standard input buffer rather than as a CLI
+  parameter that would be stored in shell history.  Echoing is also
+  disabled on stdin during reading so that the passphrase isn't visible
+  to anyone watching your screen.
+
+- [Bitcoin Core #16884][] switches the default address type
+  for users of the RPC interface (including via `bitcoin-cli`) from
+  P2SH-wrapped P2WPKH to native segwit (bech32) P2WPKH.  This change is on the
+  master development code branch and is not expected to be released
+  until Bitcoin Core 0.20.0 sometime in mid-2020.  However, a [previous
+  change][gui bech32] expected to be released as part of 0.19.0 in the
+  next month or so will switch the default address type for GUI users to
+  also use bech32 P2WPKH.
+
+- [Bitcoin Core #16507][] fixes a [rounding issue][bitcoin core #16499] where a
+  node would accept transactions into its mempool if they had a
+  feerate greater than the node's dynamic minimum feerate but wouldn't
+  relay those transactions to peers if the transactions' feerates were
+  less than minimum rounded up to next 0.00001000 BTC.
+
+- [LND #3545][] adds code and [documentation][lnd repo doc] that allows
+  users to create reproducible builds of LND.  This should allow anyone with
+  moderate technical skills to build identical binaries to those
+  released by Lightning Labs, ensuring that users are running the
+  peer-reviewed code from the LND repository and its dependencies.
+
+- [LND #3365][] adds support for using `option_static_remotekey`
+  commitment outputs as described [later in this section][opt static
+  remotekey].
+  This new commitment protocol is particularly useful when something has
+  gone wrong and you've lost data.  If that happens, you need only wait
+  for your channel counterparty to close the channel by paying a key
+  directly derived from your HD wallet.  Because the key was generated
+  without any additional data ("tweaking"), your wallet doesn't need any
+  extra data in order to find and spend your funds.  This is a simplified
+  alternative to the [data loss protection][] protocol that LND
+  previously used and continues to understand.
+
+- [C-Lightning #3078][] adds support for creating and using channels
+  that spend Liquid-BTC on the Liquid sidechain.
+
+- [C-Lightning #2803][] adds a new python package named `pyln` that
+  includes a partial implementation of the LN specification.  As
+  described in its [documentation][pyln-proto readme], "This package
+  implements some of the Lightning Network protocol in pure python. It
+  is intended for protocol testing and some minor tooling only. It is
+  not deemed secure enough to handle any amount of real funds (you have
+  been warned!)."
+
+- [C-Lightning #3062][] causes the `plugin` command to return an error if a
+  requested plugin hasn't reported successful startup within 20 seconds.
+
+- [BOLTs #676][] amends [BOLT2][] to specify that a node should not send
+  the `funding_locked` message until it has validated the funding
+  transaction.  This warns future implementers about the problem that
+  lead to the vulnerabilities [described in last week's newsletter][ln
+  vuln disclosure].
+
+- [BOLTs #642][] allows two peers opening a channel to negotiate an
+  `option_static_remotekey` flag.  If both peers set this flag, any
+  commitment transactions they create which they're able to spend
+  unilaterally (e.g. to force close the channel) must pay their peer's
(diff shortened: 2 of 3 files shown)


JeremyRubin:
@moneyball asked me to chime in here. While this is your opinion, I think your take is mostly wrong and I don't think you should be dismayed. First I'd like to correct the quote from my talk. What I actually said (slightly cleaned up) is:

    We could do OP_SECURETHEBAG as a Tapscript Extension that's how I originally proposed it,
    but given that that there may be a long schedule before taproot is ready, and this can be done as
    a normal script extension, it might make sense to just do this as a standard OP_NOP upgrade to 
    get it in a little bit sooner. But that's a question for the community.

The critical part of this being that it's a question for the community to decide the prioritization of upgrades, if any.


    We at Optech are saddened to hear these sentiments.  Fewer projects
    with a stake in proposed protocol enhancements likely means that those
    proposals will receive less testing and have less motivation for
    adoption,

Then what you're asking for is for one feature/project to serve as a development subsidy for another.

    further slowing their deployment in Bitcoin.

And speeding up the deployment of other exciting features.

    Rather than investigating alternatives to schnorr and taproot,
    
Note that it's not really an alternative to, because the implementation & BIP is not yet stable enough to build features on top of.  

    we encourage anyone upset at the slow pace of their development to investigate
    how you might contribute to getting them implemented in Bitcoin more
    quickly.

I'm not upset at the slow pace at all -- engineering something correctly takes time. The feature that I developed doesn't really even benefit from schnorr/taproot directly -- it's just that Tapscript has an easier path to adding new opcodes so *should* make it slightly more simple to do. But it turns out this isn't the case because to do an OpCode upgrade in tapscript for OP_SECURETHEBAG, it turns out we also probably want to re-write the entire way that scripts are parsed to permit multibyte opcodes.

Further using Taproot for OP_SECURETHEBAG has an additional validation cost and uses more data (causing more congestion and costing users more in fees), so the benefits are not worth the costs.

Furthermore, it's important not to make changes a "pill wrapped in cheese" (how you trick a dog to take meds). Changes should stand on their own, and not need a falsely constructed carrot. This is why we're sensitive to not make soft-fork upgrades during major releases, because you should be able to accept or reject a soft fork proposal whilst otherwise staying on the latest software.

Yes, Taproot is taking a while. If users want to see a congestion control feature available in the near term, their best bet at engendering that is as a separate change not as a post-taproot feature.


Lastly there seems to be a slight insinuation (intended or not) that developers aren't doing work to review taproot. I've extensively reviewed taproot and even refactored it to make it easier for other developers to review -- https://github.com/sipa/bitcoin/pull/116 . Rather than exhort developers to use their time in a manner more consistent with your priorities, it might be worth explaining what you actually think the obstacles are for moving it forward for taproot, it is likely more effective at attracting developer interest. I also didn't get the memo for a taproot meetup/workshop -- but from a review of Kanzure's notes, it doesn't seem that much was discussed in terms of deployment/getting it out to the world. Was there a discussion of deployment or what needs to be done to launch that did not make the notes?

I've heard various folks (who would be in the know) say that the implementation available for Taproot publicly is just an "example" and to not pay too much heed to it, because there would be a lot more to do before it would be ready. But I don't personally know what that checklist is or what the known deficiencies are. Are you guys tracking them?

harding:
@JeremyRubin Thank you for taking the time to write a detailed comment.  I've decided to remove this section as it does come across as more critical of you and Zmn than I intended (sorry).  I've privately heard some frustrations from schnorr/taproot contributors with the low amount of contribution they've received from other parts of the community and I worry about them interpreting changes such as your revised BIP and comments such as Zmn's 2p-ECDSA note as a general abandonment of their work.

Some replies to your comments:

> Then what you're asking for is for one feature/project to serve as a development subsidy for another.

I'm saying that there's often a synergy created when we work with together with other developers rather than independently from them.  It's often the case that insights from a dependent system can feed back into improvements in the higher-level system and so produce a better overall result for both systems than we'd achieve through independent development.  I thought you were achieving some of that with your ideas for templated taproot with STB.

> Changes should stand on their own, and not need a falsely constructed carrot. This is why we're sensitive to not make soft-fork upgrades during major releases, because you should be able to accept or reject a soft fork proposal whilst otherwise staying on the latest software.

We shouldn't be changing Bitcoin for the sake of change alone---changes need motivation.  I agree that if STB is better without taproot, or even if it's roughly equivalent either way, we shouldn't tie their fates together.  But if STB can be better with taproot (which still seems likely to me), I think we're better off doing things in the order that will result in the greatest payoff for the fewest changes.  I think we get a huge confidence boost by using later possible designs to prove the usefulness of earlier proposed designs.

> Was there a discussion of deployment or what needs to be done to launch that did not make the notes?

Not that I'm aware of (I didn't attend the sessions myself).

> But I don't personally know what that checklist is or what the known deficiencies are. Are you guys tracking them?

I'm not; I don't know if any other Optech contributors are doing so.

Again, thank you for your detailed reply.

JeremyRubin:
> I'm saying that there's often a synergy created when we work with together with other developers rather than independently from them. It's often the case that insights from a dependent system can feed back into improvements in the higher-level system and so produce a better overall result for both systems than we'd achieve through independent development. I thought you were achieving some of that with your ideas for templated taproot with STB.

Agreed -- love to see synergy. The feedback I received on templated taproot was that it would be worth investigating as an option (because of how it would work, can't quite be a soft fork) well after taproot has been largely adopted and we're looking for a second upgrade to get, i.e., graftroot or other incompatible changes (maybe a 3-5 year timeline). If that's the case, I don't think it makes sense to completely hitch OP_SECURETHEBAG onto that.

> We shouldn't be changing Bitcoin for the sake of change alone---changes need motivation. I agree that if STB is better without taproot, or even if it's roughly equivalent either way, we shouldn't tie their fates together. But if STB can be better with taproot (which still seems likely to me), I think we're better off doing things in the order that will result in the greatest payoff for the fewest changes. I think we get a huge confidence boost by using later possible designs to prove the usefulness of earlier proposed designs.

The main benefit of doing OP_SECURETHEBAG with taproot is that you can look up an arbitrary branch more efficiently compared to O(N) OP_IF branches if you're doing a more complicated script. This is a benefit that could be found with MAST as well, it's not strictly speaking unique to Taproot. The delaying/non-consideration of MAST is a result of similar thinking -- we can make this just a part of Taproot, so it's good enough/better. Perhaps this is in the spirit of the one change with the greatest payoff, but I'd rather see more regular progress and availability of features rather than never-ending pursuit of perfection. Further, as we try to cram everything together into one, changes take longer. Taproot could be just taproot without MAST, and you would be able to use a MAST script with Taproot to achieve the same effect, and the review and audit efforts could be separate. Composeability also keeps individual components simpler, whereas the spec for taproot is already quite complex.

_posts/en/newsletters/2019-10-09-newsletter.md
```diff
@@ -0,0 +1,176 @@
...
+    funds onchain in an address where your HD wallet will find them.
+
+{% include linkers/issues.md issues="13716,16884,16507,16499,3545,3365,3078,2803,3062,676,642" %}
+[lnd repo doc]: https://github.com/lightningnetwork/lnd/blob/master/build/release/README.md
+[core 0.19.0]: https://bitcoincore.org/bin/bitcoin-core-0.19.0/ {% comment %}<!-- FIXME: broken, binaries not uploaded yet -->{% endcomment %}
+[lnd 0.8.0-beta]: https://github.com/lightningnetwork/lnd/releases/tag/v0.8.0-beta-rc2 {% comment %}<!-- FIXME: check for more recent RC -->{% endcomment %}
+[lnd repo build]: #lnd-3545
+[noinput thread]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-September/002176.html
+[cat spinoff]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-October/002201.html
+[coshv]: http://devbcop/en/newsletters/2019/05/29/#proposed-transaction-output-commitments
```
bitschmidty: This and the next 4 links can be removed

_posts/en/newsletters/2019-10-09-newsletter.md
```diff
@@ -0,0 +1,176 @@
...
+{% comment %}<!-- include references.md below the fold but above any Jekyll/Liquid variables-->{% endcomment %}
+{% include references.md %}
+
+## Action items
+
+- **Help test Bitcoin Core 0.19.0RC1:** production users of Bitcoin Core
+  are especially encouraged to test this [latest release candidate][core
+  0.19.0] to ensure that it fulfills all of your organization's needs.
+  Experienced users who plan to test are also asked to take a few
+  moments to test the GUI and look for problems that might
```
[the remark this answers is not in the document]
harding: This is something that we've asked before (e.g. https://bitcoinops.org/en/newsletters/2018/12/11/#help-test-bitcoin-core-0-17-1rc1 ).  I can't quickly find a reference, but it stems from a an issue in the, I think, 0.17.0 release where there was a GUI issue that was missed for a long time during the RC cycle, possibly because we only encourage advanced users and sysadmins to test and many of them don't use the GUI.  There's no special motivation for GUI testing this release, just the generic motivation explained in the current text.

_posts/en/newsletters/2019-10-09-newsletter.md
```diff
@@ -0,0 +1,171 @@
...
+Bitcoin Core and LND, tracks continued discussion about the proposed
+noinput and anyprevout sighash flags, and describes several notable
+changes to popular Bitcoin infrastructure projects.
+
+{% comment %}<!-- include references.md below the fold but above any Jekyll/Liquid variables-->{% endcomment %}
+{% include references.md %}
+
+## Action items
+
+- **Help test Bitcoin Core 0.19.0RC1:** production users of Bitcoin Core
```
jonatack: FWIW the release is tagged as `v0.19.0rc1      Bitcoin Core 0.19.0 release candidate 1` (`git tag -n | sort -V`)

_posts/en/newsletters/2019-10-09-newsletter.md
```diff
@@ -0,0 +1,171 @@
...
+- [BOLTs #642][] allows two peers opening a channel to negotiate an
+  `option_static_remotekey` flag.  If both peers set this flag, any
+  commitment transactions they create which they're able to spend
+  unilaterally (e.g. to force close the channel) must pay their peer's
+  funds to a static address negotiated during the initial channel open.
+  For example, if Alice and Bob both request `option_static_remotekey`,
+  any commitment transactions that Alice can publish onchain must pay
+  bc1bob[...] and any commitment transactions that Bob can publish
+  onchain must pay bc1ally[...].  If at least one of them doesn't set
+  this flag, they'll fallback to the older protocol of using a different
```
jonatack: Unsure but iirc "fallback" is a noun and "fall back" is the verb, e.g. "we'll fall back to the fallback plan", similar to "we'll set up the setup"

_posts/en/newsletters/2019-10-09-newsletter.md
```diff
@@ -0,0 +1,171 @@
...
+  to anyone watching your screen.
+
+- [Bitcoin Core #16884][] switches the default address type
+  for users of the RPC interface (including via `bitcoin-cli`) from
+  P2SH-wrapped P2WPKH to native segwit (bech32) P2WPKH.  This is on the
+  master development code branch and is not expected to be released
+  until Bitcoin Core 0.20.0 sometime in mid-2020.  However, a [previous
+  change][gui bech32] expected to be released as part of 0.19.0 in the
+  next month or so will switch the default address type for GUI users to
+  also use bech32 P2WPKH.
```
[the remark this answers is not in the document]
harding: I think it's a recognition that it's easier to deal with in the GUI.  If you go to the withdrawal page on Foo exchange and they don't accept your bech32 address, it's easy to click the button in the GUI that gives you a backwards-compatible address.  In the RPC, you probably have to run the `help` to figure out what the option is to get a one-time P2SH-wrapped address or you need to run `-help` to find the startup option and then restart your server to always get P2SH-wrapped addresses.

Opt-in RBF is in the same situation.  It's default on in the GUI but not default on in the RPC (last I checked).

_posts/en/newsletters/2019-10-09-newsletter.md
```diff
@@ -0,0 +1,171 @@
...
+
+- [Bitcoin Core #16507][] fixes an [issue][bitcoin core #16499] where a
```
instagibbs: I cannot see @jnewbery comment here, but from e-mail:

> I don't think this is exactly correct. The behaviour before this change was that any feefilter value that wasn't a multiple of 1000 could cause an inconsistency between the mempool acceptance policy (which would allow transaction to enter the mempool if it had a fee rate > than minrelaytxfee) and the relay policy (which would round the feerate down and therefore not relay if the fee rate < (minrelaytxfee rounded up to next 1000)).

specifically it was because the feefilter was using one calculation method(which made the fee seem unusable to the peer) and not sent, due to this `%1000 != 0` issue

_posts/en/newsletters/2019-10-09-newsletter.md
```diff
@@ -0,0 +1,171 @@
...
+  vuln disclosure].
+
+- [BOLTs #642][] allows two peers opening a channel to negotiate an
+  `option_static_remotekey` flag.  If both peers set this flag, any
+  commitment transactions they create which they're able to spend
+  unilaterally (e.g. to force close the channel) must pay their peer's
+  funds to a static address negotiated during the initial channel open.
+  For example, if Alice and Bob both request `option_static_remotekey`,
+  any commitment transactions that Alice can publish onchain must pay
+  bc1bob[...] and any commitment transactions that Bob can publish
```
jnewbery: I think these made up addresses are confusing unless you introduce them earlier, eg: "if Alice and Bob both request `option_static_remotekey`, with Alice requesting that Bob pays funds to `bc1ally[...]` and Bob requesting that Alice pays funds to `bc1bob[...]`, then any commitment transaction that ...."

_posts/en/newsletters/2019-10-09-newsletter.md
instagibbs: it seems I have achieved time travel

_posts/en/newsletters/2019-10-09-newsletter.md
```diff
@@ -0,0 +1,171 @@
...
+  encouraged to [help test][lnd 0.8.0-beta] the next release.  For the
+  first time ever, this testing can include creating a [reproducible
+  build][lnd repo build] of LND and verifying that it has the same hash
+  as the binaries distributed by the LND developers.
+
+
+## News
+
+- **Continued discussion about noinput/anyprevout:** this proposed
+  sighash flag that would make possible the [eltoo][] layer for LN was
```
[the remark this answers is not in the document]
jonatack: "Enable" came to mind here for me as well.

_posts/en/newsletters/2019-02-19-newsletter.md
```diff
@@ -21,8 +21,8 @@ None this week.
 
 ## News
 
-- **Discussion about tagging outputs to enable restricted features on
-  spending:** The [BIP118][] SIGHASH_NOINPUT_UNSAFE (noinput) proposal
+- **Discussion about tagging outputs to enable restricted features on spending:**
```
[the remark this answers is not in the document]
harding: git grep -- '- \*\*[^\*]*$'
    git grep -- '- \[[^]]*$'

harding:
Pushed edits for feedback from @jonatack, @jnewbery, and @instagibbs (thanks!).

@bitschmidty I commented out the 0.19.0rc1 announcement as those binaries aren't available yet and I'm starting to doubt that they'll be up by press time tomorrow.  However, I left in the link so it should be easy for you to uncomment at merge time if you see they're available.  I also left in the FIXME on the LND link, although it's current as of this comment.  You can find LND's latest RC at https://github.com/LightningNetwork/lnd/releases and you can get a direct link to it by clicking its title.

jnewbery:
ACK ccaa3d364af4abf3dd77b4dc3d82ddfd51a3d18b.

Looks great. Thanks Harding!

bitschmidty:
ACK ccaa3d364af4abf3dd77b4dc3d82ddfd51a3d18b

bitschmidty:
Bitcoin Core and LND drop the latest RC
Name is ZmnSCPxj but just call em Z 

Regarding noinput and anyprevout
Causing some discussion there is no doubt

Core GUI getting default bech32
In a little while the RPC too

Thanks to @instagibbs @JeremyRubin @jonatack and @jnewbery for all the reviewing 
and of course on lead guitar our man Dave @harding

PR #235, opened 2019-10-10
Style: make all h1s centered

harding:
Suggested [here](https://github.com/bitcoinops/bitcoinops.github.io/pull/167#discussion_r333611118)

I tested by quickly clicking around a site preview and looking at various page titles to make sure they looked reasonably good.  The only page I didn't really like was About, I think because it's title is so short.  I think it's ok to just live with that slight ugliness in order to get nice, centered titles on all other pages.

### DIFF
--- a/_layouts/newsletter.html
+++ b/_layouts/newsletter.html
@@ -2,7 +2,6 @@
 layout: default
 --- 
 <link rel="stylesheet" href="/assets/css/main.css">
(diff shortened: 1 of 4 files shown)


jnewbery:
ACK e31eebca300f895ec640dfc5cb743c9801cbbd01

I've had a quick click around and I definitely prefer this. Thanks!

PR #240, opened 2019-10-21
Fix bullet titles and add anchors

harding:
This PR makes the newsletter content pass the tests described [here](https://github.com/bitcoinops/bitcoinops.github.io/pull/232#discussion_r332799560) and then adds those tests to the makefile.  By putting those bullet titles all on one line, they're automatically turned into anchors by a plugin added in a previous PR.

In a final commit, several anchors are manually added.  Content in the not-yet-added topics branches refers to those anchors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,10 @@ test-before-build:
 	! git --no-pager grep -L "^slug: " _posts
 	## Check that all slugs are unique
 	! git --no-pager grep -h "^slug: " _posts | sort | uniq -d | grep .
+	## Check for things that should probably all be on one line
+	@ ## Note: double $$ in a makefile produces a single literal $
+	! git --no-pager grep -- '^ *- \*\*[^\*]*$$'
+	! git --no-pager grep -- '^ *- \[[^]]*$$'
 
 	## Check that newly added or modifyed PNGs are optimized
 	_contrib/travis-check-png-optimized.sh
--- a/_posts/en/newsletters/2018-10-30-newsletter.md
+++ b/_posts/en/newsletters/2018-10-30-newsletter.md
@@ -103,8 +103,8 @@ answers made since our last update.*
   by over 97% at present, but do they also speed up sync?  Bitcoin Core
   developer Gregory Maxwell answers.
 
-- [Can someone steal from you by closing their Lightning Network payment
-  channel in a certain way?][bse 80399] Several different ways to close
+- [Can someone steal from you by closing their Lightning Network payment channel in a certain way?][bse 80399]
+  Several different ways to close
   a Lightning Network payment channel are described, and C-Lightning
   developer Christian Decker explains how a program following the LN
   protocol will protect your money in each case.
--- a/_posts/en/newsletters/2018-10-30-newsletter.md
+++ b/_posts/en/newsletters/2018-10-30-newsletter.md
@@ -216,6 +216,7 @@ it to write a good description, and I doubt non-LN devs care -->{% endcomment %}
   specification for watchtowers has not been agreed upon by the multiple
   implementations of LN, so LND is only putting this feature out for
   initial testing and is restricting its use to testnet.
+  {:#lnd-1535-1512}
 
 {% include references.md %}
 {% include linkers/issues.md issues="14451,14296,14468,14150,1981,1535,1512" %}
--- a/_posts/en/newsletters/2018-12-28-newsletter.md
+++ b/_posts/en/newsletters/2018-12-28-newsletter.md
@@ -44,6 +44,7 @@ to single-sig transactions.
 
 </div>
 
+{:#taproot}
 Building on the idea that muSig, or something like it, could become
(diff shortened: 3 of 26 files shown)


Makefile
```diff
@@ -21,6 +21,10 @@ test-before-build:
 	! git --no-pager grep -L "^slug: " _posts
 	## Check that all slugs are unique
 	! git --no-pager grep -h "^slug: " _posts | sort | uniq -d | grep .
+	## Check for things that should probably all be on one line
+	@ ## Note: double $$ in a makefile produces a single literal $
+	! git --no-pager grep -- '^ *- \*\*[^\*]*$$'
```
jnewbery: I don't think the backslash is necessary in the square brackets:

```
→ echo '*' | grep [*]; echo $?
*
0
→ echo '*' | grep [^*]; echo $?
1
```

I've verified that removing the `\` yields the same result when run against master.

_posts/en/newsletters/2018-12-28-newsletter.md
```diff
@@ -325,6 +331,7 @@ recommended) using Tor---which can provide other benefits---but enabling
 encryption by default could help protect a larger number of users from
 eavesdropping by their ISPs.
 
+{:#untrackable-auth}
```
jnewbery: Any reason not to use 'countersign' as the anchor here?

_posts/en/newsletters/2018-10-30-newsletter.md
```diff
@@ -205,6 +205,7 @@ it to write a good description, and I doubt non-LN devs care -->{% endcomment %}
 - [LND #1981][] ensures that LND doesn't leak information about any of
   its peers that aren't advertising themselves as public nodes.
 
+{:#lnd-1535-1512}
```
jnewbery: This splits the next bullet into its own list. Would it be better to just make `[LND #1535]` the id for the bullet?

_posts/en/newsletters/2019-06-12-newsletter.md
```diff
@@ -38,10 +38,10 @@ changes to popular Bitcoin infrastructure projects.
       addresses, tracking when they've been paid, and finding or
       deriving the particular private keys necessary for spending.
 
-    - June 6th: discussion of the [consensus cleanup soft fork][xs
+    - <span id="cleanup-discussion" markdown="1">June 6th: discussion of the [consensus cleanup soft fork][xs
```
jnewbery: How about:

```
    - 
      {:#cleanup-discussion}
      June 6th: discussion of the [consensus cleanup soft fork][xs
```

Same for `assume-utxo-demo` below.

bitschmidty: FWIW I couldn't get what John outlined to both work and pass the "things that should be on one line" check at the same.

harding: @bitschmidty yeah, I had the same problems when trying different variations.  However, FYI, Block Attribute Lists (BALs) can go either before or after the block you want them to act on, so I was able to put it after the first paragraph of that bullet:

```text
    - June 6th: discussion of the [consensus cleanup soft fork][xs
      cleanup sf] included its interaction with [bip-taproot][], whether
      parts of it should be dropped, and whether anything should be
      added.  ([More background][bg consensus cleanup])
      {:#cleanup-discussion}
```

harding:
Forced pushed edits for @jnewbery feedback (thanks!)

_posts/en/newsletters/2018-10-30-newsletter.md
```diff
@@ -205,7 +205,8 @@ it to write a good description, and I doubt non-LN devs care -->{% endcomment %}
 - [LND #1981][] ensures that LND doesn't leak information about any of
   its peers that aren't advertising themselves as public nodes.
 
-- LND [#1535][LND #1535] and [#1512][LND #1512] adds the server-side
+- {:#lnd-1535-1512}
```
jnewbery: This is still moving this one bullet point into its own `<ul>`. Can you move this Block Attribute List to after the text?

jnewbery: actually, I've just done this myself and pushed

harding: I think that edit is fine, but are you sure you were testing the updated branch?  On my branch (ede049d564559e65471b41c0027d7a1b20c939c9, without your change), I don't see an extraneous `<ul>`:

```html
  </li>
  <li id="lnd-1535-1512">
    <p>LND <a href="https://github.com/lightningnetwork/lnd/issues/1535">#1535</a> and <a href="https://github.com/lightningnetwork/lnd/issues/1512">#1512</a> adds the server-side
```

harding: Yeah, looking at the GitHub forced push, it looks like you merged the old version of this with your one change.

PR #243, opened 2019-10-25
Topics: add new topics (dongcarl)

harding:
Eight proposed pages for the new `/en/topics/` section of the site.  Suggested reviewer is @dongcarl 

Note: tests expected to fail on the final "production" test because of the FIXME link pointing to a topic that's part of a different PR.  That's the last test we run, so if it's the only failure, this should be ok.

PR #246, opened 2019-10-25
Topics: add new topics (jnewbery)

harding:
Eight proposed pages for the new `/en/topics/` section of the site.  Suggested reviewer is @jnewbery 

Note: tests expected to fail on the final "production" test because of the FIXME link pointing to a topic that's part of a different PR.  That's the last test we run, so if it's the only failure, this should be ok.

### DIFF
--- a/_topics/en/compact-block-filters.md
+++ b/_topics/en/compact-block-filters.md
@@ -1,0 +1,78 @@
+---
+title: Compact block filters
+
+## Optional.  An entry will be added to the topics index for each alias
+aliases:
+  - BIP157
+  - BIP158
+  - Neutrino protocol
+
+## Required.  At least one category to which this topic belongs.  See
+## schema for options
+categories:
+  - Privacy Enhancements
+  - Lightweight Client Support
+  - P2P Network Protocol
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+excerpt: >
+  **Compact block filters** are a condensed representation of the
+  contents of a block that allow wallets to determine whether the block
+  contains any transactions involving the user's keys.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  A full node uses [BIP158][] to create a Golomb-Rice Coded Set (GCS) of the
+  data from each block in the block chain. The GCSs (called filters) are then
+  distributed to wallets (such as via the P2P protocol as described in
+  [BIP157][]), allowing them to search for any matches to their scripts.  If a
+  match is found, the wallet can then download the corresponding block to access
+  any relevant transactions.
+
+    The GCS mechanism guarantees that a wallet following the protocol will find
+    any transactions matching its scripts, but it may also find some
+    false-positive matches for which it will need to download and scan
+    the block despite the block not containing any transactions relevant
+    to the wallet.
+
+    The BIP157/158 protocol is sometimes incorrectly called "Neutrino" after the
+    wallet library developed to use the protocol.  It's one of several
+    methods that lightweight clients can use to acquire data about their
+    wallet transactions.  Compared to [BIP37][] bloom filters, it offers
+    more privacy against spy nodes and less risk of attack against
+    honest nodes.  Compared to address-indexed servers (such as
+    Electrum-style servers), it also provides more privacy and requires
+    less server storage and CPU.  However, the BIP157/158 does consume
+    significantly more bandwidth in the normal case than either of those
+    other protocols.
+
+## Optional.  Produces a Markdown link with either "[title][]" or
+## "[title](link)"
+primary_sources:
+    - title: BIP157
+    - title: BIP158
+
+## Optional.  Each entry requires "title", "url", and "date".  May also use "feature:
+## true" to bold entry
+optech_mentions:
+  - title: Discussion of what data should be included in BIP158 filters
+    url: /en/newsletters/2018/06/08#bip157-bip157-bip158-bip158-lightweight-client-filters
+    date: 2018-06-08
+
+  - title: Functions for generating BIP158 filters added to Bitcoin Core
+    url: /en/newsletters/2018/08/28/#bitcoin-core-12254
+    date: 2018-08-28
+
+  - title: Basic BIP158 support merged into Bitcoin Core
+    url: /en/newsletters/2019/04/23/#basic-bip158-support-merged-in-bitcoin-core
+    date: 2019-04-23
+
+  - title: BIP157 bandwidth higher than BIP37 bloom filters
+    url: /en/newsletters/2019/07/31/#bip157-would-use-more-bandwidth-than-bip37
+    date: 2019-07-31
+
+## Optional.  Same format as "primary_sources" above
+see_also:
+  - title: BIP37 bloom filters
+    link: BIP37
+---
--- a/_topics/en/countersign.md
+++ b/_topics/en/countersign.md
@@ -1,0 +1,50 @@
+---
+title: Countersign
+
+## Required.  At least one category to which this topic belongs.  See
+## schema for options
+categories:
+  - P2P Network Protocol
+  - Security Enhancements
+  - Privacy Enhancements
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+excerpt: >
+  **Countersign** is the idea for a protocol that will allow a client and
+  server who have each other's public keys to negotiate authentication
+  without either participant revealing any identifying information to third parties.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  This would make it easier to securely set up whitelisted nodes across
+  the Internet for miners or exchanges, or to allow lightweight wallets
+  to ensure they connect to trusted nodes.  By enabling authentication
+  without revealing identity to third parties, nodes on anonymity
+  networks (such as Tor) or nodes that simply changed IP addresses
+  couldn't have their network identity tracked.
+
+  The protocol should be compatible with the [version 2 P2P transport
+  protocol][topic v2 p2p transport].
+
+## Optional.  Produces a Markdown link with either "[title][]" or
+## "[title](link)"
+primary_sources:
+    - title: Countersign
+      link: https://gist.github.com/sipa/d7dcaae0419f10e5be0270fada84c20b
+
+## Optional.  Each entry requires "title", "url", and "date".  May also use "feature:
+## true" to bold entry
+optech_mentions:
+  - title: "2018 year-in-review: untrackable authentication"
+    url: /en/newsletters/2018/12/28/#countersign
+    date: 2018-12-28
+
+  - title: "CoreDev.tech meetings: v2 P2P transport and countersign"
+    url: /en/newsletters/2019/06/12/#v2-p2p
+    date: 2019-06-12
+
+## Optional.  Same format as "primary_sources" above
+see_also:
+  - title: Version 2 P2P Transport
+    link: topic v2 p2p transport
+---
--- a/_topics/en/erlay.md
+++ b/_topics/en/erlay.md
@@ -1,0 +1,56 @@
+---
+title: Erlay
+
+## Required.  At least one category to which this topic belongs.  See
+## schema for options
+categories:
+  - P2P Network Protocol
+  - Bandwidth Reduction
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+excerpt: >
+  **Erlay** is proposal to improve the bandwidth efficiency of relaying
+  unconfirmed transactions between Bitcoin full nodes.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  In the currently-used Bitcoin gossip protocol, most full nodes are
+  configured to advertise every new transaction to all of their peers
+  unless they've previously received an advertisement about the
+  transaction from that peer.  At a minimum of 32 bytes per
+  advertised txid and nodes having a default maximum of 125 peers, this
+  consumes a large amount of redundant bandwidth given that each node
+  only needs to learn about a transaction from one of its peers.
+
+  Erlay is a two-part proposal that first limits the number of peers to
+  which a node will directly advertise transactions (default: 8) and,
+  second, uses set reconciliation based on [libminisketch][] with the
+  remainder of its peers to avoid sending the txid of any transactions
+  that the receiving peer has already seen.
+
+  Erlay scales to larger numbers of peers much better than the current
+  protocol, making it practical for nodes to accept more connections
+  than they do now.  This would improve the robustness of the relay
+  network against both accidental and deliberate network partitions.
+
+## Optional.  Produces a Markdown link with either "[title][]" or
+## "[title](link)"
+primary_sources:
+    - title: Erlay
+
+## Optional.  Each entry requires "title", "url", and "date".  May also use "feature:
+## true" to bold entry
+optech_mentions:
+  - title: Erlay proposed
+    url: /en/newsletters/2019/06/05/#erlay-proposed
+    date: 2019-06-05
+
+  - title: Draft BIP for enabling Erlay compatibility
+    url: /en/newsletters/2019/10/02/#draft-bip-for-enabling-erlay-compatibility
+    date: 2019-10-02
+
+## Optional.  Same format as "primary_sources" above
+see_also:
+  - title: Minisketch
+    link: topic minisketch
+---
--- a/_topics/en/replace-by-fee.md
+++ b/_topics/en/replace-by-fee.md
@@ -1,0 +1,83 @@
+---
+title: Replace-by-Fee
+
+aliases:
+  - BIP125
+  - Opt-in Replace-by-Fee
+
+## Required.  At least one category to which this topic belongs.  See
+## schema for options
+categories:
+  - Fee Management
+  - Transaction Relay Policy
+  - Mining
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+excerpt: >
+  **Replace by Fee (RBF)** is a node policy that allows an unconfirmed
+  transaction in a mempool to be replaced with a different transaction
+  that spends at least one of the same inputs and which pays a higher
+  transaction fee.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  Different node software can use different RBF rules, so there have
+  been several variations.  The most widely-used form of RBF today is
+  [BIP125][] opt-in RBF as
+  implemented in [Bitcoin Core 0.12.0][core12 rbf] and subsequent
+  versions; this allows the creator of a transaction to signal that
+  they're willing to allow it to be replaced by a higher-paying version.
+  An alternative form of RBF is full-RBF that allows any transaction to
+  be replaced whether or not it signals BIP125 replacability.
+
+  BIP125 requires a replacement transaction to pay both higher feerate
+  (BTC/vbyte) and a higher absolute fee (total BTC).  This can make
+  multiparty transactions that want to use RBF vulnerable to
+  [transaction pinning][] attacks, and so an occasional discussion
+  topic is proposals to allow RBF to operate solely on a feerate basis.
+
+  [core12 rbf]: https://bitcoincore.org/en/releases/0.12.0/#opt-in-replace-by-fee-transactions
+  [transaction pinning]: https://bitcoin.stackexchange.com/questions/80803/what-is-meant-by-transaction-pinning
+
+## Optional.  Produces a Markdown link with either "[title][]" or
+## "[title](link)"
+primary_sources:
+    - title: BIP125
+    - title: "Bitcoin Core PR #6871: nSequence-based Full-RBF opt-in"
+      link: https://github.com/bitcoin/bitcoin/pull/6871
+
+## Optional.  Each entry requires "title", "url", and "date".  May also use "feature:
+## true" to bold entry
+optech_mentions:
+  - title: Compatibility matrix---Replace by Fee
+    url: /en/compatibility#replace-by-fee-rbf
+    date: 2019-08-20
+
+  - title: RBF in the wild (survey of RBF usage)
+    url: /en/rbf-in-the-wild/
+    date: 2019-02-11
+
+  - title: "2018 year-in-review: transaction statistics"
+    url: /en/newsletters/2018/12/28/#opt-in-rbf
+    date: 2018-12-28
+
+  - title: Proposal to override some BIP125 RBF conditions
+    url: /en/newsletters/2019/06/12/#proposal-to-override-some-bip125-rbf-conditions
+    date: 2019-06-12
+
+  - title: LND adds support for RBF fee bumping
+    url: /en/newsletters/2019/06/19/#lnd-3140
+    date: 2019-06-19
+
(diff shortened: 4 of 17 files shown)


_topics/en/countersign.md
```diff
@@ -0,0 +1,49 @@
+---
+title: Countersign
+
+## Required.  At least one category to which this topic belongs.  See
+## schema for options
+categories:
```
jnewbery: Could also be considered a *Privacy Enhancement* since this would allow a client to try to connect to a server without revealing its identity.

_topics/en/erlay.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Required.  At least one category to which this topic belongs.  See
+## schema for options
+categories:
+  - P2P Network Protocol
+  - Bandwidth Reduction
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+excerpt: >
+  **Erlay** is proposal to use set reconciliation between Bitcoin full
```
jnewbery: I think it would be more accurate to say "...is a proposal to relay unconfirmed transactions in two phases: a low-fanout flooding phase and a set reconciliation phase."

_topics/en/erlay.md
```diff
@@ -0,0 +1,51 @@
...
+  nodes as part of relaying unconfirmed transactions.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  The currently-used Bitcoin gossip protocol expects each node to
+  advertise every new transaction to all of its peers.  However, peers
+  only need to learn about the transaction from any one of their peers,
+  so a significant amount of bandwidth is wasted on redundant
+  announcements.  Set reconciliation using [libminisketch][] would
```
jnewbery: As above, I think we should mention that erlay has two phases: fanout and reconciliation.

_topics/en/erlay.md
```diff
@@ -0,0 +1,51 @@
...
+  - Bandwidth Reduction
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+excerpt: >
+  **Erlay** is proposal to use set reconciliation between Bitcoin full
+  nodes as part of relaying unconfirmed transactions.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  The currently-used Bitcoin gossip protocol expects each node to
```
jnewbery: 'expects' seems like an odd word to me to use here, since it implies that a node has an expectation that all of its peers will relay a transaction to it.

_topics/en/erlay.md
```diff
@@ -0,0 +1,51 @@
...
+  advertise every new transaction to all of its peers.  However, peers
+  only need to learn about the transaction from any one of their peers,
+  so a significant amount of bandwidth is wasted on redundant
+  announcements.  Set reconciliation using [libminisketch][] would
+  allow a node and its peer to avoid sending any information about
+  transactions the both already knew about.
```
jnewbery: s/the both/that both/

jnewbery: perhaps 'that both already know about'?

_topics/en/erlay.md
```diff
@@ -0,0 +1,51 @@
...
+
+  Erlay also scales to larger numbers of peers much better than the
+  current protocol, allowing the possibility of having nodes connect to
+  larger numbers of peers.  This would improve the robustness of the
```
jnewbery: perhaps remove repeated 'larger numbers of peers' in this sentence?

_topics/en/transaction-pinning.md
```diff
@@ -0,0 +1,83 @@
...
+  management more difficult in multipart contract protocols (such as LN).
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  Nodes such as Bitcoin Core that allow transactions to be replaced
+  (RBF) or packaged with higher-fee child transactions (CPFP) place
+  restrictions on those replacements in order to prevent various DoS
+  attacks.  However, when two or more people each have the ability to
+  fee bump a transaction, this makes it possible for one of them to
+  pin their version of a transaction at one of the limits and prevent
```
jnewbery: Since you're introducing the term _pin_ for the first time here, perhaps use _italics_ or **bold**

_topics/en/transaction-pinning.md
```diff
@@ -0,0 +1,83 @@
...
+  other participants from using fee bumping.
+
+  Some of the limits that can be abused to enable transaction pinning
+  include:
+
+  - **[BIP125][] RBF rule #3** requires a replacement transaction
+    pay a higher absolute fee (not just feerate) than the transaction
+    being replaced and any of its children.  This can allow an attacker
+    to attach a large and low-feerate transaction to the transaction
+    they want to pin, forcing any fee bump to pay for pay for the
```
jnewbery: s/pay for pay for/pay for/

_topics/en/transaction-pinning.md
```diff
@@ -0,0 +1,83 @@
...
+    replacement of the large child transaction.  E.g., with the 2019
+    Bitcoin Core defaults, an attacker can require an honest participant
+    pay a minimum of 0.002 BTC to fee bump a transaction (or even
+    greater amounts in some cases).  <!-- 0.002 BTC = 100,000 vbyte tx
```
jnewbery: Ha. I was about to leave a review comment here saying 'I haven't checked this number - perhaps leave an HTML comment with your working', but you already have! :stars:

_topics/en/transaction-pinning.md
```diff
@@ -0,0 +1,83 @@
...
+
+  - **Maximum package size limitations** prevent CPFP from being used if
+    a transaction has more than 101,000 vbytes of children or other
+    descendants in a mempool, or has more than 25 descendants or
+    ancestors.  This can allow an attacker to completely block fee
+    bumping by creating the maximum amount of child transactions.  If
+    the attacker has to create those transactions for other reasons
+    (e.g. because they operate a service paying to users), this attack
+    can be free.  For some two-party contract protocols (such as current
+    LN), this may be mitigated by CPFP carve out.
```
jnewbery: Make 'CPFP carve out' a link? (I know there's a link below, but linking from here seems friendly)

_topics/en/transaction-pinning.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Optional.  Produces a Markdown link with either "[title][]" or
+## "[title](link)"
+primary_sources:
+    - title: What is meant by transaction pinning?
```
jnewbery: I think this should be filed under 'see also' rather than 'primary sources'. I'm sure I'm not the primary source for transaction pinning!

jnewbery: edit: I'm not sure if this link is even needed, since you (indirectly) link to it in the optech mentions section.

_topics/en/countersign.md
jnewbery: Done

_topics/en/erlay.md
jnewbery: Done

_topics/en/erlay.md
jnewbery: Done

_topics/en/erlay.md
jnewbery: done

_topics/en/erlay.md
jnewbery: reworded

_topics/en/erlay.md
jnewbery: done

_topics/en/transaction-pinning.md
jnewbery: done

_topics/en/transaction-pinning.md
jnewbery: done

_topics/en/transaction-pinning.md
jnewbery: done

_topics/en/transaction-pinning.md
jnewbery: done

_topics/en/compact-block-filters.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+excerpt: >
+  **Compact block filters** are a condensed representation of the
+  contents of a block that allow wallets to determine whether the block
+  contains any transactions affecting the wallet.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  A full node with a copy of a block and the data necessary to validate
```
jnewbery: I think we can just say 'a full node' here and assume that it has the block and the data necessary to validate the block.

_topics/en/compact-block-filters.md
```diff
@@ -0,0 +1,76 @@
...
+  that block uses [BIP158][] to create a Golomb-Rice Coded Set (GCS) of
+  data from that block (such as the prevout script from each input spent
```
jnewbery: What do you mean by "prevout scipt"? I haven't come across that terminology before.

_topics/en/compact-block-filters.md
```diff
@@ -0,0 +1,76 @@
...
+  in the block and the scriptPubKey for each output created in the
+  block).  The GCS (called a filter) is then distributed to wallets
+  (such as via the P2P protocol as described in [BIP157][]), allowing
+  them to search for any matches to each wallet's scripts.  If a match
```
jnewbery: perhaps 'their scripts' to avoid repetition of 'wallet'

_topics/en/compact-block-filters.md
```diff
@@ -0,0 +1,76 @@
...
+  is found, the wallet can then download the corresponding block to
+  access any relevant transactions.
+
+    The GCS mechanism guarantees that a properly-coded wallet will find
```
jnewbery: remove 'properly-coded' here. It's confusing, especially in the context of something called 'Golomb-Rice Coding'

_topics/en/compact-block-filters.md
```diff
@@ -0,0 +1,76 @@
...
+    any transactions matching its scripts, but it may also find some
+    false-positive matches for which it will need to download and scan
+    the block despite the block not containing any transactions relevant
+    to the wallet.
+
+    The BIP157/158 protocol is sometimes called "Neutrino" after the
```
jnewbery: Perhaps "is sometimes (incorrectly) called ..."

_topics/en/compact-block-filters.md
```diff
@@ -0,0 +1,76 @@
...
+    wallet library developed to use the protocol.  It serves as an
+    alternative to lightweight clients using [BIP37][] bloom filters
```
jnewbery: The sentence construction is slightly confusing to parse here - the 'that' clause at the end would naturally attach itself to 'It' rather than 'lightweigh clients using...' if it weren't for the verb disagreement. For example: "It is an alternative to BIP 37 that provides less privacy".

Consider: "It serves as an alternative to lightweight clients using [BIP37][] bloom filters (that provide less privacy and which create a DoS attack vector against full nodes)." or "It is an alternative to the earlier [BIP37][] bloom filters lightweight client mechanism. Bloom filters provide less privacy and create a DoS attack vector against full nodes."

_topics/en/compact-block-filters.md
```diff
@@ -0,0 +1,76 @@
...
+primary_sources:
+    - title: BIP157
+    - title: BIP158
+
+## Optional.  Each entry requires "title", "url", and "date".  May also use "feature:
+## true" to bold entry
+optech_mentions:
+  - title: Discussion of what data should be included in BIP158 filters
+    url: /en/newsletters/2018/06/08#bip157-bip157-bip158-bip158-lightweight-client-filters
+    date: 2019-06-08
```
jnewbery: s/2019/2018

also: Newsletter 0! :heart_eyes:

_topics/en/countersign.md
```diff
@@ -0,0 +1,50 @@
...
+  - P2P Network Protocol
+  - Security Enhancements
+  - Privacy Enhancements
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+excerpt: >
+  **Countersign** is the idea for a protocol that will allow a client and
+  server who have each others public keys to negotiate authentication
```
jnewbery: each other's

_topics/en/countersign.md
```diff
@@ -0,0 +1,50 @@
...
+  without either revealing any identifying information to third parties.
```
jnewbery: 'without either party revealing any identifying information.' (without a noun, either attaches to the verb revealing on initial parse)

_topics/en/compact-block-filters.md
harding: I'm not completely opposed to that, but I think it's helpful to mention those conditions to help explain why pruned nodes can't create filters for blocks where they've already pruned the block and undo data.

_topics/en/compact-block-filters.md
harding: prevout = previous output; script = scriptPubKey.  So the encumbrance that the input's scriptSig or witness fulfills.  Grepping the Bitcoin Core source, I find four instances of "prevout scriptPubKey" (all in the tests), so maybe I should change it to say that?

_topics/en/compact-block-filters.md
harding: Is it incorrect?  I don't like calling it neutrino so I'd like to be able to say with authority that it's incorrect, but the closest I've heard any of the authors of the proposal say to that effect is Osuntokun with:

> I'd wager that most developers reading this email right now are familiar with neutrino as a project. Many even routinely use "neutrino" to refer to BIP 157+158.  ([source](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2019-February/016656.html))

_topics/en/compact-block-filters.md
jnewbery: That feels a bit too in the weeds for me, but happy to go either way.

_topics/en/compact-block-filters.md
jnewbery: I think 'scriptPubKey of every output spent in the block and scriptPubKey of every new output created in the block' (or something similar) is clearer.

_topics/en/compact-block-filters.md
jnewbery: I'm pretty sure it's incorrect. Laolu seems to be drawing a clear distinction between Neutrino (implementation) and BIP157/158 (protocol) in that email.

If you didn't want to use the word "incorrectly", I think I'd be happier with 'is sometimes referred to as "Neutrino" after...' or rather than 'is sometimes called "Neutrino" after...'. It might just be me, but the first one sounds a bit more like "Neutrino" is an accepted alternative name and the second sounds like it's simply reporting on what some people say. "Is referred to by some as "Neutrino" after..." would be taking it even further.

_topics/en/compact-block-filters.md
harding: I think "scriptPubKey of every output spent in the block" is ambiguous.  It could refer (correctly) to the prevout scriptPubKeys or it could refer to a subset of that being just the scriptPubKeys of transactions in the block that were spent by later transactions also included in the same block.  I think I'll just remove the whole parenthetical; anyone who wants to know precisely what data is included can read BIP158 or our other links.

_topics/en/compact-block-filters.md
harding: As long as you're convinced, that means I can blame you when it all goes wrong.  :-)  I'll add "incorrectly".

_topics/en/replace-by-fee.md
```diff
@@ -0,0 +1,80 @@
...
+## Required.  At least one category to which this topic belongs.  See
+## schema for options
+categories:
+  - Fee Management
+  - Transaction Relay Policy
+  - Mining
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+excerpt: >
+  **Replace by Fee (RBF)** is the ability to replace an unconfirmed
```
jnewbery: 'ability' doesn't seem quite right here. It makes me think that this sentence is written from the perspective of a transaction signer who is _able_ to replace their transaction. I think instead, we want this to be from the perspective of a node operator (as the second paragraph clearly is). Perhaps something like:

"**Replace by Fee (RBF)** is a node policy allowing an unconfirmed transaction in a mempool to be replaced with a different transaction that spends at least one of the same inputs and which pays a higher transaction fee."

_topics/en/replace-by-fee.md
```diff
@@ -0,0 +1,80 @@
...
+  transaction in a mempool with a different transaction that spends at
+  least one of the same inputs and which pays a higher transaction fee.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  RBF is a policy for mining and relaying transactions, so it's
```
jnewbery: s/mining and relaying/relaying and mining/ since relaying comes first (and happens more often than mining)

_topics/en/replace-by-fee.md
```diff
@@ -0,0 +1,80 @@
...
+  something that each node can choose to implement however it would
```
jnewbery: pedantic nit: nodes don't have preferences, node operators do :)

_topics/en/replace-by-fee.md
```diff
@@ -0,0 +1,80 @@
...
+  multiparty transactions that want to use RBF vulnerable to
+  [transaction pinning][] attacks, and so an occasional discussion
+  topic is proposals to allow RBF to operate solely on a feerate basis.
+
+  [core12 rbf]: https://bitcoincore.org/en/releases/0.12.0/#opt-in-replace-by-fee-transactions
+  [transaction pinning]: https://bitcoin.stackexchange.com/questions/80803/what-is-meant-by-transaction-pinning
+
+## Optional.  Produces a Markdown link with either "[title][]" or
+## "[title](link)"
+primary_sources:
```
jnewbery: Add implementation PR: https://github.com/bitcoin/bitcoin/pull/6871 ?

_topics/en/v2-p2p-transport.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Optional.  Produces a Markdown link with either "[title][]" or
+## "[title](link)"
+primary_sources:
+    - title: v2 Transport Protocol proposed BIP
+      link: https://gist.github.com/jonasschnelli/c530ea8421b8d0e80c51486325587c52
+
+## Optional.  Each entry requires "title", "url", and "date".  May also use "feature:
+## true" to bold entry
+optech_mentions:
```
jnewbery: Perhaps add /en/newsletters/2019/06/12/#v2-p2p (Schnelli's presentation of v2 p2p at the Core dev meetup)

_topics/en/compact-block-filters.md
harding: Um, since the earliest version of that documentation, there has been a todo:

```
<!-- SOMEDAY: we need more terms than just output/input to denote the
various ways the outputs/inputs are used, such as "prevout", "nextout",
"curout", "curin", "nextin".  (Is there any use for "previn"?)  Someday,
when I'm terribly bored, I should rewrite this whole transaction section
to use those terms and then get feedback to see if it actually helps. -harding -->
```

You can consider my subsequent adoption of the term `prevout` in later documentation as work towards solving that terminology problem I first encountered in early 2014.  I'm hardly alone in that, the term is being used in many other pieces of technical documentation, including BIP68 sequence numbers, BIP143 segwit v0 signature algorithm, BIP174 PSBTs, bip-taproot, and (perhaps most obviously) bip-anyprevout.

In this case, it doesn't matter---I'm just deleting this entire parenthetical---but we'd have to have a longer conversation before you get me to give up on the term altogether as I think it's an convenient way to express a frequently-used (and particularly-import) part of validation data.

jnewbery: For future reference, `prevout` in those documents refers to the _outpoint_ (the hash and txOut index of the previous transaction output), and not the transaction output itself.

_topics/en/compact-block-filters.md
jnewbery: I think this is less clear than the original. "some related data" just begs the question "which data?" I've reverted to your original version.

jnewbery:
ACK bd7556d

Great stuff @harding . Thanks for putting up with all my pedantry :)

jnewbery:
This is ready for merge once the _minisketch_ topic is merged (#243)

_topics/en/compact-block-filters.md
```diff
@@ -22,11 +22,13 @@ excerpt: >
...
-  block.  The GCS (called a filter) is then distributed to wallets using either
-  the P2P protocol as described in [BIP157][] or some other mechanism, allowing
-  the wallet to search for any matches to their scripts.  If a match
+  A full node with a copy of a block and the data necessary to validate
+  that block uses [BIP158][] to create a Golomb-Rice Coded Set (GCS) of
+  data from that block (such as the prevout scriptPubKey from each input spent
+  in the block and the scriptPubKey for each output created in the
+  block).  The GCS (called a filter) is then distributed to wallets
+  (such as via the P2P protocol as described in [BIP157][]), allowing
+  them to search for any matches to their scripts.  If a match
```
harding: @jnewbery this commit confuses me; why did you restore the *prevout* terminology after arguing against it?

jnewbery:
> why did you restore the prevout terminology after arguing against it?

Oops. I just wanted to revert the change I was talking about here: https://github.com/bitcoinops/bitcoinops.github.io/pull/246#discussion_r343844751, and carelessly also re-added the reference to prevout.

I think the wording about 'the data necessary to validate the block' or 'some related data' is just confusing and unnecessary for the following reasons:

- A full node creates the GCS from the block data and undo data (see https://github.com/bitcoin/bitcoin/blob/master/src/blockfilter.cpp#L258). I wouldn't refer to undo data as 'data necessary to validate a block' so much as 'data necessary to rewind a block'
- If the index is sync'ed, then the full node obviously has the block data and undo data to construct the GCS.
- if the index is not sync'ed and wants to construct the GCS for an old block, then either:
    - the node has the historic block data and undo data and can construct the GCS
    - the node has pruned both the block data and undo data and can't construct the GCS (we prune block data and undo data for the same block at the same time)

The fact that a full node can construct a block filter index asynchronously from old data is really an implementation detail. I don't think it's important for understanding the concept.

I've added another commit that removes mention of 'additional data'. Let me know what you think.

harding:
Your edit 76bb411ea9055eee588aa9ec00033bd3b9a59aa8 LGTM.  Thanks.  To nitpick your message a little:

> I wouldn't refer to undo data as 'data necessary to validate a block' so much as 'data necessary to rewind a block'

Before the data is placed in the undo file, it's part of the set of data that's necessary to validate a block.  After it's in the undo file, it's part of what's used to rewind a block.  So I think it's fair to describe it either way.

> The fact that a full node can construct a block filter index asynchronously from old data is really an implementation detail. I don't think it's important for understanding the concept.

I don't think I mentioned asynchronicity.  I thought it was useful to mention the both data sources because otherwise it's hard for someone already familiar with the protocol to understand how data from just a single block can allow a wallet "to search for any matches to their scripts", since inputs don't necessary identify their scripts (e.g. the input spending a P2PK or bare multisig will just be a random-looking signature or set of signatures).

I'm sure you're right about the description getting into the weeds, which is why I'm fine with this removal, but I wanted to explain why I thought it was useful to note.

jnewbery:
This is ready for merge once _minisketch_ (#243) is merged

jnewbery:
Squashed fixup commits and cherry-picked the commits from #243.

PR #257, opened 2019-11-06
Add Bitcoin Wallet to SegWit table.

schildbach:
The source code can be reviewed at
https://github.com/bitcoin-wallet/bitcoin-wallet


schildbach:
The errors don't seem to be related to my PR.

harding:
The errors seem to be because your data file doesn't contain the RBF section.

@bitschmidty did we want to require every wallet have data in every section, or did we want to allow wallets to be part of (e.g.) the segwit section but not the RBF section?  In either case, we can update our code---for the former case, I can ensure we provide a more descriptive failure message; for the later, I can stick some if statements in there to skip attempting to generate empty entries.

schildbach:
Just for clarification: we (the Bitcoin Wallet devs) favor CPFP over RBF, so it doesn't really make sense to add a row in the RBF table.

bitschmidty:
@schildbach Thanks for your contribution, much appreciated!

A couple things here:

1. Contributing instructions were not clear in outlining that you need markdown file for each wallet as well. I have pushed a commit to your PR here (hope that is ok!) with that file. Additionally I have opened a [PR to clarify that requirement](https://github.com/bitcoinops/bitcoinops.github.io/pull/259).
2. The template for listing all wallets didn’t have a check for missing RBF data. (there was such a check for segwit). I have opened a [PR to add this conditional](https://github.com/bitcoinops/bitcoinops.github.io/pull/260).
3. The template for the wallet details page was missing a similar RBF data check (while also having the segwit data check). The PR mentioned above also resolves this.

Once the PR for the data checks is merged, we can look at merging this PR.

One thing you might consider adding, but is completely optional, are screenshots showing any applicable screenshots regarding segwit support. If you’re curious you can see other wallet pages for example screenshots.

Additionally, even though Bitcoin Wallet does not support RBF, that is still useful information to have in the matrix for people to be aware of. You could consider adding the RBF data to the compatibility file even if it rounds to "we dont support it".

PR #261, opened 2019-11-07
[newsletters] Add newsletter indexes for other languages

jnewbery:
Initially only japanese, but the machinery is in place for any other languages.

Differences between `/ja/newsletters/` and `/en/newsletters/`:

- add a note pointing would-be contributors where to go if they want to help
- remove newsletter sign-up and rss link (since both are for english language newsletters)


bitschmidty:
Concept ACK!

bitschmidty:
Tested 62d0a00 and it works as expected. 

Only thing I see missing are updates to the CONTRIBUTING.md regarding translations.

jnewbery:
> Only thing I see missing are updates to the CONTRIBUTING.md regarding translations.

Are you able to do that? You've walked CG through the process so i think you're in a better position to write it. No rush - we don't need this until we have a few more translations.

ja/newsletters.md
```diff
@@ -0,0 +1,16 @@
...
+share: false
+version: 1
+---
+
+Would you like to help translate our newsletters? See the [CONTRIBUTING
+documentation](https://github.com/bitcoinops/bitcoinops.github.io/blob/master/CONTRIBUTING.md)
+and the [Japanese translation issues and
+PRs](https://github.com/bitcoinops/bitcoinops.github.io/pulls?&q=label%3Alocalization-japanese)
+in our github repo.
+{:.center}
```
harding: Not a blocker, but I think this might look better in italics (and possibly not centered).

_layouts/newsletters.md
```diff
@@ -0,0 +1,45 @@
...
+
+<h1 class="post-title">Newsletters</h1>
+
+{%- if page.lang == "en" -%}
+  {% include newsletter-signup.html %}
+{%- endif -%}
+
+<div class="post-content">
+  {{ content }}
+</div>
```
harding: Nitpick: the English page is devoid of content (which is fine), so this renders an empty div on that page.  That doesn't really matter, but it looks odd to me.  Maybe consider gating this with an `{% if content != ""%}`

_layouts/newsletters.md
```diff
@@ -0,0 +1,45 @@
...
+  {% for lang in site.languages %}
+    | <a href="/{{ lang }}/newsletters/">{{lang}}</a>
+  {% endfor %}
+</div>
+
+<h1 class="post-title">Newsletters</h1>
+
+{%- if page.lang == "en" -%}
+  {% include newsletter-signup.html %}
+{%- endif -%}
```
harding: This is not strictly necessary, but I think you want to wrap the signup form in a div, either in this file or in `_includes/newsletter-signup.html`.  Otherwise, the raw HTML is reprocessed by Kramdon and produces some slight changes.  I'm no HTML expert, but I don't think any of those changes should produce any differences, so I'm ok with this being left as-is.

```diff
-<form action="https://bitcoinops.us18.list-manage.com/subscribe/post?u=70c3f85e5d13ffec674f30af8&amp;id=adc93c9774" method="post" id="mc-embedded-subscribe-form" name="mc-embedded-subscribe-form" class="validate" target="_blank" novalidate="">
+<form action="https://bitcoinops.us18.list-manage.com/subscribe/post?u=70c3f85e5d13ffec674f30af8&amp;id=adc93c9774" method="post" id="mc-embedded-subscribe-form" name="mc-embedded-subscribe-form" class="validate" target="_blank" novalidate>
     <div id="mc_embed_signup_scroll">
        <label for="mce-EMAIL">Subscribe to our newsletter</label>
-       <input type="email" value="" name="EMAIL" class="email" id="mce-EMAIL" placeholder="email address" required="" />
+       <input type="email" value="" name="EMAIL" class="email" id="mce-EMAIL" placeholder="email address" required>
     <!-- real people should not fill this in and expect good things - do not remove this or risk form bot signups-->
-    <div class="mc-input" aria-hidden="true"><input type="text" name="b_70c3f85e5d13ffec674f30af8_adc93c9774" tabindex="-1" value="" /></div>
-    <div class="clear"><input type="submit" value="Subscribe" name="subscribe" id="mc-embedded-subscribe" class="button" /></div>
+    <div class="mc-input" aria-hidden="true"><input type="text" name="b_70c3f85e5d13ffec674f30af8_adc93c9774" tabindex="-1" value=""></div>
+    <div class="clear"><input type="submit" value="Subscribe" name="subscribe" id="mc-embedded-subscribe" class="button"></div>
     </div>
```

ja/newsletters.md
jnewbery: done!

_layouts/newsletters.md
jnewbery: done

_layouts/newsletters.md
jnewbery: I tried adding `<div>`s but that didn't seem to prevent these changes. In any case, I've tested and the changes don't seem to have any difference in the functionality, so I'm going to leave this as-is.

bitschmidty:
tACK 90d44040108861958f650f43e9e3dd4ad82a5243

PR #263, opened 2019-11-08
fix misspelling

jlopp:


bitschmidty:
ACK

Good catch @jlopp

jnewbery:
Welcome to the party @jlopp !

PR #270, opened 2019-11-14
spell error

haegwan:
Found a spell error during JP translation. I know this is already published so just FYI (doesn't have to be incorporated).


jnewbery:
Thanks for your contribution @haegwan ! I'm very happy to take minor corrections to past newsletters.

PR #278, opened 2019-11-26
Taproot-WS: add Japanese translation

Ichiro0219:


bitschmidty:
ACK

@Ichiro0219 Thanks for translating this blog post. I think it is an important one!

(John, James, and Elichai with Japanese subtitles... some day :)

jnewbery:
Thank you @Ichiro0219 !

PR #280, opened 2019-11-27
Internationalize publications indexes

jnewbery:
CG have contributed a Japanese translation of the schnorr/taproot in #278. Make Japanese versions of the _blog_ index and _publications_ index.

@harding : there's a bit of repetitive code in these PRs. If we add more languages, I'll need to add additional _newsletters_, _blog_ and _publications_ files for each of them. I think it's ok, but if you think we should make this more automated and have suggestions for how to do that, let me know.


harding:
Not tested, but looks fine to me.

bitschmidty:
Post merge tACK

Opened some related issues: https://github.com/bitcoinops/bitcoinops.github.io/issues/285

PR #283, opened 2019-12-01
Topics: combine related entries on date index

harding:
Instead of showing the same entry for each topic it belongs to, e.g.:

    - Tapscript—Bitcoin Optech schnorr/taproot workshop
    - Taproot—Bitcoin Optech schnorr/taproot workshop
    - Schnorr signatures—Bitcoin Optech schnorr/taproot workshop

This PR combines related entries into a single item, e.g.:

    - Bitcoin Optech schnorr/taproot workshop
      Tapscript, Taproot, Schnorr signatures

I think it's already the case that the date index looks a little bit silly with duplicated entries.  As we increase the number of topics, I expect more entries to be part of multiple topics and so duplication will increase.  I think the solution in this PR will help keep the date index concise and non-ridiculous looking.

The solution combines entries that have the same date and the same URL, using the title of the last-seen entry as the title of the combined entries.  I think that's fine---our current descriptions seem to be generic enough that they can apply to each affected topic.  However, if anyone thinks we need the safety of requiring all mentions being combined use the same title, that would be easy to add, e.g. `if (x.title != y.title) abort_build`

The associated topics are printed separated by commas on the line following the entry title.  I tried several formats; this adds a fair amount of whitespace but looked best to me---but I'm happy to take suggestions on alternative formats.

Screenshot of topic index before this change (15 entries for October):

Screenshot after this change (9 entries for October but all the same topics linked):

### DIFF
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -247,3 +247,7 @@ div.localization
 {
   text-align: right;
 }
+
+.pc125 {
+  font-size: 125%;
+}
--- a/en/topic-dates.md
+++ b/en/topic-dates.md
@@ -11,23 +11,43 @@ produce code blocks -->{% endcomment %}
 {% capture raw_mentions %}
 {%- for topic in site.topics -%}
   {%- for mention in topic.optech_mentions -%}
-    {{mention.date}} <a href="{{topic.url}}">{{topic.title}}</a>&#8212;{{mention.title}}&nbsp;<a href="{{mention.url}}">🔗</a>ENDMENTION
+    {{mention.date}}DIVIDER<a class="pc125" href="{{mention.url}}">🔗</a>DIVIDER{{mention.title}}DIVIDER<a href="{{topic.url}}">{{topic.title}}</a>ENDMENTION
   {%- endfor -%}
 {%- endfor -%}
 {% endcapture %}
 {% assign mentions = raw_mentions | split: 'ENDMENTION' | sort | reverse %}
 
-{% for mention in mentions %}
-  {% assign mention_date = mention | truncate: 10, '' %}
-  {% assign year_month = mention | truncate: 7, '' %}
-  {% if year_month != lastym %}
+{%- for mention in mentions -%}
+  {%- assign mention_part = mention | split: "DIVIDER" -%}
+  {%- assign mention_date = mention_part[0] -%}
+  {%- assign mention_link = mention_part[1] -%}
+  {%- assign mention_title = mention_part[2] -%}
+  {%- assign mention_topic = mention_part[3] -%}
+
+  {%- if mention_link == old_mention_link -%}
+    {%- assign combined_topics = combined_topics | append: ', ' | append: mention_topic -%}
+    {%- capture item -%}<li><p>{{mention_title}}&nbsp;{{mention_link}}<br>{{combined_topics}}</p></li>{%- endcapture -%}
+  {%- else -%}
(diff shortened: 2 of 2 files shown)


jnewbery:
Concept ACK. A couple of questions/suggestions:

- why is the title of the mention plain text with a link icon afterwards for the link? Why not make the text a link?
- how about having the topics written after the link in parentheses to save on vertical space, similar to what I'm proposing for review club here: https://deploy-preview-79--bitcoincore-reviews.netlify.com/meetings/.

harding:
@jnewbery 

> why is the title of the mention plain text with a link icon afterwards for the link? Why not make the text a link?

It turns the page into a wall of blue, which I find hard to read (or, at least, hard to skim):



Note: I find the above screenshot a bit easier to read than the actual page in my browser, dunno why.

---

> how about having the topics written after the link in parentheses to save on vertical space

Exactly what I did at first.  I think that's ok, but I think it looks a bit ugly when the lines wrap.  Also a bit ugly when a there's a a parenthesis at the end of the mention, e.g. "Simple Non-Interactive Coinjoin with Keys for Encryption Reused (SNICKER) ([Coinjoin](#))"

Additionally, I think always starting the list of topics on a new line makes the list easier to skim.



---

For completeness, here's a preview of the two ideas together which I think additionally makes it unclear where the title link ends and the topic links begin:

jnewbery:
Thanks for providing those alternatives @harding ! I think the version you chose looks good.

I think this is ready to merge now, and I don't want to bike-shed on minor design decisions, but a couple more points for your consideration:

- I _think_ the main goal for people visiting this page is to click through to the mention. If that's the case, then I do think it's strange that the link to those mentions is such a small area of the screen that's relatively difficult to click. What do you think about making the whole mention title a link but not coloring it blue?
- If getting to the mention is the main goal, then perhaps the topics could be in parentheses on a separate line, or in a smaller font.

harding:
> I think the main goal for people visiting this page is to click through to the mention.

In my mind, the primary goal is to help the reader discover which topics are most active.  E.g., Alice wants to know what's currently hot in Bitcoin development, so she loads this page, notices that schnorr/taproot are being mentioned a lot, and clicks through to those topics to learn more about them.

A secondary goal is just to provide easy-to-grasp evidence of Bitcoin's significant ongoing development is order to combat the often-repeated lie that Bitcoin development is stagnant.  This is inspired in part by my current belief that community-driven decentralized development can't really have universal roadmaps illustrating where we're going, but we can still illustrate progress by keeping track of where we've been and occasionally looking back and marveling at how much we accomplished through months and years of patient effort.

> the link to those mentions is such a small area of the screen that's relatively difficult to click.

I increased its font size to 125%, which was the largest I could go without seeming to affect the other aesthetics of the page (or without being more invasive with CSS).

> What do you think about making the whole mention title a link but not coloring it blue?

Slightly opposed.  If people don't know it's a link, they might click on it accidentally (e.g. when trying to select the text for copying or when clicking near one of the topic links).  That's not the end of the world, but it is unexpected.

> If getting to the mention is the main goal, then perhaps the topics could be in parentheses on a separate line, or in a smaller font.

I want people to go to either depending on what they want.  I don't really see either as more important or more useful than the other.  In other words, if I didn't find the all-blue design hard to read, I'd have gone with that.  This is the best compromise I've found so far.

jnewbery:
ok, looks good. Thanks @harding , and thanks for the patient explanation of your design choices :)

bitschmidty:
Just catching up on this PR, looks much better thanks, @harding

PR #292, opened 2019-12-10
Add Spanish translation of the Schnorr/Taproot blog post

jnewbery:
Also adds Spanish as a supported language on the site.

First commit makes minor changes to where the RSS link is placed on the publications/blog indexes, and only shows it for the english indexes.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -61,3 +61,4 @@ collections:
 
 languages:
   - ja
+  - es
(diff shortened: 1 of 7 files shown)


jnewbery:
> if you intend to have no English RSS link from the /en/newsletters page, as it is still present there but not on en/blog

I left it on /en/newsletters because it's always been there, but I agree that it's somewhat inconsistent to have it there but not on /en/blog. I'll add it to both places.

bitschmidty:
ACK

PR #304, opened 2019-12-19
Testing: upgrade HTML to identify mis-cased anchors

### LINKED ISSUE #238: Investigate why HTMLProofer didn't find broken internal link
In https://github.com/bitcoinops/bitcoinops.github.io/pull/167#discussion_r335079302, @bitschmidty found a broken internal link (anchor) on a PR that passed tests---but the tests are supposed to find broken links to anchors (and they have on many occasions).  My best initial guess is that this is because the brokeness is just a capitalization issue, but I need to investigate when I get a moment and see if I can make the test more strict.

harding:
Closes #238 

The version of HTML Proofer that we were using wasn't identifying situations where internal links were broken because of uppercase characters in all-lowercase links (or other variations of case errors).  This upgrades just our testing libraries (i.e. Jekyll is not upgraded) to get a new version of HTML Proofer that does verify correct casing.

You can generate the same dependency changes as in this PR by running the following command on master:

    bundle update -g testing

If you don't run the above command but you do have this PR in your history (e.g., it's been merged to master), you can update your dependencies by running:

    bundle install

You can verify that incorrect casing is detected by running `make` with the upgraded HTML Proofer version by with the first commit in this PR dropped (as it fixes a previously-undetected incorrectly cased link).


jnewbery:
ACK 837b731bd7501aa6cf14a1867d6a514c3513866e. Thanks Dave!

PR #311, opened 2020-01-01
News 36: typo fix: s/FindAndReplace/FindAndDelete/

harding:
This newsletter correctly calls the function `FindAndDelete()` in most places, but for some reason I used the wrong name in three other places.

Just to make sure I hadn't forgotten something I knew during the writing of the newsletter, I checked both the Bitcoin Core source code and bip-cleanup to confirm that neither mentions a `FindAndReplace()` function.


bitschmidty:
ACK

PR #319, opened 2020-01-14
news52: fix rpc setwalletflag

jonatack:
Git grepping returns no instance of "setwalletflags" in Bitcoin Core.

According to git blame, the `setwalletflag` RPC was added in https://github.com/bitcoin/bitcoin/commit/f904723e0d5883309cb0dd14b826bc45c5e776fb and named as such from the start.


jnewbery:
ACK

harding:
ACK.  Thanks!

PR #318, opened 2020-01-13
Topics: updates for December and four new topics

harding:
Updates and new topics, plus a couple minor QA changes.

PR #320, opened 2020-01-14
[Topics] Change date index layout, add stats, add internal page links

harding:
This PR makes three changes to the topic index pages

## Adds stats

Each page gets a single line subtitle with some basic statistics about that page, e.g.:

> - 44 topics (and 26 aliases in italics for topics with alternative names).
> - 212 indexed events in 18 months 
> - 17 categories for 44 unique topics, with many topics appearing in multiple categories.

## Adds internal page links

The alphabetical and category indexes get a condensed table of contents linking to the various subheads below, e.g.:

A table of contents for the date index is currently commented out as it links to December 2019, which won't be added until #318 is merged (and it won't really make sense until the January 2020 topics update is merged, so I'll uncomment it in that PR).

## New layout for the date index

We just changed the layout on this page last month in #283, but looking at the huge number of entries added in #318 made me unsatisfied with it.  This new change groups entries by month (as before), then by topic---with the topics mentioned the most that month being listed first:

A topic with two or more mentions in a month gets sub-bullet points; a topic with only one mention that month gets only one bullet point, which I think is a good balance of whitespace.

I'm not entirely satisfied with this layout either, but it's the best I've been able to think of so far.  Additionally, I note that it has to loop through every topic mention for every month.  Right now, it checks 36 months and 212 events for about 0.01 seconds per iteration (~0.36 seconds total).  If we add 200 events a year for 10 years, that means it'll take ~15 seconds to compile by 2030---assuming neither our software nor our computers speeds up in that time.  That stinks, but there are ways to optimize it (write a plugin in a real programing language) or we can use the old Bitcoin.org [hack](https://github.com/bitcoin-dot-org/bitcoin.org/blob/master/docs/setting-up-your-environment.md#fast-partial-previews-or-builds) of just disabling some things during the previews where we aren't working on them.

---

I can split this PR into several separate PRs if anyone would prefer that.

### DIFF
--- a/en/topic-categories.md
+++ b/en/topic-categories.md
@@ -5,10 +5,6 @@ layout: page
 ---
 {% include linkers/topic-pages.md %}
 
-<div>{% comment %}<!-- enclosing in a div forces this to be interpreted
-as HTML rather than Markdown so indentation over 4 characters doesn't
-produce code blocks -->{% endcomment %}
-
 {% capture raw_categories %}
 {%- for topic in site.topics -%}
   {%- if topic.categories == empty -%}
@@ -20,6 +16,17 @@ produce code blocks -->{% endcomment %}
 {%- endfor -%}
 {% endcapture %}
 {% assign categories = raw_categories | split: "|" | sort_natural | uniq %}
+
+<div class="center" markdown="1">
+{{ categories | size }} categories for {{site.topics | size}} unique
+topics, with many topics appearing in multiple categories.
+
+{% for category in categories %} [{{category | replace: ' ', '&nbsp;'}}](#{{category | slugify}})&nbsp;{% unless forloop.last %}\|{% endunless %}{% endfor %}
+</div>
+
+<div>{% comment %}<!-- enclosing in a div forces this to be interpreted
+as HTML rather than Markdown so indentation over 4 characters doesn't
+produce code blocks -->{% endcomment %}
 
 {% for category in categories %}
   <h3 id="{{category | slugify}}">{{category}}</h3>
--- a/en/topic-dates.md
+++ b/en/topic-dates.md
@@ -4,10 +4,6 @@ permalink: /en/topic-dates/
 layout: page
 ---
 {% include linkers/topic-pages.md %}
-<div>{% comment %}<!-- enclosing in a div forces this to be interpreted
-as HTML rather than Markdown so indentation over 4 characters doesn't
-produce code blocks -->{% endcomment %}
-
 {% capture raw_mentions %}
 {%- for topic in site.topics -%}
   {%- for mention in topic.optech_mentions -%}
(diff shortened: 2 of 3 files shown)


jnewbery:
Looks good. Thanks Dave! Some of comments:

- in topic-dates, can we sort the topic names alphabetically within each month?
- in topic-dates, I prefer having sub-bullets for each topic, and not adding a line break between each topic:
image
```
@@ -56,16 +56,12 @@ before the main content -->
         {%- assign topic_mentions = topic_data[1] | split: "ENDMENTION" -%}
         {%- assign topic_size = topic_mentions | size -%}
 
-        {%- if topic_size > 1 -%}
           <li>{{topic_title}}
             <ul>
             {%- for topic_mention in topic_mentions -%}
               <li>{{topic_mention}}</li>
             {%- endfor -%}
-            </ul><br></li>
-        {% else %}
-          <li>{{-topic_title-}} — {{topic_mentions[0]}}</li>
-        {%- endif -%}
+            </ul></li>
```
- in topic-categories, how about the following to ensure that the category isn't line-wrapped:
```
-{% for category in categories %} [{{category}}](#{{category | slugify}})&nbsp;{% unless forloop.last %}\|{% endunless %}{% endfor %}
+{% for category in categories %} [{{category | replace: ' ', '&nbsp;'}}](#{{category | slugify}})&nbsp;{% unless forloop.last %}\|{% endunless %}{% endfor %}
```
compare before:
image

after:
image

harding:
@jnewbery 

> in topic-dates, can we sort the topic names alphabetically within each month?

On this PR, we sort by topic popularity (within that month), which has a side-effect of putting topics of equal popularity in reverse alphabetical order.  E.g.,


```text
- A
    - foo
    - foo
    - foo
    
- Z 
    - foo
    - foo
    
- Y
    - foo
    - foo
    
- B: foo
```

The advantage of the popularity sort order is that it's the fastest to skim if you're looking for what was popular in a particular month (if instead you're looking for a particular topic, you're better off using one of the other indexes).  I think putting things in pure alphabetical order mix popular topics with unpopular topics and so make skimming harder.

That said, I can make equal-popularity items sort alphabetically.

What do you think?

> [other proposed changes]

Good ideas; I'll give them a try.

harding:
> in topic-dates, I prefer having sub-bullets for each topic, and not adding a line break between each topic:

You really think that's easier to read?  I find it harder.  E.g., without that change:



With that change (plus some bolding, which I think helps here):



I feel like eliminating the extra whitespace makes the list too much of a wall of text.

harding:
Forced pushed two changes:

- For topics with equal popularity within a month, topics are sorted alphabetically (instead of reverse alphabetically)
- Non-breaking space on category table of contents links

I'm happy to continue discussing the other proposed changes.

jnewbery:
> For topics with equal popularity within a month, topics are sorted alphabetically (instead of reverse alphabetically)

Sounds good. Sorting by popularity makes sense, and sorting alphabetically as a tie-break also makes sense to me.

> Non-breaking space on category table of contents links

Thanks!

> You really think that's easier to read? I find it harder. E.g., without that change:

I prefer it with the change, where I know that anything with level-1 indentation is the topic and anything with level-2 indentation is the link (with the current layout, the link can line-wrap into level-1 indentation).

I concede that this is just personal preference though. We can either poll other optechers for their preference or merge whichever you prefer. Entirely up to you.

harding:
> We can either poll other optechers for their preference or merge whichever you prefer.

For the record, out of four optechers who gave their opinion, everyone but me preferred the high-consistency version.

harding:
Updated to use the high-consistency version.

jnewbery:
Looks great. Thanks Dave!

PR #321, opened 2020-01-15
Eliminate duplicate anchors from all pages

harding:
In https://github.com/bitcoinops/bitcoinops.github.io/pull/318#pullrequestreview-343402308 , @bitschmidty noted that several links were linking to the table of contents rather than the actual content.  This occurred because we had multiple elements set to the same anchor, e.g the table of contents anchor was `<a id="miniscript">` and the actual section about miniscript was `<div id="miniscript">`.

In this PR, all duplicate instances of duplicate anchors are removed by renaming all but one of the elements.  Additionally, a test is added to the Makefile to prevent the problem from recurring.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,15 @@ test-after-build:
 	! find _site/ -name '*.html' | xargs grep ']\[' | grep -v skip-test | grep .
 	! find _site/ -name '*.html' | xargs grep '\[^' | grep .
 
+	## Check for duplicate anchors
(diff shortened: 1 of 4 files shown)


jnewbery:
ACK 29060d64c05856584710f0a5f7f20a4cefaa3acf.

Confirmed that the new test fails on master.

bitschmidty:
post utACK

PR #322, opened 2020-01-16
Tools: add crude calculator for transaction size

harding:
This adds a crude Javascript-based transaction size calculator to `/en/tools/calc-size/`.  No links are added to the tool from other parts of the site and I'm not sure we should add any unless the tool is improved.  

However, I've found it good enough for my own use and I think it might help y'all when you're burdened with verifying my transaction size math in newsletter and workbook content.  Future improvements can let it deal with a wider range of multisig variations, differing types of input and output scripts, non-keypath P2TR, uncompressed pubkeys, low-R signatures, LN onchain scripts, and other common transaction types.

That said, if you don't think this is a good fit for the Optech website, let me know and I can easily host it somewhere else.

### DIFF
--- a/en/tools/calc-size.md
+++ b/en/tools/calc-size.md
@@ -1,0 +1,368 @@
+---
+title: "Transaction size calculator"
+layout: page
+breadcrumbs: false
+
+size:
+  ## Overhead
+  nversion: 4
+  nin: 1
+  nout: 1
+  marker_and_flag: 0.5
+  nelements: 0.25
+  nlocktime: 4
+
+  ## Input
+  outpoint: 36
+  scriptsig_length_small: 1
+  scriptsig_length_big: 3
+  p2pkh_ss: 107   # 1 + 72 + 1 + 33
+  p2sh23_ss: 254  # 1 + 1 + 72 + 1 + 72 + 1 + 1 + (1 + 1 + 33 + 1 + 33 + 1 + 33 + 1 + 1)
+  nsequence: 4
+  p2wpkh_witness: 26.75  # (73 + 34)/4
+  p2wsh23_witness: 63.25  # (1 + 73 + 73 + 106)/4
+  p2tr_witness: 16.25  # 65/4
+
+  ## Output
+  nvalue: 8
+  scriptpubkey_length: 1
+  p2pkh_spk: 25  # 1 + 1 + 1 + 20 + 1 + 1
+  p2wpkh_spk: 22 # 1 + 1 + 20
+  p2sh23_spk: 23 # 1 + 1 + 20 + 1
+  p2wsh23_spk: 34 # 1 + 1 + 32
+  p2tr_spk: 34 # 1 + 1 + 32
+
+  ## Common elements
+  ecdsa_pubkey: 33
+  ecdsa_signature: 72
+  schnorr_public_key: 32
+  schnorr_signature: 64
+  public_key_hash: 20
+  script_hash_p2sh: 20
+  script_hash_p2wsh: 32
+---
+<form action="" id="vbytescalc" onsubmit="return false;">
+   <fieldset>
+
+    Transaction type:
+    <select id="type" onchange="updateTotal()">
+    <option value="p2pkh">P2PKH</option>
+    <option value="p2wpkh">P2WPKH</option>
+    <option value="p2sh2_3">P2SH 2-of-3 multisig</option>
+    <option value="p2wsh2_3">P2WSH 2-of-3 multisig</option>
+    <option value="p2tr">P2TR (taproot)</option>
+   </select>
+
+  <br>
+  Number of inputs: <input type="number" id="inputs" min="1" value="1" onchange="updateTotal()"/><br>
+  Number of outputs: <input type="number" id="outputs" min="1" value="2" onchange="updateTotal()"/>
+
+  <br><br><hr>
+  <div id="result"></div>
+  </fieldset>
+
+</form>
+
+## Legend: data field sizes
+
+All sizes in parenthesis in the *overhead*, *input*, and *output*
+sections are [vbytes][].  Sizes in the *common elements* section are bytes.
+
+### Overhead
+
+- **nVersion** ({{page.size.nversion}}) The
+  transaction version number
+
+- **Input count** ([compactSize][]) The number of inputs included in the
+  transaction.  {{page.size.nin}} byte for up to 252
+  inputs
+
+- **Output count** ([compactSize][]) The number of outputs included in
+  the transaction.  {{page.size.nout}} byte for up to
+  252 outputs
+
+- **nLockTime** ({{page.size.nlocktime}}) The earliest [epoch time][] or block
+  height when the transaction may be included in a block
+
+- Only in transactions spending one or more segwit UTXOs:
+
+    - **Segwit marker & segwit flag** ({{page.size.marker_and_flag}}) A
+      byte sequence used to clearly differentiate segwit transactions
+      from legacy transactions
+
+    - **Witness element count** ([compactSize][]/4) The number of witness
+      elements included in the transaction. {{page.size.nelements}}
+      vbytes for up to 252 elements
+
+### Input
+
+- **Outpoint** ({{page.size.outpoint}}) The txid and vout index number
+  of the output (UTXO) being spent
+
+- **scriptSig length** ([compactSize][]) The length of the scriptSig
+  field.  {{page.size.scriptsig_length_small}} vbyte for a scriptSig up to 252
+  vbytes.  Maximum of {{page.size.scriptsig_length_big}} vbytes for a maximum-length scriptSig (10,000 vbytes).
+
+- **scriptSig** (varies) The source of witness data for legacy UTXOs.
+  This data is used to prove that the transaction is authorized by
+  someone controlling the appropriate private keys.  For the templates
+  used by this calculator, the scriptSigs sizes are:
+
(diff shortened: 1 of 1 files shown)


jnewbery:
I think it's fine to add this. I also agree that we shouldn't link to it until it's a bit more polished.

One thing that I'd like to see added is documentation on how the values are calculated.

harding:
> One thing that I'd like to see added is documentation on how the values are calculated.

I don't understand this.  Is this more code documentation (I thought that was ok), more description about what fields are included in a transaction in the HTML output, or something else?

jnewbery:
> I don't understand this. Is this more code documentation (I thought that was ok), more description about what fields are included in a transaction in the HTML output, or something else?

More description in the html about how the values are calculated.

en/tools/calc-size.md
```diff
@@ -0,0 +1,184 @@
...
+    + 1 // number of outputs, TODO: increase for >252 outputs
+    + 4 // nLockTime
+    + witness_flag
+  )
+
+  size = overhead + input_size * inputs + output_size * outputs;
+  document.getElementById('result').innerHTML = (
+    "Total size: " + size + "<br>"
+    + "Overhead: " + overhead + "<br>"
+    + "Input size = " + inputs + " x " + input_size + ' = ' + inputs * input_size + "<br>"
```
bitschmidty: I think having the 'vbytes' unit after the sizes would help any ambiguity

en/tools/calc-size.md
```diff
@@ -0,0 +1,184 @@
...
+    + "Output size: " + outputs + " x " + output_size + " = " + outputs * output_size
+  );
+}
+
+// Calculate the result for the default form parameters
+calculateTotal();
```
bitschmidty: Id suggest a default input output count here (say 1 in 2 out?) to make initial page load more informative to a new visitor

en/tools/calc-size.md
```diff
@@ -0,0 +1,184 @@
...
+      + 1 + taproot_pubkey // includes push
+    ),
+    "segwit": true
+  };
+
+  // Get the form
+  var form = document.forms["vbytescalc"];
+
+  // The input/output type
+  // TODO: maybe allow the input type to be different from the output type
```
bitschmidty: Also could allow mixed types within inputs and outputs.

Scope creep!

en/tools/calc-size.md
```diff
@@ -0,0 +1,184 @@
+---
+title: "Transaction size calculator"
+layout: page
+---
+<form action="" id="vbytescalc" onsubmit="return false;">
+   <fieldset>
+
+    <select id="type" onchange="calculateTotal()">
```
bitschmidty: Perhaps "Transaction Type" label for this dropdown field

harding:
Forced push, hopefully addressing all comments (including providing more information about how the values are calculated).  Thanks for the reviews, @jnewbery and @bitschmidty!

en/tools/calc-size.md
bitschmidty: Looks good

en/tools/calc-size.md
bitschmidty: Perfecto

en/tools/calc-size.md
bitschmidty: Future idea, resolving.

en/tools/calc-size.md
bitschmidty: Resolving

en/tools/calc-size.md
```diff
@@ -0,0 +1,331 @@
...
+### Overhead
+
+- **nVersion** ({{page.size.nversion}}) The
+  transaction version number
+
+- **Input count** ([compactSize][]) The number of inputs included in the
+  transaction.  {{page.size.nin}} byte for up to 252
+  inputs
+
+- **Output count** ([compactSize][]) The number of inputs included in
```
bitschmidty: s/inputs/outputs

en/tools/calc-size.md
```diff
@@ -0,0 +1,331 @@
...
+      elements included in the transaction. {{page.size.nelements}}
+      vbytes for up to 252 elements
+
+### Input
+
+- **Outpoint** ({{page.size.outpoint}}) The txid and vout index number
+  of the output (UTXO) being spent
+
+- **scriptSig length** ([compactSize][]) The length of the scriptSig
+  field.  {{page.size.scriptsig_length}} vbytes for a scriptSig up to 252
```
bitschmidty: s/vbytes/vbyte (since its 1?)

en/tools/calc-size.md
```diff
@@ -0,0 +1,331 @@
...
+  vbytes.  Maximum of 3 vbytes for a maximum-length scriptSig (10,000 vbytes).
+
+- **scriptSig** (varies) The source of witness data for legacy UTXOs.
+  This data is used to prove that the transaction is authorized by
+  someone controlling the appropriate private keys  For the templates
```
bitschmidty: s/private keys For/private keys. For

en/tools/calc-size.md
```diff
@@ -0,0 +1,331 @@
...
+
+    - **P2SH 2-of-3** ({{page.size.p2sh23_ss}}) `OP_0 OP_PUSH72 <ecdsa_signature> OP_PUSH72 <ecdsa_signature> OP_PUSHDATA1 105 <OP_2 OP_PUSH33 <pubkey> OP_PUSH33 <pubkey> OP_PUSH33 <pubkey> OP_3 OP_CHECKMULTISIG>`
+
+
+- **nSequence** ({{page.size.nsequence}}) The sequence number for the
+  input.  Used by [BIP68][] and [BIP125][], with other values having no
+  defined meaning
+
+- **Witness data** (varies) The source of witness data for in segwit
+  transactions.  This data is used to proves that the transaction is
```
bitschmidty: s/to proves/to prove

en/tools/calc-size.md
```diff
@@ -0,0 +1,331 @@
...
+    <option value="p2pkh">P2PKH</option>
+    <option value="p2wpkh">P2WPKH</option>
+    <option value="p2sh2_3">P2SH 2-of-3 multisig</option>
+    <option value="p2wsh2_3">P2WSH 2-of-3 multisig</option>
+    <option value="p2tr">P2TR (taproot)</option>
+   </select>
+
+  <br>
+  Number of inputs: <input type="number" id="inputs" value="1" onchange="calculateTotal()"/><br>
+  Number of outputs: <input type="number" id="outputs" value="2" onchange="calculateTotal()"/>
```
0xB10C: Both HTML form inputs could do with a `min=0` attribute. 

image

en/tools/calc-size.md
```diff
@@ -0,0 +1,331 @@
...
+    "segwit": true
+  };
+
+  // Get the form
+  var form = document.forms["vbytescalc"];
+
+  // The input/output type
+  // TODO: maybe allow the input type to be different from the output
+  //       type, or multiple inputs of different types (or multiple
+  //       outputs of different types)
```
0xB10C: That would be cool! I might do a follow up PR that adds this.

harding: That'd be great!  Also having someone work on this who actually knows JS would probably be very beneficial.  :-)

harding:
Force pushed some edits:

- Hopefully resolved @bitschmidty's editing feedback
- added `min="1"` to input and output count form fields (@0xB10C suggested `min=0`, but since consensus rules require all transactions contain at least one input and one output, I decided to use a minimum of `1`)
- Separated the size calculation logic function from the HTML presentation function
- Added some crude semi-automated testing to verify the form is generating accurate results.  I only tested the default 1-in, 2-out cases; linking each of them (except P2WSH) to an actual transaction on the block chain.  This revealed two minor errors, which I corrected:
    - P2SH 2-of-3 was off by 2 vbytes because its scriptSig field size uses a 3-byte compactSize uint instead of the 1-byte uint I expected
    - P2WSH 2-of-3 was off by 0.25 vbytes because I forgot to include the size of the witnessScript field in the calculation (which would've been obvious if I read my own documentation on the page).  Additionally, using a quick scan of blocks 600,000 to 601,000, I was unable to find a P2WSH 2-of-3 transaction with 1-in, 2-outs that used 72 byte signatures for both of its sigs, so I don't have a txid reference for that.  However, I'm sure it's correct based on a mixed 71/72 byte signature transaction I checked.
- I also disabled breadcrumbs for this page because it wants to link to an `/en/tools/` page that doesn't exist.  Obviously we can add that page later if we want to publicize this

harding:
> P2SH 2-of-3 was off by 2 vbytes because its scriptSig field size uses a 3-byte compactSize uint instead of the 1-byte uint I expected

Oh, perhaps worth a dorky note, I think that if at least one signature for P2SH 2-of-3 uses low-R (and all pubkeys are compressed), then this can be a 1-byte compactSize uint.  Obviously, anyone who cares about space savings should be using P2WSH instead, but it might be worth a footnote in the scaling workbook or somewhere else that using low-R sigs can save you up to 4 vbytes here instead of just the expected 2 vbytes.  🤓

en/tools/calc-size.md
bitschmidty: resolved

bitschmidty:
tACK 6d0712c57afdbc68cf209dc22c790be8f626d709 (Chrome, Safari, Firefox on macOS)

harding:
Pushed a single line edit to make the P2WSH 2-of-3 witness example more accurate and less confusing.

jnewbery:
Looks great. Thanks Dave!

PR #331, opened 2020-01-29
Layout: add breadcrumbs to pages >1 level deep

harding:
- [x] Need translation of `home` (as in homepage) and `newsletters` to Japanese or Spanish to demonstrate how to use the string translation feature

This PR adds breadcrumbs to pages that are more than one level removed from the home page, e.g. "/ home / newsletters /" to a newsletter entry:

Features:
- Can be disabled by putting: `breadcrumbs: false` in the page metadata (YAML header)
- Breadcrumbs can be automatically derived from the page URL, e.g. `/en/topics/anchor-outputs/` automatically gets the breadcrumb "/ home / topics /"
    - We can also tweak the breadcrumb if desired, e.g. a newsletter URL is `/en/newsletters/2020/01/22/` but we don't currently have year or month index pages, so instead of creating "/ home / newsletters / 2020 / 01 /", we simply tweak the breadcrumb to create "/ home / newsletters /".  At some later point (maybe issue 100?), I think we probably want to split up our index by year, at which point we can extended the breadcrumb to "/ home / newsletters / 2020 /"
- The text for the breadcrumbs can be translated; that way we can continue using URLs that differ only by their `/<lang>/` part but readers of the localized content will get localized paths.  See the Japanese newsletter pages for an example (initially using "FIXME")

I think these breadcrumbs benefit the site now, but I developed them as part of porting the Scaling Workbook content to the site, where I think they're especially beneficial at helping to make a book's worth of content more navigable.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -49,6 +49,7 @@ defaults:
       path: ""  ## all pages
     values:
       image: https://bitcoinops.org/img/logos/optech-notext.png
+      breadcrumbs: true
   - scope:
       path: "_topics"
     values:
--- a/_data/localization/es.yml
+++ b/_data/localization/es.yml
@@ -1,0 +1,2 @@
+home: página principal
+newsletters: boletines
--- a/_data/localization/ja.yml
+++ b/_data/localization/ja.yml
@@ -1,2 +1,0 @@
-home: FIXME
-newsletters: FIXME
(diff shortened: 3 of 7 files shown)


jnewbery:
Great stuff, Dave. Thanks! I've added a Spanish localization file.

ACK e79c86cdf67df5237e57d4d60ba65e17f8bab19b

PR #332, opened 2020-01-29
added bitbank

haegwan:
Noticed the member companies are listed by alphabetical order. Please take a look and let me know if the change doesn't make sense.


bitschmidty:
@haegwan Thanks for opening this!

If you run this command locally you can preview the site:

`make preview` 

If you run this command locally, additional checks will be done. 

`make production`

In this case the image file was not optimized so that is what you had a failed build. I optimized the image file using:

`optipng -o7 img/members/bitbank.png `

Ive pushed that commit to your branch and will squash and merge now

PR #339, opened 2020-02-10
Topics: Covenants: update CTV link to point to BIP119

### LINKED ISSUE #338: Covenants topic has broken link
clicking on "OP_CHECKTEMPLATEVERIFY proposal" leads to

https://github.com/JeremyRubin/bips/blob/ctv/bip-ctv.mediawiki => 404

harding:
Closes #338 

Broken link identified by @moneyball (thanks!)

Note: we have a couple other links to this same old proposed BIP in newsletter issues that aren't updated in this PR.  Based on the context of the newsletters, I think pointing to an actually assigned BIP would be confusing, so I think it's best to leave them broken.  (But feel free to disagree, or to open a separate issue to discuss what to do about hard-to-point-anywhere-else broken links.)

### DIFF
--- a/_topics/en/covenants.md
+++ b/_topics/en/covenants.md
@@ -26,8 +26,8 @@ primary_sources:
   - title: Enhancing Bitcoin transactions with covenants
     link: https://fc17.ifca.ai/bitcoin/papers/bitcoin17-final28.pdf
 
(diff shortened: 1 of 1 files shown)


jnewbery:
ACK. Thanks @harding 

I agree that we shouldn't retroactively change links to an earlier proposal to be links to the BIP

PR #342, opened 2020-02-11
Add BTSE as a member

bitschmidty:


jnewbery:
Rebase on master to fix the build error

PR #344, opened 2020-02-11
Newsletter-80:Translate into Japanese

Ichiro0219:


bitschmidty:
ACK

Thanks @Ichiro0219 !

PR #352, opened 2020-02-21
Config base url change to correct rss feed urls.

### LINKED ISSUE #350: Get RSS feed to validate

bitschmidty:
Closes #350 

Small change to prevent double slashes '//' in RSS feed related urls.


harding:
Tested ACK.  Thanks!

PR #359, opened 2020-03-01
Newsletters: add #87 (2020-03-04)

harding:
- [x] Highlights from the Bitcoin Core PR Review Club @jnewbery 
- [x] `[Bitcoin Core #17985][] net: Remove forcerelay of rejected txs` @moneyball 
- [x] `[Libsecp256k1 #710][] Eliminate harmless non-constant time operations on secret data` @harding

### DIFF
--- a/_posts/en/newsletters/2020-03-04-newsletter.md
+++ b/_posts/en/newsletters/2020-03-04-newsletter.md
@@ -1,0 +1,177 @@
+---
+title: 'Bitcoin Optech Newsletter #87'
+permalink: /en/newsletters/2020/03/04/
+name: 2020-03-04-newsletter
+slug: 2020-03-04-newsletter
+type: newsletter
+layout: newsletter
+lang: en
+---
+This week's newsletter describes a proposed update to BIP340 schnorr
+keys and signatures, seeks feedback on a proposal to improve startup
+feature negotiation between full nodes, examines a suggestion for a
+standardized way to prevent hardware wallets from using corrupt nonces
+to leak private keys, and links to an analysis of the properties necessary in a
+hash function for taproot to be secure.  Also included are our
+regular sections for release announcements and notable changes to
+popular Bitcoin infrastructure projects.
+
+## Action items
+
+*None this week.*
+
+## News
+
+- **Updates to BIP340 schnorr keys and signatures:** as previously
+  mentioned in two separate items in Newsletter #84 ([1][news83 safety], [2][news83 tiebreaker]),
+  several updates to [BIP340][] schnorr signatures have been proposed
+  (which also affect [BIP341][] taproot).  Pieter Wuille [suggests][wuille
+  update] changing which variant of a public key should be used with
+  [BIP340][], with the new selection being based on the evenness of the
+  key.  Also, the recommended procedure for generating a nonce has been
+  changed to include the public key in nonce generation, to include
+  independently-generated randomness when available, and to include a step in the nonce
+  generation algorithm that uses the randomness to obfuscate the private key
+  in order to protect the key against [differential power analysis][].
+
+    Because of the significance of some of these changes, the tag for
+    the tagged hashes in BIP340 is changed, ensuring that any code
+    written for the earlier drafts will generate signatures that fail
+    validation under the proposed revisions.  Wuille requests community
+    feedback on the changes.
+
+- **Improving feature negotiation between full nodes at startup:** Suhas
+  Daftuar [requested feedback][daftuar wtxid] on a proposal to insert a
+  message into the sequence used by nodes to open a connection with a
+  new peer.  The new message would make it easier for a node to
+  negotiate what features it wants to receive from its peer.  A
+  challenge here is that previous versions of Bitcoin Core would
+  terminate a new connection if certain messages didn't appear in a
+  particular order---and it's into this strict sequence that Daftuar
+  wants to insert a new message.  The proposal does increment the P2P
+  protocol version to provide backwards compatibility, but Daftuar is
+  seeking feedback from maintainers of full nodes about whether the
+  insertion of negotiation messages would cause any problems.  If you're
+  aware of any problems, please reply to the thread.
+
+- **Proposal to standardize an exfiltration resistant nonce protocol:** Stepan
+  Snigirev [started][snigirev nonce] a discussion on the Bitcoin-Dev
+  mailing list about standardizing a protocol that prevents hardware
+  wallets from using biased nonces to leak
+  users' private keys.  One [previously proposed][sign to contract]
+  mechanism for defending against this attack would be to use the *sign
+  to contract protocol* to verify that the signature commits to
+  randomness selected by the hardware wallet's host computer or mobile
+  device.  Developers of libsecp256k1 have already been working on an
+  API to both [enable generic sign to contract][secp s2c] and build an
+  [exfiltration resistant nonce capability][secp nonce] on top of it.  Snigirev's
+  email describes the currently preferred protocol and how it can be
+  extended to multiple computers and Partially Signed Bitcoin
+  Transactions ([PSBTs][topic psbt]).
+
+- **Taproot in the generic group model:** Lloyd Fournier published his
+  [poster][fournier poster] for the Financial Cryptography conference
+  two weeks ago describing what properties must be present in the hash
+  function used with taproot in order for taproot to be secure.  This
+  extends a previous [proof][poelstra proof] by Andrew Poelstra which
+  made the broader assumption that the hash function acted as a [random
+  oracle][].  Those evaluating the cryptographic security of taproot are
+  encouraged to review Poelstra's proof and Fournier's poster.
+
+## Releases and release candidates
+
+*New releases and release candidates for popular Bitcoin infrastructure.
+Please consider upgrading to new releases or helping to test release
+candidates.*
+
+- [LND 0.9.1][] is a new minor version release that doesn't contain any
+  new features but which fixes several bugs, including a bug that can
+  result in "erroneous force closes between nodes".
+
+- [Bitcoin Core 0.19.1][] (release candidate)
+
+## Notable code and documentation changes
+
+*Notable changes this week in [Bitcoin Core][bitcoin core repo],
+[C-Lightning][c-lightning repo], [Eclair][eclair repo], [LND][lnd repo],
+[libsecp256k1][libsecp256k1 repo], [Bitcoin Improvement Proposals
+(BIPs)][bips repo], and [Lightning BOLTs][bolts repo].*
+
+- [Bitcoin Core #17985][] net: Remove forcerelay of rejected txs FIXME:moneyball
+
+- [Bitcoin Core #17264][] changes the default in RPCs that create or
+  process Partially Signed Bitcoin Transactions ([PSBTs][topic psbt])
+  to include any known [BIP32][] HD wallet derivation path.  This allows
+  any other wallets that later process the PSBT to use that information
+  to select the appropriate signing key or verify that a change output
+  pays the correct address.  Those wanting to keep their derivation
+  paths private may disable sharing using the `bip32_derivs`
+  parameter.
+
+- [C-Lightning #3490][] adds a `getsharedsecret` RPC that will derive a
+  shared secret from the combination of the local node's private key and
+  a user-specified public key.  This shared secret is derived the same
+  way other shared secrets are derived in the LN protocol (SHA256 digest of the
+  [ECDH][] result), which may not be the same way other programs derive
+  shared secrets using elliptic curve cryptography (e.g. for [ECIES][]).
+  This RPC can be useful for plugins that want to encrypt their
+  communication with other LN nodes.
+
+- [Eclair #1307][] updates the packaging scripts used by Eclair,
+  producing a zip file with everything needed to use the software.  This
+  new method allows Eclair GUI to be built deterministically in addition
+  to the core library (which we reported in [Newsletter #83][news83
+  eclair determ]).  [Reproducible builds][topic reproducible builds] help users ensure that the
+  software they run is based on the source code that's been publicly
+  reviewed.
+
+- [Libsecp256k1 #710][] makes minor changes to the library to help with
+  testing.  In one of the changes, a [function][secp256k1_ecdh] that
+  could previously return a range of values (contrary to its documented
+  behavior) will now only return either `0` or `1`.  At least one other
+  library [used][rust-secp256k1 ecdh ret] the old behavior and a concern
+  was [mentioned][ruffing concern] in the #secp256k1 IRC chatroom that
+  other programs or libraries might also be using the old unadvised
+  behavior.  If any of your programs use the `secp256k1_ecdh` function,
+  consider reviewing the discussion on [this PR][710 comment] and in the
+  related [rust-secp256k1 issue][rust-secp256k1 #196].
+
+- [BIPs #886][] updates [BIP340][] schnorr signatures with two
+  recommendations: (1) to include entropy in the nonce calculation
+  whenever a random number generator is available, and (2) to verify any
+  signatures produced by the software are valid before distributing them
+  to outside programs, at least whenever the extra computation and delay
+  isn't burdensome.  These two steps help ensure an attacker can't
+  recover a user's private key by obtaining invalid signatures with
+  reused nonces; see [Newsletter #83][news83 safety] for attack details.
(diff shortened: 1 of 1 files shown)


dongcarl:
@harding What about: 

#710 allows us to get a clean valgrind memtest report in #708 by eliminating the remaining 
(harmless) non-constant time operations on secret data. Notably, this PR introduces a change in behaviour that tightens up secp256k1_ecdh to better match the public documentation. Previously, secp256k1_ecdh could return a value that wasn't 0 or 1 if the hashfp being passed in did so, this is now impossible.

We could link to the documentation here: https://github.com/bitcoin-core/secp256k1/blob/96d8ccbd16090551aa003bfa4acd108b0496cb89/include/secp256k1_ecdh.h#L36-L48

harding:
@dongcarl 

> #710 allows us to get a clean valgrind memtest report in an upcoming PR #708 by eliminating the remaining (harmless) non-constant time operations on secret data. Notably, this PR introduces a change in behaviour that tightens up secp256k1_ecdh to better match the public documentation. Previously, secp256k1_ecdh could return a value that wasn't 0 or 1 if the hashfp being passed in did so, this is now impossible.

Think about this description from the perspective of a reader whose only technical experience is reading *(Mastering|Groking|Inventing) Bitcoin*.  I think they'd have questions like:

- What is valgrind?
- What's `secp256k1 #708` do for us?
- What's the normal benefit of constant time operations and why doesn't that benefit apply here?
- Why is behaviour spelled with a `u`?  :-)
- What is `secp256k1_ecdh` (a few items earlier in this newsletter is a separate link to the Wikipedia article on ECDH, so it is possible to go light on a description of what ECDH does).
- What is `hashfp` (even I'm not sure.  Hash floating point?  Hash fixed point?  Hash function pointer?)
- What other return values were possible before, and why were they possible?

Related to that last question is the special reason I made sure to put this otherwise seemingly minor change on the list for this week: Tim Ruffing's concerns about this API change being a breaking change for rust-bitcoin and it possibly breaking other libraries that called this function.

Note that the description doesn't have to answer all of those questions.  What it should do is communicate what's changing, what the benefit of the change is, and what the costs of the change are (if any).

Let me know if you're busy and you want me to take a crack at it.

jnewbery:
I've pushed a commit that adds the PR review club summary. I think it's ok, but I'm not thrilled with it, and I'm not sure how much value it adds to the newsletter. Happy to hear any suggestions for making it more compelling, or to spike it entirely if we don't think it's going to be that useful.

jnewbery:
oops, I broke the netlify build with a link failure. Pushed a fix.

bitschmidty:
Very informative thanks @jnewbery and @harding Will review once the remaining PRs are added.

harding:
@jnewbery 

> I've pushed a commit that adds the PR review club summary. I think it's ok, but I'm not thrilled with it, and I'm not sure how much value it adds to the newsletter.

Sadly, I have to agree that this doesn't seem to add value equal to its length.  I edited one of the points to see if I could work some magic, but I wasn't really able to improve upon it.  Some questions / comments:

1. Do you think the problem is that I asked you to keep the summaries short?  If they were longer, do you think they would be more informative?

2. What about making the summaries shorter?  (Or maybe even dropping the summaries altogether and just publish a list of links.)  The shorter the text, the less need there is for it to provide compelling information.

3. Do you think a completely different format would be better?  Reading through the notes and log for one of the meetings, we could do a longer summary of a meeting that went into more detail about each PR and highlighted its most notable questions and answers.  E.g., we could pick the best club meeting of the month and do it for that.

Alternatively to making any significant changes, I'm ok with publishing this as-is and waiting to see if there's any reaction to it.

_posts/en/newsletters/2020-03-04-newsletter.md
```diff
@@ -0,0 +1,226 @@
...
+experienced Bitcoin Core contributor provides background information about a
+selected PR and then leads a discussion on IRC.
+
+The Review Club is an excellent way to learn about the Bitcoin protocol, the
+Bitcoin Core reference implementation, and the process of making changes to
+Bitcoin. Notes, questions and meeting logs are posted on the website for those
+who are unable to attend in real time, and as a permanent resource for those
+wishing to learn about the Bitcoin Core development process.
+
+In this section, we summarise some recent Review Club meetings.
```
jonatack: I'm sympathetic to the British "ise" but find it a bit strange to mix "ize" and "ise" in the same newsletter without mentioning different authors for each part, otherwise people may think one of the two styles is a typo :)

_posts/en/newsletters/2020-03-04-newsletter.md
```diff
@@ -0,0 +1,226 @@
...
+
+- **[Use wtxid for transaction relay][review club 18044]:**
+  This PR ([#18044][Bitcoin Core #18044]) changes the way in which Bitcoin Core
+  announces transactions that it learns about to its peers. Instead of
+  announcing the transaction identifier or _txid_, Bitcoin Core would announce them
+  using the _wtxid_. The txid is a hash digest of the transaction which
```
jonatack: suggest: "using the witness transaction identifier or _wtxid_."

_posts/en/newsletters/2020-03-04-newsletter.md
```diff
@@ -0,0 +1,226 @@
...
+  excludes the witness, whereas the wtxid digest includes the witness.
+
+    Discussion in the meeting covered bandwidth usage, DoS vectors and
+    forwards/backwards compatibility for peer-to-peer changes.
```
jonatack: forward/backward

_posts/en/newsletters/2020-03-04-newsletter.md
```diff
@@ -0,0 +1,226 @@
...
+
+- **[O(1) OP_IF/NOTIF/ELSE/ENDIF script implementation / Abstract out script execution out of VerifyWitnessProgram()][review club 16902]:**
+  These PRs ([#16902][Bitcoin Core #16902] and [#18002][Bitcoin Core #18002]) are
+  both small refactors that have been pulled out of the [Schnorr/Taproot
+  implementation PR][Bitcoin Core #17977]. The author explains in PR 18002:
+  "[this] simplifies the changes for Taproot significantly [...] I may as well
+  give it some additional attention by PRing it independently." Both PRs make
+  changes to Bitcoin Core's script interpreter code -- a section of the codebase
+  which is very seldom changed by developers.
```
jonatack: "very" before "seldom" might be redundant

_posts/en/newsletters/2020-03-04-newsletter.md
```diff
@@ -0,0 +1,226 @@
...
+  any other wallets that later process the PSBT to use that information
+  to select the appropriate signing key or verify that a change output
+  pays the correct address.  Those wanting to keep their derivation
+  paths private may disable sharing using the `bip32_derivs`
+  parameter.
+
+- [C-Lightning #3490][] adds a `getsharedsecret` RPC that will derive a
+  shared secret from the combination of the local node's private key and
+  a user-specified public key.  This shared secret is derived the same
+  way other shared secrets are derived in LN (SHA256 digest of the
```
jonatack: s/LN/the LN/?

The style guide permits the acronym, but it may be more readable here to write "in the Lightning Network"

_posts/en/newsletters/2020-03-04-newsletter.md
```diff
@@ -0,0 +1,226 @@
...
+
+- **Updates to BIP340 schnorr keys and signatures:** as previously
+  mentioned in two separate items in Newsletter #84 ([1][news83 safety], [2][news83 tiebreaker]),
+  several updates to [BIP340][] schnorr signatures have been proposed
+  (which also affect [BIP341][] taproot).  Pieter Wuille [suggests][wuille
+  update] changing which variant of a public key should be used with
+  [BIP340][], with the new selection being based on the eveness of the
+  key.  Also, the recommended procedure for generating a nonce has been
+  changed to include the public key in nonce generation, to include
+  extra randomness when available, and to include a step in the nonce
```
jonatack: perhaps "in order to increase randomness" to not repeat "to include" in the same sentence

jonatack:
> I've pushed a commit that adds the PR review club summary. Happy to hear any suggestions for making it more compelling

On the individual meeting reports, perhaps they might be stuck in the middle ground of being neither super brief (one-line links to check out?), nor long enough to provide any interesting depth... the format might be a bit repetitive; too much formatting and not enough content.

jnewbery:
> Do you think the problem is that I asked you to keep the summaries short? If they were longer, do you think they would be more informative?

Certainly they'd be more informative if they were longer, but I think the critical metric we're looking at is value:length. If I make the summaries longer then they'd need to be even more interesting to justify taking up a huge part of the newsletter.

> What about making the summaries shorter? (Or maybe even dropping the summaries altogether and just publish a list of links.) The shorter the text, the less need there is for it to provide compelling information.

I don't think this will be very compelling. We'd be providing about as much value by saying "there's a meeting every week. They're here: https://bitcoincore.reviews/meetings/"

> Do you think a completely different format would be better? Reading through the notes and log for one of the meetings, we could do a longer summary of a meeting that went into more detail about each PR and highlighted its most notable questions and answers. E.g., we could pick the best club meeting of the month and do it for that.

I do think this would be better. Being more selective would naturally mean what we do choose would be more interesting and informative. With the short length requirement removed, I could provide a bit more context/interest.

> I'm ok with publishing this as-is and waiting to see if there's any reaction to it.

I'd prefer not to. If we agree that this isn't interesting enough we should just dump it rather than inflict it on the readers. It might also prevent people from reading through to the Notable Code Changes and Releases section because they get bored.

> they might be stuck in the middle ground of being neither super brief (one-line links to check out?), nor long enough to provide any interesting depth... the format might be a bit repetitive; too much formatting and not enough content.

Yes, I think this is exactly the problem.

I'm not sure if I'm going to have time to rework this to the new suggested format for this week's newsletter, but that's ok. We can potentially include it next week (and then the 2nd newsletter of the month going forward).

_posts/en/newsletters/2020-03-04-newsletter.md
```diff
@@ -0,0 +1,226 @@
...
+  challenge here is that previous versions of Bitcoin Core would
+  terminate a new connection if certain messages didn't appear in a
+  particular order---and it's into this strict sequence that Daftuar
+  wants to insert a new message.  The proposal does increment the P2P
+  protocol version to provide backwards compatibility, but Daftuar is
+  seeking feedback from maintainers of full nodes about whether the
+  insertion of negotiation messages would cause any problems.  If you're
+  aware of any problems, please reply to the thread.
+
+- **Proposal to standardize an anti covert nonce protocol:** Stepan
```
jnewbery: I don't really like the phrasing "anti covert nonce protocol", but I don't think I much prefer "anti-covert-nonce protocol", which is the first alternative I thought of. Is there a way we can rephrase this to remove the slightly ambiguous sequence of modifiers?

Same comment for "anti covert nonce capability" below.

I don't feel strongly about this. If you think "anti covert nonce <thing>" is the clearest way of describing this, then let's stick with it.

_posts/en/newsletters/2020-03-04-newsletter.md
```diff
@@ -0,0 +1,226 @@
...
+  [ECDH][] result), which may not be the same way other programs derive
+  shared secrets using elliptic curve cryptography (e.g. for [ECIES][]).
+  This RPC can be useful for plugins that want to encrypt their
+  communication with other LN nodes.
+
+- [Eclair #1307][] updates the packaging scripts used by Eclair,
+  producing a zip file with everything needed to use the software.  This
+  new method allows Eclair GUI to be built deterministically in addition
+  to the core library (which we reported in [Newsletter #83][news83
+  eclair determ]).  Reproducible builds help users ensure that the
```
jnewbery: Link to https://bitcoinops.org/en/topics/reproducible-builds/?

_posts/en/newsletters/2020-03-04-newsletter.md
jnewbery: Yes, the spelling should be consistent within a document, and Dave is the main author of this newsletter, so I should follow his preference.

_posts/en/newsletters/2020-03-04-newsletter.md
```diff
@@ -0,0 +1,226 @@
...
+  wants to insert a new message.  The proposal does increment the P2P
+  protocol version to provide backwards compatibility, but Daftuar is
+  seeking feedback from maintainers of full nodes about whether the
+  insertion of negotiation messages would cause any problems.  If you're
+  aware of any problems, please reply to the thread.
+
+- **Proposal to standardize an anti covert nonce protocol:** Stepan
+  Snigirev [started][snigirev nonce] a discussion on the Bitcoin-Dev
+  mailing list about standardizing a protocol that prevents hardware
+  wallets from generating a signature that uses a biased nonce to leak
```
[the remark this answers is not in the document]
harding: > I think you'd need a series of biased nonces to leak information

Definitely not my expertise, but I assume one biased nonce would be enough to recover the individual private key if you knew enough about the algorithm used to create the bias (e.g. what if the bias was always `nonce = 0`).  I think leaking the wallet seed, master private key, or the chain code corresponding to a private key would probably take at least two signatures/nonces (and more than that if you wanted to prevent third parties from detecting the exfiltration or using it to their own advantage before you).

_posts/en/newsletters/2020-03-04-newsletter.md
```diff
@@ -0,0 +1,226 @@
...
+  the user's private key.  One [previously proposed][sign to contract]
+  mechanism for defending against this attack would be to use the *sign
```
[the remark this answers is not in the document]
harding: I'm confused, the attack is described earlier: "generating a signature that uses a biased nonce to leak the user's private key".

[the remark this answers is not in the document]
jonatack: Right, it's a bit buried in that long sentence. My thinking was that splitting the sentence into two smaller ones (attack, then starting a solution discussion) might bring out the attack more explicitly.

harding:
Pushed three commits:

1. Edits for feedback to my parts (thanks @adamjonas, @jonatack, and @jnewbery!)
2. Addition of a description for the missing merged PR
3. Edits to @moneyball's merged PR (good description!  just tightening it up)

Still todo is either remove or update the PR Club section @jnewbery 

Thanks everyone!

_posts/en/newsletters/2020-03-04-newsletter.md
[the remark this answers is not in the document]
jnewbery: I'm also not an expect. I think the new wording ("using biased nonces") is safer though!

jnewbery:
I've removed my PR Review Club summaries. I'll aim to write up one or two in more detail for next week's newsletter.

ACK d4c1b24e68966f9bb7f054b92364a6ab6343589f. Really informative newsletter this week, Dave. Thanks as ever!

_posts/en/newsletters/2020-03-04-newsletter.md
```diff
@@ -0,0 +1,181 @@
...
+[secp s2c]: https://github.com/bitcoin-core/secp256k1/pull/589
+[secp nonce]: https://github.com/bitcoin-core/secp256k1/pull/590
+[ecdh]: https://en.wikipedia.org/wiki/Elliptic_curve_Diffie-Hellman
+[ecies]: https://en.wikipedia.org/wiki/Integrated_Encryption_Scheme
+[bip340 update]: #updates-to-bip340-schnorr-keys-and-signatures
+[news83 eclair determ]: /en/newsletters/2020/02/05/#eclair-1295
+[secp256k1_ecdh]: https://github.com/bitcoin-core/secp256k1/blob/96d8ccbd16090551aa003bfa4acd108b0496cb89/src/modules/ecdh/main_impl.h#L29-L69
+[rust-secp256k1 ecdh ret]: https://github.com/rust-bitcoin/rust-secp256k1/blob/master/src/ecdh.rs#L162
+[ruffing concern]: https://gist.githubusercontent.com/harding/603c2b18241bf61bb0bbe7a0383cf1c9/raw/20656e901472f217d1faa381ddda1d11214900da/foo.txt
+[710 comment]: https://github.com/bitcoin-core/secp256k1/pull/710/files#r370987476
```
bitschmidty: This link doesn’t appear to work anymore. Perhaps the comment was resolved.

harding: Pushed updated link that works for me.

PR #360, opened 2020-03-02
change: link to offical BIPs for BIP340-342

0xB10C:
fixes https://github.com/bitcoinops/bitcoinops.github.io/issues/347

I left the following untouched as the link to the _transaction digest_ section is [no longer up to date](https://github.com/sipa/bips/commit/c0d2f93f3c3c3dcee2318ea26f946ae17de37865#diff-506cf22853a59a8c57550c1f1d9cedfb). I'll let you to decide how to follow up with this one.
https://github.com/bitcoinops/bitcoinops.github.io/blob/f3f9b7a1bc833cba0f8d31a167d6a86c0abba9ff/_topics/en/op_codeseparator.md#L110-L111

### DIFF
--- a/_includes/references.md
+++ b/_includes/references.md
@@ -47,9 +47,9 @@
 [bip-anyprevout]: https://github.com/ajtowns/bips/blob/bip-anyprevout/bip-anyprevout.mediawiki
 [bip-cleanup]: https://github.com/TheBlueMatt/bips/blob/cleanup-softfork/bip-XXXX.mediawiki
 [bip-coshv]: https://github.com/JeremyRubin/bips/blob/op-checkoutputshashverify/bip-coshv.mediawiki
-[bip-schnorr]: https://github.com/sipa/bips/blob/bip-schnorr/bip-schnorr.mediawiki
-[bip-taproot]: https://github.com/sipa/bips/blob/bip-schnorr/bip-taproot.mediawiki
-[bip-tapscript]: https://github.com/sipa/bips/blob/bip-schnorr/bip-tapscript.mediawiki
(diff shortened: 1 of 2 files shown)


jnewbery:
I think netlify is a spurious failure. I don't think I have access to restart builds so I just force-pushed an empty update to kick off the build again.

> I left the following untouched as the link to the transaction digest section is no longer up to date. I'll let you to decide how to follow up with this one.

I think it should link to https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#common-signature-message, which describes the transaction digest.

There are still a bunch of links to `https://github.com/sipa/bips/blob/bip-schnorr` in old newsletters which are now dead, but I think that's the nature of writing a newsletter. Some of the links will die over time and we shouldn't go back and change them if that happens.

0xB10C:
> I think it should link to https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#common-signature-message, which describes the transaction digest.

Pushed a commit addressing that.

> There are still a bunch of links to https://github.com/sipa/bips/blob/bip-schnorr in old newsletters which are now dead, but I think that's the nature of writing a newsletter. Some of the links will die over time and we shouldn't go back and change them if that happens.

I as well think the links in the newsletters should not be changed afterwards. This would create a lot of maintenance work and, in this case, they still link to the correct repo which was relevant at that time. Somebody doing archaeology in maybe 10 years might be grateful to find the links unchanged.

harding:
Untested LGTM.  Thanks @0xB10C!

I agree with generally avoiding changing broken external links in old newsletters.

PR #369, opened 2020-03-11
Plugins: create clickable anchor links

harding:
On Twitter, @Christewart [suggested](https://twitter.com/Chris_Stewart_5/status/1237766220193181696) adding hover links to our content similar to those GitHub automatically generates for its readme-style documentation.  I think that's a good idea, but in addition to the subheads people expect to have anchors, we also automatically generate anchors for most of our bullet lists.  Thinking about how to add links for those without making things too unsightly, I had the idea to replace the bullet point (•) with a clickable link icon (🔗).  For example:


jnewbery:
Concept ACK. I do the right click->inspect element->copy id tag dance frequently, so even if it's only for selfish reasons, I like this.

I clicked around a bit and most of the pages look good. I checked https://deploy-preview-369--bitcoinops.netlify.com/en/newsletters/2019/12/28/#contents and some of those contents entries don't have a generated link. It also seems weird to have links generated for the table of contents.

bitschmidty:
Was just playing with an alternative where the icon appears at the end of the element when you hover the text. Hover helps with keeping the clutter down. But might not be as intuitive.

ezgif-5-0000dfa59cb9

Any feedback on that usability approach? Not great for mobile (maybe even better disabled for mobile)

harding:
@bitschmidty how does it link to the particular element?  Most things I've seen of that nature link to the element by its xpath, which works for perfectly static content but which will break in a way if anything on the page changes.

As you mention, I don't like hover things because they don't work on mobile (and, when possible, I prefer to keep the site looking basically the same on mobile as desktop, besides basic reflowing and resizing).

Finally, I'm guessing your solution may use JS.  As someone who browses as much as possible with NoScript, I prefer to minimize the use of JS for basic site features.

That all said, I really appreciate you experimenting with other approaches!  (FYI, although it suffers from the problems described above, a really nifty feature is linkable highlighting as implement on SNI; e.g.: https://satoshi.nakamotoinstitute.org/emails/cryptography/1/#selection-95.4-99.10 )

_plugins/auto-anchor.rb
```diff
@@ -21,7 +21,8 @@
         ## Remove double-quotes from titles before attempting to slugify
         title.gsub!('"', '')
         ## Use Liquid/Jekyll slugify filter to choose our id
-        id_prefix = '- {:#{{ "' + title + '" | slugify: "latin" }}} '
+        slug = '#{{ "' + title + '" | slugify: "latin" }}'
+        id_prefix = '- {:' + slug + ' .foo} <a href="' + slug +'" class="bar">🔗</a>'
```
jonatack: I'm not sure how Jekyll plugins work yet, but if this is dynamically run and performance matters, it should be faster to use string interpolation, e.g. `"#{title} #{slug}"` (which is the usual practice), rather than concatenation with the `+` operator.

harding: Jekyll is a static site render (although it has a live preview mode for development), so none of the code is run dynamically.  That said, I agree that this is getting hairy.  Using string concatenation made sense to me for the first version of this plugin when I just wanted to avoid having to escape quotes, but we keep doing more stuff, so now I agree it's time to bite the bullet and format it properly.

jonatack: @jnewbery pointed me to this in order to work on something similar for the review club.

Here it is, if helpful (don't hesitate to review/improve my regex): https://github.com/bitcoin-core-review-club/bitcoin-core-review-club.github.io/pull/147/files

harding:
This is now ready for full review.  I've addressed @jnewbery's comment about the year-in-review newsletter and @jonatack's code formatting issue.  I've also resolved the test failure---it looks Jekyll was silently not creating anchor links for cases where the entire anchor (id tag) was numbers (which is not allowed in HTML) or was empty (which often happens with the Japanese translations).  Since this PR links to those non-existent elements, that produced an broken link which our tests created.  I've fixed all of those and the tests should catch any similar problems in the future.  We already have documentation about adding text in HTML comments into to effect links in https://github.com/bitcoinops/bitcoinops.github.io/blob/master/CONTRIBUTING.md

_posts/ja/newsletters/2019-10-16-newsletter-ja.md
```diff
@@ -20,7 +20,7 @@ lang: ja
 - **Eclair 0.3.2へのアップグレード:** [Eclair's newest release][eclair 0.3.2]
  は、バックアップが改善され、ゴシップデータの同期がより帯域幅効率的になります（特にモバイルデバイス上のノードなどの非ルーティングノード用）。またその他の多くの新機能とバグ修正が含まれています。
 
-- **リリース候補（RC）のレビュー:** 2つの主要なBitcoinインフラストラクチャプロジェクトが、ソフトウェアの次のバージョンのリリース候補（RC）を発表しました。一般公開前に問題を発見して修正できるよう、開発者と経験豊富なユーザーはこれらのRCのテストを支援するようお願いします。
+- **<!--n1-->リリース候補（RC）のレビュー:** 2つの主要なBitcoinインフラストラクチャプロジェクトが、ソフトウェアの次のバージョンのリリース候補（RC）を発表しました。一般公開前に問題を発見して修正できるよう、開発者と経験豊富なユーザーはこれらのRCのテストを支援するようお願いします。
```
jnewbery, line 23: I'm not sure this is needed, since there are no other anchors on this page called with id `#rc`. The link currently works on the live site: https://bitcoinops.org/ja/newsletters/2019/10/16/#rc

How did you find where to add these comments?

_posts/ja/newsletters/2019-10-16-newsletter-ja.md
```diff
@@ -45,7 +45,7 @@ lang: ja
 
     Wuilleのメールと更新されたBIPは、これらの各決定の理論的根拠と、それらに関する以前の議論へのリンクを提供しています。
 
-- **デフォルトのLN feeの引き上げ提案:** Rusty Russellは、ノードが使用するデフォルトのチャネル内feeを、1,000ミリサトシ（msat）+100万分の1（ppm）から5,000 msat+500 ppmに増やすことを提案しました。彼は、現在のデフォルトではユーザーが料金を引き下げる余地があまりなく、現在高い料金を請求しているノードが主流である要因として多くのユーザーが現在料金に敏感ではないことを指摘しています。このメールには、さまざまな金額をUSD単位で10,000ドル/ BTCで送金するための新旧のコストを見積もるチャートが記載されています。
+- **<!--n2-->デフォルトのLN feeの引き上げ提案:** Rusty Russellは、ノードが使用するデフォルトのチャネル内feeを、1,000ミリサトシ（msat）+100万分の1（ppm）から5,000 msat+500 ppmに増やすことを提案しました。彼は、現在のデフォルトではユーザーが料金を引き下げる余地があまりなく、現在高い料金を請求しているノードが主流である要因として多くのユーザーが現在料金に敏感ではないことを指摘しています。このメールには、さまざまな金額をUSD単位で10,000ドル/ BTCで送金するための新旧のコストを見積もるチャートが記載されています。
```
jnewbery, line 48: This is also not needed since the existing anchor id `#ln-fee` works.

_posts/ja/newsletters/2019-10-16-newsletter-ja.md
```diff
@@ -59,7 +59,7 @@ lang: ja
 
     C-LightningのメンテナーであるZmnSCPxjとEclairのメンテナーであるPierre-Marie Padiouは、提案への支持を示しました。LNDのメンテナーであるOlaoluwa Osuntokunは、提案の背後にある要因にはいくつかの欠陥があると考え、代わりに「将来のルーティングノードオペレーターにベストプラクティスへの教育や分析ツール提供などを行い、市場参加者に任せて、安定した経済的合理的な料金にすること」を主張しました。この記事の執筆時点では、このトピックに関する議論は継続中です。
 
-- **会議概要：暗号経済システムサミット:**MITキャンパスで先週末に開催されたこの会議では、暗号通貨の有用性と安全性の確保に関するさまざまなトピックを取り上げました。  [DCIのYouTubeチャンネル][css vids]にてビデオは視聴可能となっており、いくつかの講演の[トランスクリプト][css ts]はBryan Bishopなどによって提供されました。 [講演の内容][css ts]のうちの次の3つのトピックは、このニュースレターの読者にとって特に技術的な関心があると思われます。
+- **<!--n3-->会議概要：暗号経済システムサミット:**MITキャンパスで先週末に開催されたこの会議では、暗号通貨の有用性と安全性の確保に関するさまざまなトピックを取り上げました。  [DCIのYouTubeチャンネル][css vids]にてビデオは視聴可能となっており、いくつかの講演の[トランスクリプト][css ts]はBryan Bishopなどによって提供されました。 [講演の内容][css ts]のうちの次の3つのトピックは、このニュースレターの読者にとって特に技術的な関心があると思われます。
```
jnewbery, line 62: This _is_ required. Should we try to make the anchor ids a bit more meaningful (eg change this to `#mit`) or does that not seem worthwhile?

_plugins/auto-anchor.rb
```diff
@@ -21,7 +21,8 @@
         ## Remove double-quotes from titles before attempting to slugify
         title.gsub!('"', '')
         ## Use Liquid/Jekyll slugify filter to choose our id
-        id_prefix = '- {:#{{ "' + title + '" | slugify: "latin" }}} '
+        slug = "\#{{ \"#{title}\" | slugify: 'latin' }}"
+        id_prefix = "- {:#{slug} .anchor-list} <a href=\"#{slug}\" class=\"anchor-list-link\">🔗</a>"
```
jnewbery, line 25: Now that I've looked at the preview for a bit of time, I don't really like covering the newsletters with the anchor glyph. I also don't think it's immediately intuitive that it's a link to that section of the document. For example, the links in the 'notable commits' section (eg https://deploy-preview-369--bitcoinops.netlify.com/en/newsletters/2019/10/16/#bitcoin-core-15437) could easily be interpreted as a link to the PR itself.

I'm leaning more towards just having these links as bullets, perhaps styled in blue as a subtle hint that they're clickable. I think that's much more pleasant to look at and doesn't display a minor feature in front of everyone the whole time:

```suggestion
        id_prefix = "- {:#{slug} .anchor-list} <a href=\"#{slug}\" class=\"anchor-list-link\">●</a>"
```

This is obviously just an aesthetic preference on my part. Let me know what you think.

_posts/ja/newsletters/2019-10-16-newsletter-ja.md
harding: > How did you find where to add these comments?

The failing HTMLProofer test identified the broken file but not the line number; I opened the files and guess where the links were broken.  Looks like I guessed wrong here.  I'll back out this particular change.

_posts/ja/newsletters/2019-10-16-newsletter-ja.md
harding: Oh, sure, meaningful is better than meaningless, but for a lot of these I'll need to cross-check the English (or use Google Translate) to determine what the context is---and that didn't seem worth it for anchors nobody previously noticed were missing.

_plugins/auto-anchor.rb
harding: I like that idea.  I'll give it a try and see how it looks.

harding:
Forced pushed changes: removed unnecessary changes identified by @jnewbery and changed the anchor link to a blue bullet per his suggestion.  I think that looks a lot nicer.  To help keep everything looking consistent, I changed all of our other bullets to also use the filled circle (disc) icon (we previously used hollow circles for sub items).

harding:
Forced pushed with no changes over @jnewbery's previous commit (I just merged that code into the relevant previous commits and added a Co-Authored-By tag).  I think that resolves all pending concerns.  (Thanks, John!)

jnewbery:
force-pushed some more very minor changes. Thanks for persevering with this, @harding !

PR #387, opened 2020-04-08
Add helper function for creating Review Club <details> lists

### LINKED ISSUE #385: Implement helper for lists of <details> elements
We use lists of `<details>` with nested `<summary>` tags in the monthly Bitcoin Core PR Review Club section.  These are not only kind of ugly but Kramdown also doesn't process any Markdown inside these HTML tags, so `code` text inside backtics, *italics*, or `[Markdown][]` links aren't processed unless every HTML tag contains `markdown="1"`.  Even if that invocation is added, our current version of Kramdown wants the `</details>` and `</span>` tags to appear on lines separate from the text and insists on wrapping the text in paragraph tags.

We should be able to resolve those issues using a helper.  I'd prefer something that just uses Liquid templating via an `{% include %}` directive, since (in my experience) that's the least likely to break because of a new Jekyll release.  If we can't create an easy-to-use and not-too-ugly helper, then creating a plugin would be better than continuing to create these fold-outs by hand.

harding:
Closes #385 

This provides an `{% include %}` template that can be used to create the `<details>` lists currently used for the monthly Bitcoin Core PR Review Club summaries, although it's also hopefully generic enough that it can be reused for any other `<details>` lists we create.  Here's a *before* screenshot of a manually created list from https://bitcoinops.org/en/newsletters/2020/03/11/#bitcoin-core-pr-review-club with one item expanded:

Here's the same list using this helper:

There's some extra whitespace in the second example; that wasn't deliberate but I liked it so I haven't removed it---but I can if anyone wants it gone.

---

If you build the website using `JEKYLL_ENV=email`, e.g. `JEKYLL_ENV=email make clean preview`, the list is formatted the same way @bitschmidty has used for the emailed newsletter.  E.g., here's a screenshot of Newsletter 88:

And here's the `env=email` version:

---

Feedback requested, particularly from @jnewbery, @jonatack, and @bitschmidty as those most likely to use this, about whether it looks practical.  If it seems good, the todos are:

- [x] Convert the remaining `<details>` lists in Newsletters 88 and 92 (both English and Japanese) to use this helper
- [x] Remove the old `review-club-questions` CSS class


assets/css/main.scss
```diff
@@ -265,3 +265,20 @@ div.localization
...
+  list-style-type: none;
+}
+
+.qa_details details {
+  border: thin solid silver;
+  margin: 2pt;
+  padding: 2pt;
+}
+
+.qa_details details summary {
```
bitschmidty, line 282: I think having a :hover style for the summary which changes the mouse to pointer would be a good usability addition.

harding:
@bitschmidty 

> Suggest mouse pointer hover style (noted inline comment)

Good idea!  Implemented.

> Doesn’t seem we have an anchoring capabilities for the qa blocks or individual questions/answers.

Even better idea!  Implemented.  Anchors are attached to the entire details box, using the question text to create the slug.  We don't have an easy glyph to attach the anchors like the bullet in #369, so I added a link icon to end of the answer field, which I think is ok---but I'm happy to try something else.  Unfortunately, there's no way to trigger the opening of the details box using pure CSS so anyone following a link to a particular question will still have to click on the box (or we'll have to use JS, yuck).

I left the email version free of anchors or links, since I wasn't sure whether that'd be useful.  I can easily add them---let me know.

> My flow would still be to copy/paste the newsletter content from the live production site, since all of the links are correct there. [...] Or is there a better approach for me to take here? (building prod locally w/ correct domain name, etc)

#389 should fix the domain issue.

---

If everything is looking basically ok, I'll convert the remaining lists to use this function.

_posts/en/newsletters/2020-04-08-newsletter.md
```diff
@@ -7,7 +7,7 @@ type: newsletter
 layout: newsletter
 lang: en
 ---
-This week's newsletter describes work on simplified ECDSA adaptor
+Foo This week's newsletter describes work on simplified ECDSA adaptor
```
jnewbery, line 10: Barrrr

harding: Whoops.  Fixed.  Thanks!

_includes/functions/details-list.md
```diff
@@ -0,0 +1,15 @@
+<div class="{{include.class | default: 'qa_details'}} {{jekyll.environment}}" markdown="1">
+{% for i in (0..9) %}
```
jonatack, line 2: 9 questions/answers seems a reasonable limit; should it be made explicit somewhere?

_posts/en/newsletters/2020-04-08-newsletter.md
```diff
@@ -79,25 +79,29 @@ to obtain the transaction as fast as possible from another peer.
...
+      relay of [BIP152][] compact blocks."
+
+  q3="When and why was `mapRelay` originally added?"
+  a3="`mapRelay` was present in the first version of Bitcoin. It ensures that if
+     a node announced a transaction, it can be downloaded even if it is confirmed in
+     a block between being announced and the peer requesting it."
+
+  q4="Describe one issue with removing `mapRelay`?"
+  a4="It could cause requested transactions in honest situations to more often be
+      `notfound` with delays of up to 2 minutes, hurting propagation."
```
jonatack, line 103: Markdown works here now? Nice!

harding:
@jonatack 

> would it be feasible for it to convert markdown code chars to html code tags?

I should've mentioned it in my OP, but this will indeed make that conversion, providing full markdown conversion in the context of the evaluated page (so, e.g., we can put reference-style link definitions at the bottom of the page like normal).  For the sake of a demo, I quickly converted the first section of your section from last week to the new format (changing the code tags to backtics) so you can see that it works.  See: https://deploy-preview-387--bitcoinops.netlify.com/en/newsletters/2020/04/08/#why-could-retrying-notfound-more-quickly-be-helpful

> 9 questions/answers seems a reasonable limit; should it be made explicit somewhere?

Good idea.  I'll see if I can get it to fail to build if anyone tries more than 9.

Thanks for the review!

_posts/en/newsletters/2020-04-08-newsletter.md
jonatack: Oh, `markdown="1"` seem to do it :+1:

jonatack: Oh right, `markdown="1"` seems to do it :+1:

assets/css/main.scss
bitschmidty: The hover style looks good!

jnewbery:
Concept ACK. I think this can be updated to use simplified styling for JEKYLL_ENV=email now that #389 is merged.

harding:
All feedback should be addressed (thanks!).

_includes/functions/details-list.md
```diff
@@ -0,0 +1,25 @@
...
+  {% capture q %}q{{i}}{% endcapture %}
+  {% capture a %}a{{i}}{% endcapture %}
+  {% if include.[q] == nil %}{% break %}{% endif %}
+  {% assign qslug = include.[q] | slugify: 'latin' %}
+  {% if jekyll.environment == "email" %}
+   - <i markdown="1">{{include.[q]}}</i><br>{{include.[a]}}
+  {% else %}
+   - <details id="{{qslug}}" markdown="1"><summary><span markdown="1">{{include.[q]}}</span></summary>
+     {{include.[a]}}&nbsp;<a href="#{{qslug}}">🔗</a>
```
jnewbery, line 21: I suggest we drop the link icon here:

- I think the most intuitive interpretation is that clicking on this will link to the discussion or source, not this section of the newsletter
- it's unlikely we (or anyone else) will need to link to these often. If we do, we can just inspect the html and grab the id from there.

harding:
Dropped link icon.  Also tested in an RSS feed reader (not great, but ok I think):

PR #389, opened 2020-04-10
Preview: allow setting base URI to https://bitcoinops.org

### LINKED ISSUE #393: Email Build: Strip anchors from bullets
In https://github.com/bitcoinops/bitcoinops.github.io/pull/369 we added hoverable, clickable anchor links plugin.

During email newsletter generation I currently remove these bullets to keep the email version clean. It would be nice to disable this plugin/feature when building for email (JEKYLL_ENV=email)

harding:
Closes #393 

Related to https://github.com/bitcoinops/bitcoinops.github.io/pull/387#pullrequestreview-390772699 and some internal discussion, I think this small change should make it possible to generate a local preview of the site content that's suitable for copy/pasting into MailChimp for sending the newsletter.

## How to use

1. Generate a preview normally, e.g. `make preview`
2. Browse to the page you will want to copy/paste
3. Kill the preview server (e.g. Ctrl-C)
4. Generate a fresh preview using the `JEKYLL_ENV=email` environment variable, e.g. `JEKYLL_ENV=email make clean preview` (note: you **need** clean for this)
5. Reload the page in your browser

All the relative links on the site should've been transformed in your browselr from, e.g., `http://localhost:4000/foo/bar` to `https://bitcoinops.org/foo/bar`.  (Note: in the source code, they'll still be relative links, so this only works for rich text copying.)

The reason the process above has you browse to the page you want before creating the `env=email` preview is that you can't browse from the main index to particular pages after you start the `email` preview (because all relative links now point away from your localhost to bitcoinops.org).

## Notes

Technically, this change is bad HTML.  This setting should be in the `<head>` section, but that's currently sourced from the Ruby gem package that provides our template and so can't be edited without copying the file into our repository.  Since this PR is just a workaround that doesn't affect our actual site content, I think it's ok to be kludgy here.

Looking through our internal process doc, I think this change also address the issue with footnote links (they'll now link to the main site rather than internally).  If not, I can look into turning them off. Besides the change in #387, is there anything else y'all can think of that we can conditionally change to make creating the newsletter easier?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,3 +58,9 @@ production-test:
 
 new-topic:
 	_contrib/new-topic
+
+email: clean
+	@ echo
+	@ echo "[NOTICE] Latest Newsletter: http://localhost:4000/en/newsletters/$$(ls _posts/en/newsletters/ | tail -1 | tr "-" "/" | sed 's/newsletter.md//')"
+	@ echo
+	$(MAKE) preview JEKYLL_ENV=email
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -3,6 +3,7 @@
 <html lang="{{ page.lang | default: site.lang | default: "en" }}">
 
   {%- include head.html -%}
+  {% if jekyll.environment == 'email' -%}<base href="https://bitcoinops.org/">{% endif %}
 
   <body>
 
--- a/_plugins/auto-anchor.rb
+++ b/_plugins/auto-anchor.rb
@@ -8,8 +8,9 @@
 # - [Summary](URL): Details
 
 Jekyll::Hooks.register :documents, :pre_render do |post|
-  ## Don't process documents if YAML headers say: "auto_id: false"
-  if post.data["auto_id"] != false
+  ## Don't process documents if YAML headers say: "auto_id: false" or
+  ## we're formatting for email
+  unless post.data["auto_id"] == false || ENV['JEKYLL_ENV'] == 'email'
     post.content.gsub!(/^ *- .*/) do |string|
       ## Find shortest match for **bold**, *italics*, or [markdown][links]
       title = string.match(/\*\*.*?\*\*|\*.*?\*|\[.*?\][(\[]/).to_s


bitschmidty:
> Technically, this change is bad HTML...

Agreed. The copied HTML for the newsletter actually doesn’t involve anything outside the main content (not even the nav), so that violation wont make it out of my browser.

bitschmidty:
tACK 83e2d943587ef6d546b10a889f8a6e5a65ffd981

Thanks for this convenience idea, @harding

jnewbery:
> How to use
> 1. Generate a preview normally, e.g. make preview...

How about adding this target to the Makefile:

```
email:
	## rebuild index.html with no JEKYLL_ENV=email
	rm _site/index.html || true
	bundle exec jekyll build --incremental
	## rebuild most recent newsletter with JEKYLL_ENV=email
	find . -type f -wholename '*en/newsletters/2*' | tail -1 | xargs rm --
	JEKYLL_ENV=email bundle exec jekyll serve --future --drafts --unpublished --incremental
```

Then the process becomes: run `make email`, go to localhost:4000 and click the link to the most recent newsletter. The links in the index page will point to the locally built site, and the links in the most recent newsletter will point to bitcoinops.org.

> Technically, this change is bad HTML. This setting should be in the <head> section, but that's currently sourced from the Ruby gem package that provides our template and so can't be edited without copying the file into our repository.

Perhaps we should upgrade our minima version once https://github.com/jekyll/minima/pull/468 is merged into a release and include this in a custom-head.html file?

harding:
@jnewbery I'm sorry, I just re-read my last comment and it comes across as aggressively critical.  That was not my intention.  I was trying to express my concern about there being several ways in which the command in the makefile version could end up producing a page that doesn't contain the absolute-path links and which might not be noticed until after the emails were sent.  That said, I think there's a fair counterargument that manually typing the environment prefix each week could end up with a typo producing the same problem, so ultimately I think we should just leave the choice to whatever @bitschmidty prefers.

Again, my apologies.

jnewbery:
> @jnewbery I'm sorry, I just re-read my last comment and it comes across as aggressively critical.

That's not at all how it came across to me, but thank you for the additional rationale :heart: 

We can do something in the middle: have a `make email` target that cleans and builds with the `JEKYLL_ENV=email` env variable. That addresses your concerns about my hackey solution. It would take a few extra seconds to build, but that should be tolerable once a week (and is certainly less time than merging to master and waiting for it to be deployed by netlify before copying).

I agree that the decision should be with @bitschmidty

jnewbery:
Makefile target that doesn't hack around with the build:

```
email:
	echo "Latest Newsletter: http://localhost:4000/en/newsletters/$$(ls _posts/en/newsletters/ | tail -1 | tr "-" "/" | sed 's/newsletter.md//')"
	bundle exec jekyll clean
	JEKYLL_ENV=email bundle exec jekyll serve --future --drafts --unpublished --incremental
```

In the make output, there's a link that takes you to the most recent newsletter:

```
Latest Newsletter: http://localhost:4000/<path to most recent newsletter>
```

harding:
Pushed makefile target.  Good idea for printing the link!  That commit can either be dropped or squashed with the first commit depending on what @bitschmidty decides.

jnewbery:
Thanks @harding . ACK 8443680fb5945fbddc03d600a993c9a4cdb37071.

harding:
Rebased on top of master to bring in the changes from #369 so that I could also address #393 (see additional commit in this PR that modifies auto-anchors.rb).

jnewbery:
ACK ff229fdb0c7f71d4d117a5cddcce3cfad251e4a5

@bitschmidty - feel free to squash/merge if you like these changes.

bitschmidty:
tACK ff229fdb0c7f71d4d117a5cddcce3cfad251e4a5

I like the idea of the email make target.

bitschmidty:
Squashed the makefile commit down and merged! Thanks @harding and @jnewbery !

PR #400, opened 2020-05-08
Topics: add new topics for addrv2, output linking, package relay

harding:
This is 3 out of the 16 new topics I was scheduled to have submitted by now.  Sorry for the delay.  :man_facepalming:


_topics/en/addr-v2.md
```diff
@@ -0,0 +1,58 @@
...
+
+  - title: Version 2 `addr` message assigned BIP155
+    url: /en/newsletters/2019/07/31/#bips-766
+    date: 2019-07-31
+
+  - title: Proposed per-node signaling for address relay with v2 `addr` messages
+    url: /en/newsletters/2019/11/06/#signaling-support-for-address-relay
+    date: 2019-11-06
+
+## Optional.  Same format as "primary_sources" above
```
bitschmidty, line 54: Might also be a useful link: https://bitcoin.org/en/developer-reference#addr (or equivalent)

_topics/en/unannounced-channels.md
```diff
@@ -0,0 +1,84 @@
...
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  Most unannounced channels are believed to belong to users that simply
+  don't intend to route payments, such as users of mobile clients that
+  aren't always online to route payments.
+
+  Unannounced channels channels are sometimes called **private
+  channels** but this is [discouraged][unpublished channels delenda est]
+  by some experts because, if the channels aren't used for routing, then
+  the publicly known counteparty of an unannounced channel knows that
```
bitschmidty, line 28: s/counteparty/counterparty

harding:
@bitschmidty ﬁxed, thanks!

_topics/en/unannounced-channels.md
```diff
@@ -0,0 +1,84 @@
...
+  the network for use in routing.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  Most unannounced channels are believed to belong to users that simply
+  don't intend to route payments, such as users of mobile clients that
+  aren't always online to route payments.
+
+  Unannounced channels channels are sometimes called **private
+  channels** but this is [discouraged][unpublished channels delenda est]
```
jnewbery, line 26: I don't fully understand z-man's argument. He says that if a channel is unannounced, then the connected party in that channel knows that all payments to/from that channel are ultimately to/from the other party. He then says that all channels should be announced. But in that case, doesn't the connected party still know that all payments to/from the other party are ultimately to/from the other party if there's only one announced channel to that pubkey? To avoid that, the other party needs to have at least two channels open *and* route payments (otherwise the spy could send probe payments via the victim and see that it never routes and only ever terminates payments).

Z-man's argument is against non-routing nodes rather than against unannounced channels.

_topics/en/unannounced-channels.md
```diff
@@ -0,0 +1,84 @@
...
+
+  - title: "C-Lightning #2230 adds a `private` flag to the `listpeers` RPC"
+    url: /en/newsletters/2019/01/15/#c-lightning-2230
+    date: 2019-01-15
+
+  - title: "C-Lightning #2234 adds route hints for private channels to `invoice` RPC"
+    url: /en/newsletters/2019/01/22/#c-lightning-2234
+    date: 2019-01-22
+
+  - title: "Talk about LN topology, private channels not included"
```
jnewbery, line 56: "Talk about LN topology, private channels not included" is a bit confusing because they're explicitly mentioned in the talk and write-up.

_topics/en/package-relay.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  Without package relay, it's not possible to effectively [CPFP][topic
+  cpfp] fee bump a transaction that's below the minimum feerate nodes
+  accept.  Nodes will reject the parent transaction for its too low
+  feerate and then ignore the fee-bumping child transaction because the
+  parent transaction is needed in order to validate the child.  This is
+  especially problematic considering that the minimum feerate nodes
+  accept varies depending on the contents of each node's memory pool
```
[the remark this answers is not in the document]
harding: And indeed it's already in our style guide.  :facepalm: Thanks!

_topics/en/unannounced-channels.md
harding: Good argument.  I'm going to yank the unannounced-channels topic from this PR, open a new PR with just it, and then invite Zmn to comment.  I don't want to change the name or URL of the topic after publication, so I think it's good to take our time to get it right---but there's no reason to hold up the other topics.

harding:
Edits made for @jnewbery feedback (thanks!).  This includes a new commit that automatically creates Markdown reference-style links for the topics (with the ability to define a *short name* for topics whose long names I don't want to frequently type) and the removal of the *unannounced channels* topic (which I'll put in a separate commit).

jnewbery:
ACK ed307eb8f5b2a06d31dec30fbf0b813461bbcb7e

Looks great. New automatic topics links are :fire:

PR #403, opened 2020-05-15
Topics: add new topic for unannounced channels

harding:
Adds a new topic.  An unresolved issue is what to call this topic: **unannounced channels** or **private channels**.  The name we choose affects the URL and I *really* don't want to break a URL (or even setup a redirect) if we change names later, so let's try to choose a good term now.

## Arguments for "unannounced"

- It's an accurate description of the key attribute of such channels
- It has no particular negative or positive connotations (AFAIK)
- It has use going back to at least [2017](https://lists.linuxfoundation.org/pipermail/lightning-dev/2017-December/000846.html) and a quick search of Lightning-Dev shows it being used by Rusty Russell, Christian Decker, and Bastien Teinturier.  ZmnSCPxj seems to use the related term "unpublished channels".  Those are all significant contributors to LN development and I like to honor contributor terminology choices when they don't get in the way of clear communication

## Arguments for "private"

- It's also an accurate description of the key attribute, although it may give the mistaken impression that private channels are resistant against surveillance (which they may not be, not any more than, say, private property is resistant against surveillance)
- It's the term used in BOLTs 4 and 11
- It seems to be the term used by everyone but the people noted above on the Lightning-Dev mailing per a quick search of the archives via my MUA

---

I originally went with "unannounced" because of [Zmn's argument against such channels](https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-December/002408.html) but @jnewbery [questioned](https://github.com/bitcoinops/bitcoinops.github.io/pull/400#discussion_r424703452) the logic there:

> I don't fully understand z-man's argument. He says that if a channel is unannounced, then the connected party in that channel knows that all payments to/from that channel are ultimately to/from the other party. He then says that all channels should be announced. But in that case, doesn't the connected party still know that all payments to/from the other party are ultimately to/from the other party if there's only one announced channel to that pubkey? To avoid that, the other party needs to have at least two channels open and route payments (otherwise the spy could send probe payments via the victim and see that it never routes and only ever terminates payments).
>
> Z-man's argument is against non-routing nodes rather than against unannounced channels.

I think I still prefer the "unannounced" name because of its lack of association with the word "privacy".  I think it's better to use bland names than give users potentially unreasonable expectations.  However, I don't have a strong opinion about which name we choose as long as it's something we can stick with long term.

### DIFF
--- a/_topics/en/unannounced-channels.md
+++ b/_topics/en/unannounced-channels.md
@@ -1,0 +1,84 @@
+---
+title: Unannounced channels
+
+## Optional.  An entry will be added to the topics index for each alias
+aliases:
+  - Private channels
+
+## Required.  At least one category to which this topic belongs.  See
+## schema for options
+categories:
+  - Lightning Network
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+## Should be less than 500 characters
+excerpt: >
+  **Unannounced channels** are LN channels that are not advertised to
+  the network for use in routing.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  Most unannounced channels are believed to belong to users that simply
+  don't intend to route payments, such as users of mobile clients that
+  aren't always online to route payments.
+
+  An alternative name is **private channels,** but there's contention
+  between experts about whether the channels [improve privacy][privacy
+  in pcns] or [not][unpublished channels delenda est], so it may be
+  preferable to use the universally accepted name "unannounced
+  channels."
+
+  [privacy in pcns]: https://arxiv.org/pdf/1909.02717.pdf "Section 4.1: endpoint uncertainty"
+  [unpublished channels delenda est]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-December/002408.html
+
+## Optional.  Produces a Markdown link with either "[title][]" or
+## "[title](link)"
+#primary_sources:
+#    - title: Test
+#    - title: Example
+#      link: https://example.com
+
+## Optional.  Each entry requires "title", "url", and "date".  May also use "feature:
+## true" to bold entry
+optech_mentions:
+  - title: "LND #1944 tweaks `sendtoroute` RPC to allow routing via private channels"
+    url: /en/newsletters/2018/11/13/#lnd-1944
+    date: 2018-11-13
+
+  - title: "C-Lightning #2230 adds a `private` flag to the `listpeers` RPC"
+    url: /en/newsletters/2019/01/15/#c-lightning-2230
+    date: 2019-01-15
+
+  - title: "C-Lightning #2234 adds route hints for private channels to `invoice` RPC"
+    url: /en/newsletters/2019/01/22/#c-lightning-2234
+    date: 2019-01-22
+
+  - title: "Talk about LN topology and lack of public info about unannounced channels"
+    url: /en/newsletters/2019/09/18/#lightning-network-topology
+    date: 2019-09-18
+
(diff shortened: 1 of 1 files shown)


jnewbery:
I think 'unannounced' is better. Privacy is multidimensional and to some extent subjective, whereas unannounced is an objective description - it's a channel which is opened and not announced on the lightning gossip network.

I like the shorter _unpublished channels delenda est_ summary.

ZmnSCPxj:
>  To avoid that, the other party needs to have at least two channels open and route payments (otherwise the spy could send probe payments via the victim and see that it never routes and only ever terminates payments).

If the party with one *published* channel itself has unpublished / unannounced channels with other peers (who are foolish enough to use them) then it *can* route, and the surveilling node will mis-assign a set of payments as terminating to the one-published-channel node without realizing that in fact this is a routing node with its own unpublished peers.

Which suggests to me that basically the non-publishing nodes get their privacy leeched off by the published peer, thus unpublished channels delenda est.

> Z-man's argument is against non-routing nodes rather than against unannounced channels.

This is correct.  But a node with *one* published channel can still route if it has unpublished peers, whereas a completely unpublished node *cannot* route at all.

In any case "unannounced" / "unpublished" *is* a better term as it does not imply any privacy --- a node might be more concerned about making sure its onchain funds are not linked to its offchain Lightning activity (which is revealed on publication of a channel), and trust a particular Lightning Service Provider to keep mum about its offchain activity (which I would argue is not something you want to rely on --- people change, and since ultimately all Lightning Service Providers are made up of people, so too will they change, and the entity you find trustworthy today may not be trustworthy tomorrow, and yet the logs exists, thus, unpublished channels delenda est).

jonatack:
I prefer "unannounced" as well, for the good reasons discussed above.

harding:
Ok, unanimous agreement so far on naming.  I've rephrased the final paragraph to give the reason for discouraging the name "private channels" as it confusing people into thinking they're more private.  It still links to @ZmnSCPxj post, as I think that's still the best extended explanation of the privacy implications.

jnewbery:
@ZmnSCPxj - thanks for dropping in :)

I really don't understand the _delenda est_ terminology, or for that matter any other argument along the lines of "to use Bitcoin/Lightning the way I want to, I need to prevent others from using Bitcoin/Lightning the way they choose to." We recently saw similar arguments about the mortal threat posed to all users of Bitcoin by people choosing to use BIP 157. I still believe in the approach of trying to build something better, rather than trying to prevent people from using something worse. That approach seems much more likely to succeed in the long run (and much more interesting!)

> If the party with one published channel itself has unpublished / unannounced channels with other peers (who are foolish enough to use them) then it can route

ok, so lets imagine this world where you've successfully managed to _delere_ all unannounced channels. In this world, nodes with announced channels no longer receive this privacy boost. In the world we live in, the fact that unannounced channels exist is enough to provide privacy cover for all announced channels.

> the non-publishing nodes get their privacy leeched

Not true. The privacy/deniability benefits are shared by _all_ announced channels, since any of them could be routing for unannounced channels. Yes, the unannounced channel counterparty does give up some of _its_ privacy for the convenience of not having to be a routing node, which I think is a trade-off that some users want to make. Saying "that's not something I want to use or work on" makes sense to me. Saying "this poses a mortal thread to the entire system" is stretching.

harding:
Good discussion!  I forced pushed another attempt at rewriting the final paragraph to try to fairly summarize the issue in as few words as possible (but with references).

jnewbery:
ACK 90418f647168190f9511e07e55817e9e2bbeb7f1. Looks good to me. Neutral language is best for the topics summaries!

ZmnSCPxj:
ACK 90418f6

@jnewbery argument accepted.

PR #409, opened 2020-05-29
About: remove incorrect Twitter link for Casares

harding:
Current link is for https://twitter.com/wences , which isn't the person we want.  I can't find an account for Xapo's Casares, so this PR just drops the link.  I checked that the page renders fine without the link.

### DIFF
--- a/_data/sponsors.yml
+++ b/_data/sponsors.yml
@@ -4,7 +4,6 @@
 sponsors:
   -
     name: Wences Casares
-    twitter: wences
     avatar: casares.jpg
     quote: >-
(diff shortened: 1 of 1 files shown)


jonatack:
ACK, that Twitter account is now someone else and I didn't find him either. His [crunchbase profile](https://www.crunchbase.com/person/wensceslao-casares) has Facebook and LinkedIn links (yuck...I NoScript both of those) and also the out-of-date Twitter account.

jnewbery:
ACK. This definitely used to be his twitter handle. I was vaguely aware that he'd deactivated his account a while back, but I guess he deleted it entirely and the handle has been recycled. Good catch!

PR #411, opened 2020-06-01
New topic: vaults

harding:


_topics/en/vaults.md
```diff
@@ -0,0 +1,89 @@
...
+## Required.  At least one category to which this topic belongs.  See
+## schema for options
+categories:
+  - Scripts and Addresses
+  - Security Enhancements
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+## Should be less than 500 characters
+excerpt: >
+  **Vaults** are a type of covenant that requires two separate
```
bitschmidty, line 17: Might be better to have the covenant link up here instead of below

_topics/en/vaults.md
```diff
@@ -0,0 +1,89 @@
...
+  to block the second transaction that completes the spend.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  Usually, there's a minimum amount of time or number of blocks that
+  must pass between the two transactions, giving the user that amount of
+  time to notice if someone stole their private key and is attempting to
+  steal their money.  If the user detects the theft attempt, most vault
+  designs also allow the user to either send the money to a safe address
+  that uses an extra secure script or to permanently destroy the money
```
bitschmidty, line 30: extra secure script is ambiguous between an additional script or super-secure script

_topics/en/vaults.md
harding: I agree this would be better, but we can't have links in the excerpt (this is a restriction of Jekyll, not ours, although we could work around this if we had to).

jnewbery:
I've rebased and added the vaults links to the `OP_CHECKSIGFROMSTACK` and `OP_CHECKTEMPLATEVERIFY` topics pages.

_topics/en/vaults.md
harding: Fixed.  Thanks!

bitschmidty:
ACK 1d10c3c32fc454a7cf53917fc7f8ee261cceaf18

_topics/en/vaults.md
jnewbery, line 17: Should this be 'require' to agree with 'vaults'. Alternatively: "A vault is a type of covenant that requires"

_topics/en/vaults.md
```diff
@@ -0,0 +1,89 @@
...
+excerpt: >
+  **Vaults** are a type of covenant that requires two separate
+  transactions to appear in two different blocks in order for a user to
+  spend money from their wallet.  The first transaction signals that
+  someone is attempting to spend the money and gives the user a chance
+  to block the second transaction that completes the spend.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  Usually, there's a minimum amount of time or number of blocks that
```
jnewbery, line 25: There seems to be some difference between the excerpt "transactions ... appear in two different blocks" and this "Usually, there's a minimum amount of time or blocks".

The first states that the transactions must appear in different blocks. The second implies with the "usually" that there could be cases where the transactions appear in the same block.

_topics/en/vaults.md
```diff
@@ -0,0 +1,89 @@
...
+  time to notice if someone stole their private key and is attempting to
+  steal their money.  If the user detects the theft attempt, most vault
+  designs also allow the user to either send the money to a safe address
+  that uses a more secure script or to permanently destroy the money
+  to prevent the thief from profiting from their attack.
+
+  Some vault designs rely on [covenants][topic covenants] that require
+  consensus changes to Bitcoin.  Other vault designs use existing
+  protocol features plus techniques such as multisig, presigned
+  transactions, and ephemeral private keys.
```
jnewbery, line 36: I don't think 'ephemeral' emphasizes the right thing. The critical part is that the keys are **deleted**.

PR #415, opened 2020-06-03
Newsletters: News100: add donation link for Chris Belcher

harding:
@chris-belcher asked if we could add a donation link to the Coinswap implementation description.


harding:
Including this donation link would set a precedent that we include such links, which would necessitate that we evaluate the worthiness of future requests.  In this particular case, I think Belcher has proven to be entirely worthy, but there are lots of people and projects out there that I consider to be dubious or even sleezy, and we don't need the drama of internal Optech debates about who's worthy of what or external pressure to include certain links (or external shame for not including certain links).  It's best that we simply focus on technical news.

(I've also sent a separate apology to Belcher for my flip-flopping on this issue.)

PR #416, opened 2020-06-06
Change to common *input* ownership assumption

MaxHillebrand:
As far as I know the term, as well as in the wiki that is linked, the term discussed here is called the `common input ownership assumption`, and not `common wallet ownership assumption`. 

If I am mistaken here, and `wallet` is the intended word, then please dismiss and close the PR.

---

Congrats on #100 - this is incredible work!


harding:
Yeah, I agree with the change and I don't know why I said "wallet" instead of "input".  However, this PR also needs to make the same change to the link definition at the bottom of the email or the newsletter won't render correctly (that's why the tests are failing).

Thanks for the PR and the congrats!

jonatack:
Thanks @MaxHillebrand, good catch! Can you force-push after updating the link name at the bottom?

MaxHillebrand:
Sorry, I missed that link reference at the bottom... It's added now, but in a separate commit. Is this alright if you now do a squash merge commit @jonatack?

jnewbery:
Squashed and merged. Thanks Max!

PR #417, opened 2020-06-08
Newsletters: add 101 (2020-06-10)

harding:
- [x] Bitcoin Core PR review club section @jonatack 
- [x] `Eclair #1440` @adamjonas 
- [x] `LND #4251` @dongcarl

PR #419, opened 2020-06-09
news101: review club monthly feature

jonatack:
The writeup is in commit a3ea3e1d7ea770332736a2a3667ef6bd14442035


harding:
Added to #417.  Thanks!

PR #422, opened 2020-06-15
Update miniscript, rpc references

bitschmidty:
1. updates miniscript link to miniscript topic
2. updates rpc base url to 0.20

If there is a cleaner way to do # 1, please opine, @harding


harding:
Untested ACK.  # 1 looks perfect to me.

jnewbery:
I think https://github.com/bitcoinops/bitcoinops.github.io/blame/a1d67247e742f839a4354a27ca15b63d6d6b9f07/_posts/en/newsletters/2019-08-28-newsletter.md#L86 is the only place where the new miniscript link makes the text incorrect. You could fix that by changing the text to say 'see our miniscript topic page' or similar.

bitschmidty:
Pushed a commit to implement @jnewbery 's suggestion (good idea!).

PR #432, opened 2020-07-07
WIP: style guide updates

jonatack:
Various items noticed while reviewing; open for discussion.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -36,8 +36,11 @@ sentence or heading.
 - Introduce sighash with "signature hash (sighash)", but don't introduce sighash
   types (`SIGHASH_SINGLE`, `SIGHASH_ALL`, etc) since they're technical terms
   and not abbreviations.
+- IBD (Initial Block Download)
+- HTLC (Hash Time Locked Contract)
+- PTLC (Point Time Locked Contract)
 
-#### Unintroduced abreviations
+#### Unintroduced abbreviations
 
 The first occurrence of an abbreviation in a document should be preceded
 by its full name (e.g. Foo Bar Baz (FBB)) with the exception of the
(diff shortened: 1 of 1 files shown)


jonatack:
Maybe also "on-chain" instead of "onchain' if there's consensus.

harding:
+1 for HTLC, since (as Jon noted elsewhere) different people expand that abbreviation differently
+1 for IBD, since the Bitcoin Core source code sometimes expands that to "Initial **Block** Download" and othertimes to "Initial **Blockchain** Download"

+0.5 (no strong opinion) on PTLC, since that seems less likely to end up inconsistent but there's no harm in adding it

+100 for the actual spelling typo fix

-0.1 for changing onchain to on-chain (no strong opinion, I just don't want to change it unless other people strongly desire it)

jnewbery:
Here's a controversial take for you: in our own writing we should prefer _Initial Sync_ over _Initial Block Download_. That's because the thing we care about is synchronizing our view of the block chain and UTXO set to the network consensus view. _Initial Block Download_ implies that the only thing we care about is downloading the blocks.

harding:
Per a quick grep, we actually already use "initial sync" more than "IBD", so I don't think it's that controversial a take.  :-)  I prefer *sync* for the same reason you specify; I was probably led astray last week by the title of the PR including "IBD".

The real controversy is dropping the "initial" from "initial sync" since it's a mode we use each time the node is more than 100 blocks behind the probable chain tip.  :-)

jnewbery:
> The real controversy is dropping the "initial" from "initial sync" since it's a mode we use each time the node is more than 100 blocks behind the probable chain tip. :-)

Ah! But it's a latch in memory: https://github.com/bitcoin/bitcoin/blob/abdfd2d0e3ebec7dbead89317ee9192189a35809/src/validation.cpp#L1299-L1301 which means that whenever you start bitcoind it's initially in initial block download. We don't ever revert to being in IBD while running.

jnewbery:
Thanks @jonatack !

harding:
> whenever you start bitcoind it's initially in initial block download. We don't ever revert to being in IBD while running.

Interesting; I didn't know that.

jonatack:
I agree, *initial sync* is fine. *IBD* may have caught on because of the handy acronym that rolls off the tongue well. For "initial sync" there's also an abbreviated version: https://www.youtube.com/channel/UCjkyfFH-MWZhasolgds05EA

PR #435, opened 2020-07-13
Topics: add adaptor signatures

harding:
Don't let the confident tone of this text fool you---it's author still feels he has a very poor understanding of adaptor sigs.  I've tried to make my description as easy to prove wrong as possible (while also hopefully keeping it readable by non-experts), but please don't assume that its detail and assertiveness are proxies for correctness.

### DIFF
--- a/_topics/en/adaptor-signatures.md
+++ b/_topics/en/adaptor-signatures.md
@@ -18,7 +18,7 @@ categories:
 ## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
 ## Should be less than 500 characters
 excerpt: >
-  **Adaptor signatures** are auxiliary signature data that commit to a a
+  **Adaptor signatures** are auxiliary signature data that commit to a
   hidden value.  When an adaptor signature is combined with a final
   signature, it reveals the hidden value.  Alternatively, when combined
   with the hidden value, it reveals the final signature.  The commitment
@@ -42,7 +42,7 @@ extended_summary: |
   Alice hasn't paid Bob yet.
 
   What the adaptor signature does provide Bob is a commitment to Alice's
-  hidden value.  This commitment is made in the form of a value Bob can
+  hidden value.  This commitment is made in the form of a value Bob
   can use to create a second adaptor signature that commits to the same
   hidden value as Alice's adaptor signature.  Bob can make that
   commitment even without knowing Alice's hidden value or his own final
@@ -54,7 +54,7 @@ extended_summary: |
   transaction that pays her.  She broadcasts the transaction and
   receives Bob's payment.  When Bob sees that transaction onchain, he
   can combine its final signature with the adaptor signature he gave
-  Alice, allowing him to to derive the hidden value.  Then he can
+  Alice, allowing him to derive the hidden value.  Then he can
   combine that hidden value with the adaptor signature Alice gave him to
   get her final signature for the first transaction.  Bob broadcasts
   that transaction to receive Alice's payment, completing the coinswap.
@@ -106,7 +106,7 @@ extended_summary: |
     rather than the needed `R`.
 
   Therefore Bob can't use the adaptor signature as a final signature.
-  However, he can create his own signature using it.  This is similar to
+  However, he can create his own signature using it.  This is similar to the
   signature Alice created but Bob doesn't commit to `t` here, since Bob
   doesn't know that value.  All variables here except `T` are different
   for Bob than they were for Alice:
@@ -152,7 +152,7 @@ extended_summary: |
   Alice could double spend her payment to Bob after she learned Bob's
   final signature, or Bob could've tried the same in reverse (with more
   difficulty since we assumed Alice's transaction had one confirmation).
-  This is usually addressed by combining adaptor signatures with
+  This is typically addressed by combining adaptor signatures with
   multiparty signatures.  For example, Alice deposits her money into an
   address that requires signatures from both her and Bob to spend.  Now
   Alice can provide Bob with an adaptor signature for her half of the
--- a/_topics/en/adaptor-signatures.md
+++ b/_topics/en/adaptor-signatures.md
@@ -254,4 +254,7 @@ see_also:
 
   - title: Using schnorr subtraction to create more private coinswaps
     link: https://joinmarket.me/blog/blog/flipping-the-scriptless-script-on-schnorr/
+
+  - title: Adaptor signatures for discreet log contracts
+    link: https://lists.launchpad.net/mimblewimble/msg00485.html
 ---
--- a/_topics/en/adaptor-signatures.md
+++ b/_topics/en/adaptor-signatures.md
@@ -1,0 +1,242 @@
+---
+title: Adaptor signatures
+
+## Optional.  Shorter name to use for reference style links e.g., "foo"
+## will allow using the link [topic foo][].  Not case sensitive
+# shortname: foo
+
+## Optional.  An entry will be added to the topics index for each alias
+#aliases:
+
+## Required.  At least one category to which this topic belongs.  See
+## schema for options
+categories:
+  - Contract Protocols
+  - Privacy Enhancements
+  - Scripts and Addresses
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+## Should be less than 500 characters
+excerpt: >
+  **Adaptor signatures** are auxiliary signature data that commit to a a
+  hidden value.  When an adaptor signature is combined with a final
+  signature, it reveals the hidden value.  Alternatively, when combined
+  with the hidden value, it reveals the final signature.  The commitment
+  may be reused in other signatures without the secondary signers
+  knowing the hidden value or their own final signature.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  Adaptor signatures never need to be published onchain.   To anyone
+  without a corresponding adaptor signature, a final signature looks
+  like any other digital signature, giving adaptor signatures efficiency
+  and privacy advantages over other onchain locking mechanisms such as
+  hashlocks and preimages.
+
+  ### Example
+
+  The multiple uses of adaptor signatures can be seen in a simple
+  coinswap protocol.  For example, Alice can give Bob an adaptor
+  signature for a transaction that promises to pay him 1 BTC.  An
+  adaptor signature by itself can't be used as a final signature, so
+  Alice hasn't paid Bob yet.
+
+  What the adaptor signature does provide Bob is a commitment to Alice's
+  hidden value.  This commitment is made in the form of a value Bob can
+  can use to create a second adaptor signature that commits to the same
+  hidden value as Alice's adaptor signature.  Bob can make that
+  commitment even without knowing Alice's hidden value or his own final
+  signature for that commitment.  Bob gives Alice his adaptor signature
+  and a corresponding transaction that promises to pay her 1 BTC.
+
+  Alice has always known the hidden value, so she can combine the hidden
+  value with Bob's adaptor signature to get his final signature for the
+  transaction that pays her.  She broadcasts the transaction and
+  receives Bob's payment.  When Bob sees that transaction onchain, he
+  can combine its final signature with the adaptor signature he gave
+  Alice, allowing him to to derive the hidden value.  Then he can
+  combine that hidden value with the adaptor signature Alice gave him to
+  get her final signature for the first transaction.  Bob broadcasts
+  that transaction to receive Alice's payment, completing the coinswap.
+
+  Besides coinswaps, there are several other [proposed uses][scriptless
+  scripts repo] for adaptor signatures.
+
+  <div class="qa_details">
+  <details markdown="1"><summary>Click to show the same coinswap example in mathematical terms</summary>
+  *In the following example, we assume the use of BIP340-compatible
+  schnorr signatures.  We use lowercase variables for scalars and
+  uppercase variables for elliptic curve points.  We represent
+  concatenation with `||` and the hash function with `H()`.*
+
+  Alice creates a valid final signature (`s`) for the transaction paying Bob
+  (`m`) using her private key (`p`), which corresponds to her public key
+  (`P = pG`).  She also uses a private random nonce (`r`), a hidden value
+  (`t`), and the elliptic curve points for them (`R = rG, T = tG`):
+
+      s = r + t + H(P || R + T || m) * p
+
+  She removes `t` from the final signature to produce an adaptor
+  signature:
+
+      s' = s - t
+
+  She gives Bob the adaptor signature, which consists of the following
+  data:
+
+      s', R, T
+
+  Bob can verify the adaptor signature:
+
+      s' * G ?= R + H(P || R+T || m) * P
+
+  But it's not a valid signature.  For a valid signature, BIP340 expects
+  `x` and `Y`, using them with the expression:
+
+      x * G ?= Y + H(P || Y || m) * P
+
+  However,
+
+  - If Bob sets `Y=R` so that it matches the `s'` he received in the
+    adaptor signature, then BIP340 is going to fail on `H(P || R || m)`
+    since Alice computed her hash with `H(P || R + T || m)`.
+
+  - If Bob sets `Y=R + T` so that it matches `H(P || R + T || m)`, BIP340
+    is going to fail on the initial `Y` since Bob is providing `R + T`
+    rather than the needed `R`.
+
+  Therefore Bob can't use the adaptor signature as a final signature.
+  However, he can create his own signature using it.  This is similar to
+  signature Alice created but Bob doesn't commit to `t` here, since Bob
+  doesn't know that value.  All variables here except `T` are different
+  for Bob than they were for Alice:
+
+      s = r + H(P || R + T || m) * p
+
+  Unlike Alice, Bob doesn't need to tweak his signature.  Bob's `s` is
+  not a valid signature because it commits to `r` and `R + T`, which
+  won't pass BIP340 verification for the same reasons previously described.
+  To be valid, the signature needs to commit to `r + t` and `R + T`,
+  which Bob can't produce since he doesn't know `t`.
+
+  Bob gives Alice his adaptor signature:
+
+      s, R, T
+
+  Alice already knew `T`, but `(s, R, T)` is a standard adaptor
+  signature so we use its full form.  Alice can produce a final
+  signature from that adaptor signature using the hidden `t` value that
+  only she knows so far:
+
+      (s + t) * G ?= R + T + H(P || R + T || m) * P
+
+  Alice uses the final signature to broadcast Bob's transaction that
+  pays her.  When Bob sees `(s + t)` onchain, he can learn the value of `t`:
+
+      t = (s + t) - s
+
+  He can then use `t` to solve the adaptor signature Alice gave him
+  earlier:
+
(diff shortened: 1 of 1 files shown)


_topics/en/adaptor-signatures.md
```diff
@@ -0,0 +1,242 @@
...
+categories:
+  - Contract Protocols
+  - Privacy Enhancements
+  - Scripts and Addresses
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+## Should be less than 500 characters
+excerpt: >
+  **Adaptor signatures** are auxiliary signature data that commit to a a
+  hidden value.  When an adaptor signature is combined with a final
```
adamjonas, line 22: Maybe another sentence for this intro section and motivation? Something like: "These are partial signatures that, if completed, will reveal valuable information to the signer. (In practice | For example ), when ..." (taken from abstract of [here](https://pdfs.semanticscholar.org/fdef/41889571e22f86e95a6eb1838aaa06bb5737.pdf))

_topics/en/adaptor-signatures.md
```diff
@@ -0,0 +1,242 @@
...
+extended_summary: |
+  Adaptor signatures never need to be published onchain.   To anyone
+  without a corresponding adaptor signature, a final signature looks
+  like any other digital signature, giving adaptor signatures efficiency
+  and privacy advantages over other onchain locking mechanisms such as
+  hashlocks and preimages.
+
+  ### Example
+
+  The multiple uses of adaptor signatures can be seen in a simple
```
adamjonas, line 38: In other explanations I've heard (e.g. https://youtu.be/jWAhDH07bUY?t=2647), understanding the signature as "offset" by some random nonce has helped me better conceptualize how these signatures can be combined and subtracted.

_topics/en/adaptor-signatures.md
```diff
@@ -0,0 +1,242 @@
...
+  </div>
+
+  ### Relationship to multiparty signatures
+
+  Adaptor signatures usually can't secure a contract entirely by
+  themselves.  For example, in the above description of a coinswap,
+  Alice could double spend her payment to Bob after she learned Bob's
+  final signature, or Bob could've tried the same in reverse (with more
+  difficulty since we assumed Alice's transaction had one confirmation).
+  This is usually addressed by combining adaptor signatures with
```
adamjonas, line 155: nit: usually repetition
```suggestion
  Adaptor signatures usually can't secure a contract entirely by
  themselves.  For example, in the above description of a coinswap,
  Alice could double spend her payment to Bob after she learned Bob's
  final signature, or Bob could've tried the same in reverse (with more
  difficulty since we assumed Alice's transaction had one confirmation).
  This is typically addressed by combining adaptor signatures with
```

harding:
@adamjonas thanks for the review!  Obviously, I either need a new glasses prescription, a duplicate words CI test, or both.  :-)

real-or-random:
Oh, interesting, I'll try to have a look.

_topics/en/adaptor-signatures.md
```diff
@@ -0,0 +1,242 @@
...
+      s' * G ?= R + H(P || R+T || m) * P
+
+  But it's not a valid signature.  For a valid signature, BIP340 expects
+  `x` and `Y`, using them with the expression:
+
+      x * G ?= Y + H(P || Y || m) * P
+
+  However,
+
+  - If Bob sets `Y=R` so that it matches the `s'` he received in the
```
jonatack, line 100: maybe `Y = R` (spaces around the equal sign) for consistency, idem for line 104

_topics/en/adaptor-signatures.md
```diff
@@ -0,0 +1,242 @@
...
+  Alice, allowing him to derive the hidden value.  Then he can
+  combine that hidden value with the adaptor signature Alice gave him to
+  get her final signature for the first transaction.  Bob broadcasts
+  that transaction to receive Alice's payment, completing the coinswap.
+
+  Besides coinswaps, there are several other [proposed uses][scriptless
+  scripts repo] for adaptor signatures.
+
+  <div class="qa_details">
+  <details markdown="1"><summary>Click to show the same coinswap example in mathematical terms</summary>
```
jonatack, line 66: perhaps s/show/see/

_topics/en/adaptor-signatures.md
```diff
@@ -0,0 +1,242 @@
...
+  </div>
+
+  ### Relationship to multiparty signatures
+
+  Adaptor signatures usually can't secure a contract entirely by
+  themselves.  For example, in the above description of a coinswap,
+  Alice could double spend her payment to Bob after she learned Bob's
+  final signature, or Bob could've tried the same in reverse (with more
+  difficulty since we assumed Alice's transaction had one confirmation).
+  This is typically addressed by combining adaptor signatures with
```
jonatack, line 155: perhaps s/This/This issue|insecurity, etc./

_topics/en/adaptor-signatures.md
```diff
@@ -0,0 +1,242 @@
...
+  address that requires signatures from both her and Bob to spend.  Now
+  Alice can provide Bob with an adaptor signature for her half of the
+  multiparty transaction, which Bob can accept with perfect safety
+  knowing that Alice couldn't double spend the transaction without a
+  signature from him.  This may additionally require a timelocked refund
+  option in case one party refuses to sign.
+
+  In the [schnorr signature scheme][topic schnorr signatures], adaptor
+  signatures are usually proposed to be combined with signature
+  aggregation schemes such as [MuSig][topic musig] to allow the final
```
jonatack, line 166: s/Musig/musig/ per style guide, idem for the link in line 239 below (and the topic page needs updating WRT this)

harding: Huh?  Style guide says:

> Proper nouns
> [...]
> MuSig (the pubkey and signature aggregation scheme. Don't refer to Musig, muSig or musig)

(I remember this one because I voted for lowercase "musig" but lost.  :-)

harding:
@jonatack the only resource I used that's not listed here---but which was quite influential on my (uncertain) understanding---was @AdamISZ's https://joinmarket.me/blog/blog/flipping-the-scriptless-script-on-schnorr/ .  In particular, I used his nomenclature for the initial equations I started with (and then tweaked them to make more sense to me, e.g. he used `x` for the private key and `P` for the corresponding public key, but it made sense to me to use the same letter for both, just with different capitalization).  I didn't include his post in our `see_also` links because I usually want to focus on specifications rather than explainers, but his post was so useful to me that I probably should make an exception.

jonasnick:
Another resource (that is not linked here) is the scriptless scripts repo that contains [Andrew's writeup on atomic swaps with adaptor sigs](https://github.com/ElementsProject/scriptless-scripts/blob/master/md/atomic-swap.md) as well as a few other adaptor sig techniques.

_topics/en/adaptor-signatures.md
jonatack: Quite right; I misremembered (note bene, seems Blockstream calls it MuSig).

harding:
@jonasnick it's linked in this PR at https://github.com/bitcoinops/bitcoinops.github.io/pull/435/files#diff-aae86f4a434defe011f0ddb578bcc697R176 .  I found that document particularly useful in comparison with waxwing's blog post; reconciling the slight differences between those two descriptions gave me a valuable opportunity to test my understanding.

_topics/en/adaptor-signatures.md
real-or-random: To be honest, I think the term "final signature" is not necessary. We have a term for this thing, namely "digital signature". 

Tbh, I think the term "adaptor signature" is misleading, it's rather a signature adaptor. But I guess we can't change this. That being said, I think it's clever to avoid the plain term "signature" below, unless it's really clear from the context. I think it should always be "adaptor signature" or "digital signature" (or "Schnorr signature"/"BIP340 signature")

AdamISZ: Yes, I've also multiple times expressed concern that "adaptor signature" will lead to people thinking of it as in some way a signature, when it is about as far from a signature as possible, considering it's trivially unsound/forgeable (and indeed that's kind of the point!), taken on its own.
Even more confusing is the fact that the concept of "partial signature" is important in context, because we are usually talking about using adaptors within a multisig context, and it's very easy i think to be confused between partial signature and adaptor (signature). (And then even *more* confusing is the recent single-signer adaptor signature usage, as per Fournier's recent work - which is single signer ... but still used in multisig ...).
I kind of warmed to Fournier's otVES name. It's somehow a much clearer way to express what we are actually talking about. (paper: https://github.com/LLFourn/one-time-VES/blob/master/main.pdf ).

_topics/en/adaptor-signatures.md
```diff
@@ -0,0 +1,242 @@
...
+  get her final signature for the first transaction.  Bob broadcasts
+  that transaction to receive Alice's payment, completing the coinswap.
+
+  Besides coinswaps, there are several other [proposed uses][scriptless
+  scripts repo] for adaptor signatures.
+
+  <div class="qa_details">
+  <details markdown="1"><summary>Click to show the same coinswap example in mathematical terms</summary>
+  *In the following example, we assume the use of BIP340-compatible
+  schnorr signatures.  We use lowercase variables for scalars and
```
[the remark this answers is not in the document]
harding: We decided to lowercase schnorr (and other names like merkle) in our style guide: https://github.com/bitcoinops/bitcoinops.github.io/blob/master/STYLE.md#common-nouns

_topics/en/adaptor-signatures.md
```diff
@@ -0,0 +1,242 @@
...
+
+  ### Relationship to multiparty signatures
+
+  Adaptor signatures usually can't secure a contract entirely by
+  themselves.  For example, in the above description of a coinswap,
+  Alice could double spend her payment to Bob after she learned Bob's
+  final signature, or Bob could've tried the same in reverse (with more
+  difficulty since we assumed Alice's transaction had one confirmation).
+  This is typically addressed by combining adaptor signatures with
+  multiparty signatures.  For example, Alice deposits her money into an
```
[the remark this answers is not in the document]
harding: I've been using "multiparty signature" to distinguish collaboratively-generated signatures (e.g. MuSig) from the independently-generated signatures glued together with script (e.g. `OP_CHECKMULTISIG`) that we tend to call "multisig" in Bitcoin.  In that context, is this suggestion a minor terminology problem, or do you think it's likely to lead to misunderstadings?

_topics/en/adaptor-signatures.md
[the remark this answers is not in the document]
real-or-random: Well, I think that's strange but okay...

_topics/en/adaptor-signatures.md
```diff
@@ -0,0 +1,242 @@
...
+  signature from him.  This may additionally require a timelocked refund
+  option in case one party refuses to sign.
+
+  In the [schnorr signature scheme][topic schnorr signatures], adaptor
+  signatures are usually proposed to be combined with signature
+  aggregation schemes such as [MuSig][topic musig] to allow the final
+  multiparty signature to look like a single-party signature, enhancing
+  privacy and efficiency.  This is also possible in ECDSA but it
+  requires additional security assumptions and relatively novel
+  algorithms.  Instead, an [alternative scheme][otves] for adaptor
```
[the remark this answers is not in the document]
harding: It's described in section 4.2 of the One-Time Verifiably Encrypted Signatures paper linked here.  Is there another link you're thinking of?

harding:
Thanks everyone for the reviews so far!  I've left comments on the suggested changes I'm skeptical about but everything else sounds reasonable to me and I'll try to make those changes early next week.

_topics/en/adaptor-signatures.md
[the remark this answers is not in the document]
real-or-random: That makes sense. I don't think it will lead to misunderstandings. It's just a mess that we have settled on so many terms that are used differently by different communities, and that are not often not very well defined or used imprecisely. I believe in particular the latter is one of the main reasons why it's hard to read technical texts in the Bitcoin space.

_topics/en/adaptor-signatures.md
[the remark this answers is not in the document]
real-or-random: Oh sure, I missed this. I think you should also add https://eprint.iacr.org/2018/472.pdf  . Not sure if it's worth adding, but also see Appendix A in https://eprint.iacr.org/2020/476.pdf for an overview of research covering this, including criticism of the OT VES paper.

_topics/en/adaptor-signatures.md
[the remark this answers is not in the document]
harding: Do you capitalize "algorithm" just because it's named after ["Algorithmi"](https://en.wikipedia.org/wiki/Muhammad_ibn_Musa_al-Khwarizmi)?  :-)  The general rule in most style guides is to stop capitalizing eponyms after they enter common use---and I think we're well beyond the point for a technical publication about Bitcoin where schnorr signatures and various merkle things are in common use.

_topics/en/adaptor-signatures.md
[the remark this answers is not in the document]
harding: Nice, thanks!

_topics/en/adaptor-signatures.md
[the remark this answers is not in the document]
real-or-random: Yeah, idk, I believe it just seems strange to me because noone in academia does this.

If you ask me for an argument, Schnorr and Merkle are living persons. I believe it's somewhat strange to not spell their names as names. On the other hand it's a honor then that their names are now proper words, so there are good arguments for either side. :)

jnewbery:
> Tbh, I think the term "adaptor signature" is misleading, it's rather a signature adaptor. 

If everyone agrees with this we could rename the topic to "signature adaptors" (and have an alias from "adaptor signatures")

harding:
"Signature adaptors" sounds fine to me.

real-or-random:
I'm not sure. I'd be careful with changing it now, in particular based on a suggestion I spend maybe 30 seconds on. It may not be a good term either, and even if, having two terms creates confusion, too.

real-or-random:
@apoelstra fyi

_topics/en/adaptor-signatures.md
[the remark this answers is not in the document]
harding: I've skimmed both of the papers linked above.  Although they both contain formalized descriptions of signature adaptors, they're also both focused on their use in payment channel networks.  I've already started writing a topic for PTLCs and I think that'd probably be a better place to include links to these two papers.  Does that seem reasonable?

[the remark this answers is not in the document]
real-or-random: Indeed, that seems reasonable. I think as soon is this ready, it'll be also a good idea to have links between the two topics. (I guess you had this idea already.)

harding:
I've pushed an edit for (hopefully) all comments to date.  Thank you everyone for the reviews!  This edit does not change the name of the topic but does use the "signature adaptor" phrase throughout the document.  If people like it, I think we can discuss using that as the main title of the topic and making "adaptor signatures" the AKA.

_topics/en/adaptor-signatures.md
```diff
@@ -18,153 +20,164 @@ categories:
 ## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
 ## Should be less than 500 characters
 excerpt: >
-  **Adaptor signatures** are auxiliary signature data that commit to a
-  hidden value.  When an adaptor signature is combined with a final
-  signature, it reveals the hidden value.  Alternatively, when combined
-  with the hidden value, it reveals the final signature.  The commitment
-  may be reused in other signatures without the secondary signers
-  knowing the hidden value or their own final signature.
+  **Adaptor signatures** (also called **signature adaptors**) are
```
AdamISZ, line 23: Did you mean to write that the other way round I think?

harding: For now, I've left the page title as "adaptor signatures", so I think it makes sense for that to be the primary term in the description.  If we decide to retitle the page, I'll reverse the term and the alias.

AdamISZ: Oh sorry I didn't read carefully what you wrote in the comment above. OK got it, thanks. And yes :+1: from me for 'signature adaptors', although the confusion will continue into perpetuity is my prediction :)

real-or-random:
> I've pushed an edit for (hopefully) all comments to date. Thank you everyone for the reviews! This edit does not change the name of the topic but does use the "signature adaptor" phrase throughout the document. 

I like that it's an easy solution for the "final signature" thing. The text is much easier to read now because not everything is called a signature.

> If people like it, I think we can discuss using that as the main title of the topic and making "adaptor signatures" the AKA.

I think both is fine in the end but I'd stick adaptor signatures as the title, because this is what people will look for. But in either case, I suggest adding a comment that explains that "adaptor signatures" is the standard term but the text here uses "signature adaptor" because it fits better and it simplifies the text.

real-or-random:
Sorry if this holds things up further but maybe @LLFourn wants to comment, too.

LLFourn:
Hey nice work. I'm glad you are taking the time to explain this super important topic. I think this gives a very good explanation in terms of the operations involved in the Schnorr scheme.
If you were to go into more detail, I always like to point out that adaptor signatures are primarily used in two distinct ways:

1. To do _fair exchange of signatures_. Alice gets a signatures on transaction A only in the case Bob gets a signature on transaction B (because when Alice broadcasts A, Bob can get the signature on B). This use is what you cover at the beginning. 
2. As a _signature encryption_ scheme. Alice can give a signature to Bob he can't access until he learns Y. This is for cases where the secret revelation aspect is not important, for example, in the adaptor version of discrete log contracts[1].

[1]: https://lists.launchpad.net/mimblewimble/msg00485.html

_topics/en/adaptor-signatures.md
```diff
@@ -0,0 +1,257 @@
...
+  adaptor by itself can't be used as a BIP340 signature, so
+  Alice hasn't paid Bob yet.
+
+  What the adaptor does provide Bob is a commitment to Alice's hidden
+  value.  This commitment includes a parameter Bob can use to create a
+  second adaptor that commits to the same hidden value as Alice's
+  adaptor.  Bob can make that commitment even without knowing Alice's
+  hidden value or his own signature for that commitment.  Bob gives
+  Alice his adaptor and a corresponding unsigned transaction that
+  promises to pay her 1 BTC.
```
LLFourn, line 64: The interesting thing is that from this adaptor "commitment" is that if you have the "hidden value" you can extract something valuable from it. As far as my reading of cryptography goes  whenever we have have such a commitment it's called an "encryption" and the "hidden value" is called a "decryption key". This would make the "parameter" that you can re-use to create another "commitment" a "public encryption key".

I understand that this comment is probably very helpful at this stage but I couldn't resist pointing this out as I think this paragraph demonstrates how difficult it is to talk about this without using public key encryption terminology.

_topics/en/adaptor-signatures.md
LLFourn: heh you'll be happy to know that in https://eprint.iacr.org/2020/476.pdf  there is now additionally something called a "pre-signature" and an "adaptor" actually refers to the party that does the "adapting"!

I think it's fine if everyone sticks with adaptor signature as it's a rather well know term by now. otVES is a bit wonky. If I could do over I would just go with _one-time signature encryption_. I'm still working on how best to explain it. Whenever I try to explain it 1-on-1 I start of with "adaptor signatures are a kind of signature encryption" and go from there. That seems to work.

jnewbery:
Thanks @harding and thanks to all our guest reviewers!

PR #436, opened 2020-07-13
Test: duplicate words check (with fixes for existing duplicates)

harding:
I appear to to have a a problem typing the the same word twice.  @adamjonas has has been catching most of of these, but but hopefully this this test will allow me to to address my own failings.

Test produced one false positive on on our existing documents, which I addressed with with a small rewrite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,21 @@ test-before-build:
 	## Check for unnecessarily fully qualified URLs (breaks local previews and internal link checking)
 	! git --no-pager grep -- '^\[.*]: https://bitcoinops.org' | grep -v skip-test | grep .
 	! git --no-pager grep -- '](https://bitcoinops.org' | grep -v skip-test | grep .
+	## Check for duplicate words
+	@ ## Only applies to *.md files
+	@ ## Add <!-- skip-duplicate-words-test --> to same line to skip this test
+	@ ## Ignores any strings with non alphabetical characters
+	export LC_ALL=C ; ! git ls-files '*.md' | while read file ; do \
+	    cat $$file \
+	      | sed '/skip-duplicate-words-test/d' \
+	      | sed '/^#/d' \
+	      | tr ' ' '\n' \
+	      | sed 's/ *//g;' \
+	      | sed 's/^.*[^a-zA-Z].*/ /; /^$$/d;' \
+	      | uniq -d \
+	      | sed '/^ $$/d' \
+	      | sed "s|.*|Duplicate word in $$file: &|" ; \
+	done | grep .
 
 	## Check that newly added or modifyed PNGs are optimized
 	_contrib/travis-check-png-optimized.sh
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -32,6 +32,19 @@ test-before-build:
 	## Check for unnecessarily fully qualified URLs (breaks local previews and internal link checking)
 	! git --no-pager grep -- '^\[.*]: https://bitcoinops.org' | grep -v skip-test | grep .
 	! git --no-pager grep -- '](https://bitcoinops.org' | grep -v skip-test | grep .
+	## Check for duplicate words
+	@ ## Only applies to *.md files
+	@ ## Add <!-- skip-test --> to same line to skip this test
```
jnewbery, line 37: Perhaps change this to skip-duplicate-words-test or similar in case we add any more linters.

bitschmidty:
On OSX (10.15.5) this resulted in a series of 

`sed: RE error: illegal byte sequence`

outputs when building. 

Troubleshooting via https://stackoverflow.com/questions/19242275/re-error-illegal-byte-sequence-on-mac-os-x/23584470 by prefixing sed commands with locale related variables at the moment. To no avail yet.

harding:
@bitschmidty do you have the `iconv` command available on macos?  If so, you can replace the `cat` command in the test with `iconv -f UTF-8`, which should convert any UTF-8 document to your current locale.

(I can't easily replicate your problem here---I'm using GNU sed, which seems to ignore most locale problems, but you're presumably using BSD/Darwin sed, which is apparently more picky about its input.

bitschmidty:
@harding I do have `iconv` here. However replacing cat with `iconv -f UTF-8` still yields the slew of `sed: RE error: illegal byte sequence` messages

Removing the last `g` from `	      | sed 's/\>/\n/g; s/ *//g;' \` resolves the error messages, but I am not exactly sure why. But perhaps a clue?

harding:
@bitschmidty 

> Removing the last g from | sed 's/\>/\n/g; s/ *//g;' \ resolves the error messages, but I am not exactly sure why. But perhaps a clue?

That's good detective work.  Some suggestions:

Can you edit the Makefile on your system and put `echo $$file ; \` on the line above `cat $$file \`?  That should print each file name before it's processed and the error is generated, allowing you to see which files have issues.

Once you identified a file, you can try narrowing down which line has problems.  Using a normal command prompt (not via the Makefile), try running the problem command on the file by itself, e.g. if the file is `foo.md`, try: `sed 's/ *//g;' foo.md`.  If that prints the same error as before, then you can try checking each individual line.  For example, `file=foo.md ; for i in $( seq $( cat $file | wc -l ) ) ; do echo $i ; head -n$i $file | sed 's/ *//g' > /dev/null || break ; done`

When you know what line is having problems, you can look at it in your text editor to see if you can find the problem.

bitschmidty:
ACK f42afa5a374dab49fb7dd0148723364ef033a19f

@harding latest checks work for me! Thanks for troubleshooting with me

harding:
@bitschmidty thanks for the careful review and for testing over and over and over again!

Makefile
jnewbery: @harding do you think this is worth updating?

jnewbery: ha. You just did. Thanks!

jnewbery:
Thanks for persisting with this @bitschmidty and @harding !

PR #443, opened 2020-07-28
Mention c-lightning adding multipath payments

laanwj:
Add a link to item mentioning the c-lightning PR to the list of mentions.

### DIFF
--- a/_topics/en/multipath-payments.md
+++ b/_topics/en/multipath-payments.md
@@ -58,6 +58,10 @@ primary_sources:
 ## Optional.  Each entry requires "title", "url", and "date".  May also use "feature:
 ## true" to bold entry
 optech_mentions:
+  - title: "C-Lightning #3809 adds support for sending of multipath payments"
+    url: /en/newsletters/2020/07/22/#c-lightning-3809
+    date: 2020-07-22
+
   - title: "LN protocol 1.1 goals: multipath payments"
     url: /en/newsletters/2018/11/20/#multi-path-payments
     date: 2018-11-20


_topics/en/multipath-payments.md
```diff
@@ -58,6 +58,10 @@ primary_sources:
 ## Optional.  Each entry requires "title", "url", and "date".  May also use "feature:
 ## true" to bold entry
 optech_mentions:
+  - title: "C-Lightning #3809 adds support to C-Lightning for effective sending of multipath payments"
```
jonatack, line 61: nits:

- maybe omit "to C-Lightning ", per the format of other entries

- unsure if "effective" if needed, though it is present in the first sentence of the linked-to news

jonatack: Does the entry need to be at the bottom of the list (in the code, the display is sorted)?

harding:
@laanwj Thanks for the PR!  We usually update the topic index on the first of each new month with links from the previous month, so the news from last week's newsletter would be added to the index in a few days, e.g. last month's update: https://github.com/bitcoinops/bitcoinops.github.io/pull/428  However, I have no objection to adding stuff early if people really want it.

FYI, this is failing because the title field is limited to 75 characters: https://travis-ci.org/github/bitcoinops/bitcoinops.github.io/builds/712578266#L673  I think you could just remove "to C-Lightning" from the title to get it to pass.

_topics/en/multipath-payments.md
laanwj: Thanks, updated for both

laanwj:
OK, thanks, Seems fine to do it in one go later then. I don't know the process, just randomly noticed that this was missing.

jonatack:
re-tested ACK

_topics/en/multipath-payments.md
harding: @jonatack FYI: entries can be anywhere in the YAML; the display logic sorts by date.

harding:
Untested ACK.  Thanks!

jnewbery:
Thanks  @laanwj ! You've just joined a very elite team of optech website contributors :astronaut:

PR #447, opened 2020-08-03
Style: use lowercase names for P2P message types

jnewbery:


jnewbery:
@harding

harding:
ACK.  Thanks!

PR #451, opened 2020-08-11
News109: signets use proof of work

jnewbery:
Thanks to @pinheadmz for reporting!

### DIFF
--- a/_posts/en/newsletters/2020-08-05-newsletter.md
+++ b/_posts/en/newsletters/2020-08-05-newsletter.md
@@ -241,8 +241,8 @@ release candidates.*
 
 - [BIPs #947][] updates the [BIP325][] specification of [signet][topic
   signet] to change the block signature verification method.  Signets
-  are test networks where valid blocks are signed by trusted signers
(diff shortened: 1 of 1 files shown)


harding:
Thanks @pinheadmz for reporting and @jnewbery for editing!

I agree the original is incorrect, but I think this edit could accidentally mislead people into thinking PoW plays a significant role in signets (it can, if you want, but it doesn't need to).  I think we can just drop the: "rather than using proof of work" so the text reads, "Signets are test networks where valid blocks are signed by trusted signers, [...]".  I think this is even more accurate as, in addition to valid blocks still needing a correct amount of PoW, they also need to follow all the other consensus rules---in Bitcoin, sufficient PoW doesn't eliminate the need to follow all the other rules, and in signet, a valid signature doesn't either.

jnewbery:
> I think we can just drop the: "rather than using proof of work"

I agree. Thanks @harding!

pinheadmz:
LGTM :+1:

bitschmidty:
ACK

harding:
ACK.  Thanks!

PR #452, opened 2020-08-14
topics: take date from URL for newsletter links

jnewbery:
Less typing for Harding, less checking for reviewers, and fewer typos.

This caught 5 inconsistencies between the date field and the newsletter date of publication:

```
diff -r _site.old/en/topics/hwi/index.html _site/en/topics/hwi/index.html
94c94
<   <li>2019-12-18 <a href="/en/newsletters/2019/12/28/#core-hwi">2019 year-in-review: HWI</a></li>
---
>   <li>2019-12-28 <a href="/en/newsletters/2019/12/28/#core-hwi">2019 year-in-review: HWI</a></li>
diff -r _site.old/en/topics/multipath-payments/index.html _site/en/topics/multipath-payments/index.html
140c140
<   <li>2019-09-15 <a href="/en/newsletters/2019/09/11/#lnd-3390">LND #3390 separates tracking of HTLCs from invoices as necessary for AMP</a></li>
---
>   <li>2019-09-11 <a href="/en/newsletters/2019/09/11/#lnd-3390">LND #3390 separates tracking of HTLCs from invoices as necessary for AMP</a></li>
diff -r _site.old/en/topics/musig/index.html _site/en/topics/musig/index.html
100c100
<   <li>2019-12-30 <a href="/en/newsletters/2019/12/04/#composable-musig">Composable MuSig—concerns about safely using signer sub-groups</a></li>
---
>   <li>2019-12-04 <a href="/en/newsletters/2019/12/04/#composable-musig">Composable MuSig—concerns about safely using signer sub-groups</a></li>
diff -r _site.old/en/topics/op_checktemplateverify/index.html _site/en/topics/op_checktemplateverify/index.html
113c113
<   <li>2020-02-19 <a href="/en/newsletters/2020/04/22/#vaults-prototype">Vault prototype with sample implementation using OP_CHECKTEMPLATEVERIFY</a></li>
---
>   <li>2020-04-22 <a href="/en/newsletters/2020/04/22/#vaults-prototype">Vault prototype with sample implementation using OP_CHECKTEMPLATEVERIFY</a></li>
diff -r _site.old/en/topics/taproot/index.html _site/en/topics/taproot/index.html
101c101
<   <li>2020-05-28 <a href="/en/newsletters/2020/05/20/#evaluate-proposed-changes-to-bip341-taproot-transaction-digest">Request for comments on amending BIP341 taproot transaction digest</a></li>
---
>   <li>2020-05-20 <a href="/en/newsletters/2020/05/20/#evaluate-proposed-changes-to-bip341-taproot-transaction-digest">Request for comments on amending BIP341 taproot transaction digest</a></li>
```

The date field can be removed from all newsletter references after this PR (or in the same PR if desired)

We can also swap the order of the conditional if we want the option of overriding the newsletter publication date.

### DIFF
--- a/_layouts/topic.html
+++ b/_layouts/topic.html
@@ -52,19 +52,14 @@ layout: post
 
   <!-- Build list of internal optech mentions -->
   {% assign references = '' %}
-  {% assign mentions = page.optech_mentions | sort: "date" | reverse %}
-  {% for mention in mentions %}
+  {% for mention in page.optech_mentions %}
     {% if mention.feature == true %}
-      {% assign bold = site.bold %}
+      {% assign bold = '{:.bold}' %}
     {% else %}
       {% assign bold='' %}
     {% endif %}
-    {% if mention.url contains "/en/newsletters" %}
-      {% assign date = mention.url | remove_first: "/en/newsletters/" | slice: 0, 10 | replace: "/", "-" %}
-    {% else %}
-      {% assign date = mention.date %}
-    {% endif %}
-    {% capture references %}{{references}}{{newline}}- {{bold}}{{date}} [{{mention.title}}]({{mention.url}}){{bold}}{% endcapture %}
+    {% include functions/get-mention-date.md %}
+    {% capture references %}{{references}}{{newline}}- {{date}} [{{mention.title}}]({{mention.url}}){{bold}}ENDENTRY{% endcapture %}
   {% endfor %}
 
   <!-- Build list of see also entries -->
@@ -99,7 +94,7 @@ layout: post
   {%- if page.optech_mentions and page.optech_mentions != '' -%}
     ## Optech newsletter and website mentions
 
-    {{references}}
+    {{ references | split: 'ENDENTRY' | sort | reverse}}
   {% endif %}{{newline}}{{newline}}
 
   {%- if page.see_also and page.see_also != '' -%}
--- a/_layouts/topic.html
+++ b/_layouts/topic.html
@@ -59,7 +59,12 @@ layout: post
     {% else %}
       {% assign bold='' %}
     {% endif %}
-    {% capture references %}{{references}}{{newline}}- {{bold}}{{mention.date}} [{{mention.title}}]({{mention.url}}){{bold}}{% endcapture %}
+    {% if mention.url contains "/en/newsletters" %}
+      {% assign date = mention.url | remove_first: "/en/newsletters/" | slice: 0, 10 | replace: "/", "-" %}
+    {% else %}
+      {% assign date = mention.date %}
+    {% endif %}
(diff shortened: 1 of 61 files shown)


harding:
@jnewbery this is wonderful, thanks!  When we discussed this last year, I was only thinking of a perfect solution that would apply to all internal references, including non-newsletter references, but your solution here is very nicely pareto optimized for the overwhelming number of our references that comes from newsletters.  Nice!

> The date field can be removed from all newsletter references after this PR (or in the same PR if desired)

I think for entries without dates to pass tests, you will need to delete this line: https://github.com/bitcoinops/bitcoinops.github.io/blob/master/_data/schemas/topics.yaml#L87 (that should be the only change necessary)

> We can also swap the order of the conditional if we want the option of overriding the newsletter publication date.

That's good flexibility to have, but I can't see any use for it at present.  Maybe mention that in a comment so dumb people like don't forget it.  E.g.:

```html
{% comment %}<!-- Can swap the order of the conditional if we want the option of overriding the newsletter publication date -->{% endcomment %}
```

_layouts/topic.html
```diff
@@ -59,7 +59,12 @@
     {% else %}
       {% assign bold='' %}
     {% endif %}
-    {% capture references %}{{references}}{{newline}}- {{bold}}{{mention.date}} [{{mention.title}}]({{mention.url}}){{bold}}{% endcapture %}
+    {% if mention.url contains "/en/newsletters" %}
+      {% assign date = mention.url | remove_first: "/en/newsletters/" | slice: 0, 10 | replace: "/", "-" %}
+    {% else %}
+      {% assign date = mention.date %}
```
harding, line 65: If we make `date:` non-mandatory in the YAML, we need to catch the case here of entries that aren't newsletters but also don't have dates defined.  E.g. (untested):

```
[...]
{% else %}
  {% if mention.date == empty %}
    {% include ERROR_44_MISSING_DATE %}
  {% else %}
    {% assign date = mention.date %}
  {% endif %}
{% endif %}
[...]
```

jnewbery:
I tried removing the data field from the newsletter entries and it breaks the ordering, since they're sorted based on that date field before we do anything else.

I think the easiest thing to do here is simply add an assertion that the date field matches the newsletter URL, unless @harding can think of some liquid magic.

harding:
Sorting is easy.  I just realized that if we remove dates from the YAML, we'll also need to update the https://bitcoinops.org/en/topic-dates/ page.  Let me take a look at get back to you either with a patch or to just do an assert like you suggest.

jnewbery:
Yeah, I realised the topic-dates thing too. Just pushed a change for that.

harding:
Opened https://github.com/jnewbery/bitcoinops.github.io/pull/1

jnewbery:
Thanks @harding. I took your commit and added a third commit that removes the now unnecessary date fields for all newsletter links.

PR #460, opened 2020-09-21
Topics: update with mentions from August 2020

harding:
&nbsp;

### DIFF
--- a/_topics/en/psbt.md
+++ b/_topics/en/psbt.md
@@ -117,6 +117,15 @@ optech_mentions:
   - title: "LND #4455 makes it safe to batch open channels using PSBTs"
     url: /en/newsletters/2020/07/29/#lnd-4455
 
+  - title: BIP174 specification of PSBT updated in response to fee overpayment attack
+    url: /en/newsletters/2020/08/05/#bips-948
+
+  - title: "Bitcoin Core #18654 adds new RPC specifically for RBF fee bumping PSBTs"
+    url: /en/newsletters/2020/08/19/#bitcoin-core-18654
+
+  - title: "BIPs #955 updates BIP174 PSBT to standardize supplying hash preimages"
+    url: /en/newsletters/2020/08/26/#bips-955
+
 ## Optional.  Same format as "primary_sources" above
 see_also:
   - title: Output Script Descriptors
--- a/_topics/en/schnorr-signatures.md
+++ b/_topics/en/schnorr-signatures.md
@@ -95,6 +95,9 @@ optech_mentions:
   - title: Presentations and discussions about schnorr signatures
     url: /en/newsletters/2020/07/01/#schnorr-signatures-and-multisignatures
 
+  - title: Proposal to update BIP340 schnorr signatures to use evenness tiebreaker
+    url: /en/newsletters/2020/08/19/#proposed-uniform-tiebreaker-in-schnorr-signatures
+
 ## Optional.  Same format as "primary_sources" above
 see_also:
   - title: Will a schnorr soft fork introduce a new address format?
--- a/_topics/en/addr-v2.md
+++ b/_topics/en/addr-v2.md
@@ -48,6 +48,9 @@ optech_mentions:
   - title: Proposed per-node signaling for address relay with v2 `addr` messages
     url: /en/newsletters/2019/11/06/#signaling-support-for-address-relay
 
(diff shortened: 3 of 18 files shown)


_topics/en/psbt.md
```diff
@@ -117,6 +117,15 @@ optech_mentions:
   - title: "LND #4455 makes it safe to batch open channels using PSBTs"
     url: /en/newsletters/2020/07/29/#lnd-4455
 
+  - title: BIP176 specification of PSBT updated in response to fee overpayment attack
```
bitschmidty, line 120: s/BIP176/BIP174

bitschmidty (changes requested):
Tested links. 

Just one text item caught.

_topics/en/schnorr-signatures.md
```diff
@@ -95,6 +95,9 @@ optech_mentions:
   - title: Presentations and discussions about schnorr signatures
     url: /en/newsletters/2020/07/01/#schnorr-signatures-and-multisignatures
 
+  - title: Proposal to update BIP340 schnorr signatures to use eveness tiebreaker
```
jnewbery, line 98: s/eveness/evenness/

jnewbery (commented):
One minor typo inline.

I've always found that reviewing these PRs involves more overhead than I'd like (opening a diff, opening a browser, figuring out which topic page to open, scanning through that page for new links). This time, I used the following method, which seemed more efficient:

- Open a terminal and browser side by side
- Run `git diff HEAD~ | sed "s'url: /'url: https://bitcoinops.org/'" | colordiff | less -r` in the terminal (converts internal links to absolute links and recolorizes the output).
- Click on the new links in the less output to open the link in the browser.

@bitschmidty - I don't know what you're doing to review these, but perhaps the method above will save you a bit of time too?

@harding - I don't know how you're collecting these. I wonder if adding tags to the newsletter files would make them easier to collect/verify that you haven't missed any?

harding:
@jnewbery 

> @harding I don't know how you're collecting these. I wonder if adding tags to the newsletter files would make them easier to collect/verify that you haven't missed any?

I collect them by just skimming the newsletters from the previous month and editing the topic files as appropriate.  We already have a partial tagging system in place: when the newsletter mentions a topic, it usually links to the corresponding topic page, so the blue links are an obvious signal.  I don't currently do any verification.

A test we could add (perhaps independent from CI) is a backlink check: everytime the newsletter (or maybe other pages) links to a topic page, that topic page has to link back to the newsletter/page.  This would be fallible when we discuss a topic more than once in a single newsletter, and it would need the ability to override backlinks sometimes (e.g. when a topic is only mentioned incidentally in a news item), but I think such a test would be possible.

Either with or without that test, something to consider would be moving to more frequent topic updates.  We currently do monthly updates (when I remember!) but we could also bundle the updates with newsletters.  If we made the updates in the same PR as added the newsletter, then reviewers would be more likely to notice when we'd forgotten to update the topic listing for content in that newsletter; a downside would possibly be a bit more churn on the newsletter diff.  Alternatively, if we made the updates for newsletter n the following week in newsletter n+1, churn would be minimal and reviewers might still have a good enough recollection of the previous week's content to warn about missing links.  The advantages to both approaches would be fewer changes per update to review and the topic index would remain always up-to-date.

I slightly prefer combining topic updates with the newsletter PRs in order to keep everything up to date, but I don't want to place too much burden on reviewers, so please say something if that sounds annoying.

jnewbery:
> A test we could add ... This would be fallible

Sounds to me like too much machinery/overhead to be worth it. @harding - if you think it'd be helpful for you and want to implement it, I'm happy to review, but it doesn't sound worth it to me.

> something to consider would be moving to more frequent topic updates

ACK from me. Sounds like this would be more efficient for you, and I review both the newsletters and topics updates anyway, so it's a win for me. Newsletter reviewers who don't want to review the new topics links shouldn't feel obligated to - they can just review the newsletter as normal.

bitschmidty:
> I don't know what you're doing to review these

I currently build locally and then review text and check new links in browser for each file/topic changed.

> I slightly prefer combining topic updates with the newsletter PRs in order to keep everything up to date

Im happy to help with this review (and adding links when authoring) each week.

harding:
@jnewbery @bitschmidty re: you both being good with topic updates in newsletters: awesome, thanks!  Next Wednesday after Mike merges `#117`, I'll open a topic update for September; then I'll include per-newsletter topic updates starting with `#118`.

PR #462, opened 2020-09-24
Topics: add HTLCs and PTLCs

harding:
(We normally do one topic at a time now, but these were related so I did them together).

### DIFF
--- a/_topics/en/htlc.md
+++ b/_topics/en/htlc.md
@@ -1,0 +1,134 @@
+---
+title: Hash Time Locked Contract (HTLC)
+
+## Optional.  Shorter name to use for reference style links e.g., "foo"
+## will allow using the link [topic foo][].  Not case sensitive
+shortname: htlc
+
+## Optional.  An entry will be added to the topics index for each alias
+#aliases:
+#  - Foo
+
+## Required.  At least one category to which this topic belongs.  See
+## schema for options
+categories:
+  - Contract Protocols
+  - Lightning Network
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+## Should be less than 500 characters
+excerpt: >
+  **Hash Time Locked Contracts (HTLCs)** are conditional payments used
+  in LN payment channels, cross-chain atomic swaps, same-chain
+  coinswaps, zero-knowledge contingent payments, and other contract
+  protocols.
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  HTLCs have two fundamental clauses: a payment clause secured with a
+  *hash lock* and a refund clause secured with a *time lock*.  To open a
+  hash lock and claim a payment, the receiver needs to reveal the
+  preimage to a hash digest encoded in the contract.  To open a time
+  lock and receive a refund, the spender needs to wait until after a
+  certain time encoded in the contract.
+
+  Because revealed preimages and past times don't uniquely identify the
+  person who should receive either the payment, HTLCs are only secure if
+  they also require a unique signature matching the public key of either
+  the spender (refundee) or the intended receiver.  That makes the
+  minimal HTLC look something [like this][htlc in minsc] in the
+  [Minsc][] scripting language:
+
+  ```hack
+  (pk(Receiver) && sha256(H)) || (pk(Refundee) && older(10))
+  ```
+
+  ### History
+
+  Hash locks and time locks as independent features were clearly
+  designed into the original version of Bitcoin.  As far as we're aware,
+  the earliest description combining the two to create a conditional
+  payment---what'd we now call an HTLC---is a set of posts ([1][maxwell
+  hashlock], [2][maxwell timelock]) from July 2012. However, it's
+  possible that a [December 2010 post][nakamoto risk free trade] was
+  alluding to the same basic idea when mentioning a "cryptographically
+  [...] risk free trade".
+
+  ### Future
+
+  [Point Time Lock Contracts][topic ptlc] (PTLCs) perform the same
+  function as HTLCs but can provide better privacy, use less block
+  space, and prevent routing interception.  As a downside, they depend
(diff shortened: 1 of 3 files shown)


_topics/en/htlc.md
```diff
@@ -0,0 +1,133 @@
...
+  [...] risk free trade".
+
+  ### Future
+
+  [Point Time Lock Contracts][topic ptlc] (PTLCs) perform the same
+  function as HTLCs but can provide better privacy, use less block
+  space, and prevent routing interception.  As a downside, they depend
+  on [signature adaptors][topic adaptor signatures] which can only
+  implemented using Bitcoin's existing ECDSA signatures by making
+  additional security assumptions or by using an `OP_CHECKMULTISIG`
```
jnewbery, line 64: I believe it's possible to implement signature adaptors in ECDSA _without_ additional security assumptions (e.g. by using https://eprint.iacr.org/2019/503.pdf), but I'm certainly not an expert. Perhaps rewording this to "by using complex multi-party protocols" would be more accurate.

I see that the adaptor signatures topic page also states that additional security assumptions are required.

_topics/en/ptlc.md
```diff
@@ -0,0 +1,93 @@
...
+  unlocking method:
+
+  - **HTLC hash locks:** are locked using a hash digest and unlocked by
+    providing the corresponding preimage.  The most commonly used hash function is
+    SHA256, which produces a 256-bit (32-byte) digest commonly generated
+    from a 32-byte preimage.
+
+      When used to secure multiple payments (e.g. a routed LN payment or
+      an atomic swap), all payments use the same preimage and hash lock.
+      This creates a link between those payments if they are published
```
jnewbery, line 40: s/they are/they're/

jnewbery (commented):
Just one small style suggestion and one question about _additional security assumptions_.

jnewbery:
Perhaps @LLfourn or @real-or-random would be kind enough to confirm whether https://github.com/bitcoinops/bitcoinops.github.io/pull/462#discussion_r494900113 is correct.

_topics/en/htlc.md
harding: Interesting!  Instead of dropping "additional security assumptions", we could possibly add an additional problem like "slow".  E.g. the paper you cites shows its protocol and the protocol it's comparing to requiring several seconds for key generation on a 256-bit curve with 128 bits security:



(And that's on a 3.6 GHz x86_64 machine (single core); if multisig keys need to be generated on a hardware wallet, that seems essentially impractical with current Trezor, Ledger, or ColdCard devices.)

LLFourn: > I believe it's possible to implement signature adaptors in ECDSA _without_ additional security assumptions (e.g. by using https://eprint.iacr.org/2019/503.pdf), but I'm certainly not an expert.

This paper *does* involve very bespoke assumptions about *class groups of imaginary quadratic fields* (which are among the wildest assumptions you can make). However, there are papers that do multi-party ECDSA without extra assumptions over ECDSA. See https://eprint.iacr.org/2018/499.pdf.

The question is: Does (non-multiparty) ECDSA adaptor signatures require additional assumptions over ECDSA itself. My research says yes because the scheme has a bug: each ECDSA adaptor signature leaks a Diffie-Hellman key between the signing and encryption key (referred to here as `T`). In https://eprint.iacr.org/2020/476.pdf they fix this by making the receiver always provide a proof of knowledge for `T` i.e. they could have produced the Diffie-Hellman key without the adaptor signature since they know `t` so it doesn't matter. If you do this then it requires no new assumptions.

Now in practice, this is all moot because by using single-signer adaptor ECDSA adaptor signatures you are using a script 2-of-2 and therefore p2wsh which means you are relying on collision resistance of HASH160 which arguably the worst assumption of all.

Having said all that, I think the current paragraph is good as it is. It feels like working in this information about the bug in ECDSA adaptor signatures is too much for the intended audience. so LGTM!

jnewbery: Thanks Lloyd! Really appreciate the explanation.

harding: @llfourn

Awesome explanation, thanks!  One note:

> using single-signer adaptor ECDSA adaptor signatures you are using a script 2-of-2 and therefore p2wsh which means you are relying on collision resistance of HASH160

P2WSH uses SHA256.  See [BIP141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#Witness_program), "If the version byte is 0, and the witness program is 32 bytes:  [...] SHA256 of the witnessScript must match the 32-byte witness program."

_topics/en/ptlc.md
```diff
@@ -0,0 +1,93 @@
...
+    64 bytes.  However, using either multiparty ECDSA or schnorr key
+    aggregation and signing, the keys and signature can be combined
+    with other keys and signatures needed to authorize any spend,
+    allowing point locks to use zero bytes of distinct block space.
+
+      Each point lock can use different keys and signatures, so there is
+      nothing about the point lock that correlates different payments
+      either onchain or when routed offchain through surveillance nodes.
+
+  Implementation of PTLCs in Bitcoin requires creating [signature
```
bitschmidty, line 57: We link to adaptor signatures and schnorr sigs back to back just above this, so might be redundant linking.

bitschmidty (commented):
tACK acb09d4d624bafade41baa41f2db95058e74b7e6

Looking good. Just one small suggestion.

_topics/en/ptlc.md
jnewbery: Seems ok to me. I don't have a strong opinion either way.

jnewbery:
Merging this now. We can remove the extra links (https://github.com/bitcoinops/bitcoinops.github.io/pull/462#discussion_r496171928) later if others agree they should go.

_topics/en/htlc.md
LLFourn: Thanks for the correction @harding! That's good news!

PR #459, opened 2020-09-20
Newsletters: add 116 (2020-09-23)

harding:
- [x] Update releases/RCs on Tuesday @harding 
- [x] Clients and services @bitschmidty 
- [x] `Bitcoin Core #19643` add -netinfo peer connections dashboard @dongcarl

PR #463, opened 2020-09-24
References: deprecated some old quick links

harding:
In last week's newsletter, @bitschmidty [caught](https://github.com/bitcoinops/bitcoinops.github.io/pull/459#discussion_r492166331) me accidentally linking directly to Wuille's miniscript page rather than the Optech topic page I wanted.  This happened because, prior to the addition of the topics page, I had defined some quick reference links for highly-referenced subjects; obviously, many of those subjects later became topics.

This PR deprecates those old links.  On pages with dates in the past, the old link definitions will be used.  On pages with newer dates (or no dates at all, like the topic pages), the old link definitions aren't evaluated and so the hyperlink won't be generated, causing the post-build test to fail---hopefully ensuring I insert the appropriate topic link (or whatever).

Except for two places where we revise a link on a topics page to point at another topics page (rather than a primary source), this PR should not result in any changes to the compiled site.

### DIFF
--- a/_includes/references.md
+++ b/_includes/references.md
@@ -28,24 +28,31 @@
 [bolts repo]: https://github.com/lightningnetwork/lightning-rfc/
 [c-lightning]: https://github.com/ElementsProject/lightning
 [c-lightning repo]: https://github.com/ElementsProject/lightning
-[cve-2012-2459]: https://bitcointalk.org/?topic=102395
-[cve-2017-12842]: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-12842
-[cve-2018-17144]: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-17144
-[descriptor]: https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md
 [eclair repo]: https://github.com/ACINQ/eclair
-[eltoo]: https://blockstream.com/eltoo.pdf
-[erlay]: https://arxiv.org/pdf/1905.10518.pdf
-[hwi]: https://github.com/bitcoin-core/HWI
+[hwi repo]: https://github.com/bitcoin-core/HWI
 [libminisketch]: https://github.com/sipa/minisketch
 [libsecp256k1]: https://github.com/bitcoin-core/secp256k1
 [libsecp256k1 repo]: https://github.com/bitcoin-core/secp256k1
 [lnd repo]: https://github.com/lightningnetwork/lnd/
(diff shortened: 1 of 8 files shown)


_includes/references.md
```diff
@@ -28,24 +28,31 @@
...
+[hwi repo]: https://github.com/bitcoin-core/HWI
+[libminisketch]: https://github.com/sipa/minisketch
+[libsecp256k1]: https://github.com/bitcoin-core/secp256k1
+[libsecp256k1 repo]: https://github.com/bitcoin-core/secp256k1
+[lnd repo]: https://github.com/lightningnetwork/lnd/
+[rust-lightning repo]: https://github.com/rust-bitcoin/rust-lightning
+
+{% comment %}<!-- deprecated links; don't use these any more -->{% endcomment %}
+{% assign pagedate_epoch = page.date | date: '%s' %}
+{% assign deprecated_links_v0_epoch = '2020-09-24' | date: '%s' %}
```
jnewbery, line 41: What's the plan if we want to deprecate more links in future. Does this date need to be updated?

jnewbery (commented):
Looks ok.

Would it makes sense to follow this up by just removing the deprecated links from references.md and moving them to the newsletters that need them where necessary?

_includes/references.md
harding: My idea was that, if we want to deprecate links en masse in the future, we'd create a v1 deprecation by copying this template.  E.g. 

    {% assign pagedate_epoch = page.date | date: '%s' %}
    {% assign deprecated_links_v0_epoch = '2020-09-24' | date: '%s' %}
    {% if pagedate_epoch < deprecated_links_v0_epoch %}
    [... links deprecated in this PR ...]
    {% endif %}

    {% assign pagedate_epoch = page.date | date: '%s' %}
    {% assign deprecated_links_v2_epoch = '2022-01-01' | date: '%s' %}
    {% if pagedate_epoch < deprecated_links_v0_epoch %}
    [... links deprecated in future PR ...]
    {% endif %}

harding:
@jnewbery 

> Would it makes sense to follow this up by just removing the deprecated links from references.md and moving them to the newsletters that need them where necessary?

I initially tried that, but it looked like it was going to be a very large edit (and it wasn't easy to do with sed due to multi-word link references sometimes spanning across newlines).  However, I have no objection to it---it should be easy to verify a correct change of that type by diffing the built site before and after the edit, confirming that there are no link changes.

jnewbery:
ACK. Confirmed that the only changes in the generated site are the links to musig in the schnorr signatures and taproot pages.

PR #464, opened 2020-09-24
Topics: bucket mentions by year

harding:
We're starting to get some topics with a lot of mentions, which produces a long hard-to-read list.  This PR buckets the entries by year and drops the per-entry date (which is now less important and which was always a bit ugly).  Example:

### DIFF
--- a/_layouts/topic.html
+++ b/_layouts/topic.html
@@ -59,7 +59,7 @@ layout: post
       {% assign bold='' %}
     {% endif %}
     {% include functions/get-mention-date.md %}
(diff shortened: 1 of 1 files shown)


jnewbery:
ACK. I agree that this looks a lot better.

We're now only ever using the year and month of the mention dates. Perhaps we should:

- modify get-mention-date.md to only return YYYY-MM
- not require the day in explicit mention dates

If you agree, I can make a follow-up PR that does that.

harding:
I think we still want days, otherwise lists may sort in strange orders, e.g.:

- PR 1234 adds nifty new feature
- Idea discussed for nifty new feature
- Foo released with nifty new feature

bitschmidty:
Post-merge ACK, thanks @harding

PR #465, opened 2020-09-25
Add local site diff'ing instructions

jnewbery:
Useful for reviewing changes to site-generating code.

Not sure if it's worth packaging this up into a script?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,6 +54,21 @@ translation PRs are opened within a week of the original newsletter being
 published for new newsletters. That said, we also encourage translation of
 older newsletters and blog posts as well.
 
+### Reviewing changes to the website generation code
+
+When reviewing changes to the Jekyll code that generates the website, it's useful to compare
+the generated website before and after the change. To do so, you can run the following commands:
+
+```bash
+git checkout $reviewbranch  # Checkout the branch that you want to review
+rm -rf _site.bak  # Remove the backup site directory if it exists
+JEKYLL_ENV=local make build  # Do a clean build of the site to the _site directory. The local env is to remove the 'updated' field which causes irrelevant diffs
+mv _site _site.bak  # Rename the generated site directory
+git checkout master  # Checkout the base branch
+JEKYLL_ENV=local make build  # Do a clean build of the site to the _site directory
+diff -ruN _site _site.bak  # Compare the generated sites (-r for recursive, -u for unified, -N for new file)
+```
+
 ## Compatibility Matrix Data
 
 The compatibility matrix section of the website is built from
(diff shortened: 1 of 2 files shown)


jonatack:
Interesting! For the newsletters, which are mostly content addition, I haven't felt the need for diffing but I'll try this on Monday to see if it's a good complement to reviewing the newsletter in localhost.

jnewbery:
> For the newsletters, which are mostly content addition, I haven't felt the need for diffing but I'll try this on Monday to see if it's a good complement to reviewing the newsletter in localhost.

It's probably not going to be useful for reviewing newsletters, since generally the change there is just an additional page. The intended purpose for this is reviewing changes like #463, where the code change is supposed to make either no difference or specific differences to the generated site.

jonatack:
Makes sense. I'll try these instructions on next review.

CONTRIBUTING.md
```diff
@@ -54,6 +54,21 @@ translation PRs are opened within a week of the original newsletter being
...
+the generated website before and after the change. To do so, you can run the following commands:
+
+```bash
+git checkout $reviewbranch  # Checkout the branch that you want to review
+rm -rf _site.bak  # Remove the backup site directory if it exists
+JEKYLL_ENV=local make build  # Do a clean build of the site to the _site directory. The local env is to remove the 'updated' field which causes irrelevant diffs
+mv _site _site.bak  # Rename the generated site directory
+git checkout master  # Checkout the base branch
+JEKYLL_ENV=local make build  # Do a clean build of the site to the _site directory
+diff _site _site.bak  # Compare the generated sites
```
harding, line 69: My diff (GNU diffutils) doesn't automatically diff subdirs.  E.g.:

    $ cd /tmp
    $ mkdir -p a/b c/b
    $ echo foo > a/b/foo.txt
    $ diff a c
    Common subdirectories: a/b and c/b

It also doesn't automatically display the diffs of files that only exist in one version (it just prints a line about "Only in c: foo.txt", which is easy to miss in a big diff).

For those reasons, I typically use this diff command: `diff --recursive --newfile` (or `diff -rN`).  (Personally, I also use `-u` for unified diffs, which makes the nice mnemonic `-ruN`)

harding:
Excluding [this concern](https://github.com/bitcoinops/bitcoinops.github.io/pull/465#discussion_r494920885), tested and works for me.  Great idea commenting out the variable date field!

@jnewbery 

> Not sure if it's worth packaging this up into a script?

I don't have any objection to this, but I think there might be a fair number of cases where people want to diverge from the instructions.  E.g., I typically use a unified diff and often I want to diff to the merge base rather than current master.  That said, I have no objection to a script to automates the steps for other people.

CONTRIBUTING.md
jnewbery: Much better. I usually use meld locally for running diffs and I wanted to put `diff` here to make the instructions universal. The options you suggest are good suggestions.

PR #469, opened 2020-10-01
Updated Installing doc.

MAYANK25402:


jnewbery:
I don't think this is the right change. All these verbs are in the indicate case since they're instructions (install this, do this, read this, preview this, ...)

unnown:
Please mark as invalid and spam, this is an obvious attempt to get a free t-shirt by 'contributing'(hacktoberfest)
This will also ban the user for trying to abuse the system by doing marginal effort

PR #363, opened 2020-03-05
Newsletter-82:Translate into Japanese

Ichiro0219:

PR #472, opened 2020-10-05
Topics: update for mentions from September

harding:
See [this comment](https://github.com/bitcoinops/bitcoinops.github.io/pull/460#pullrequestreview-494504105) for suggestions about how to review this PR.

### DIFF
--- a/_topics/en/eclipse-attacks.md
+++ b/_topics/en/eclipse-attacks.md
@@ -82,6 +82,12 @@ optech_mentions:
   - title: "Bitcoin Core 0.20 released with asmap feature to improve eclipse resistance"
     url: /en/newsletters/2020/06/10/#bitcoin-core-0-20-0
 
+  - title: "Bitcoin Core #19670 reserves inbound slots for block-relay-only peers"
+    url: /en/newsletters/2020/09/09/#bitcoin-core-19670
+
+  - title: "CVE-2018-17145 InvDoS attack could've been used to eclipse nodes"
+    url: /en/newsletters/2020/09/16/#inventory-out-of-memory-denial-of-service-attack-invdos
+
 ## Optional.  Same format as "primary_sources" above
 # see_also:
 #   - title:
--- a/_topics/en/hwi.md
+++ b/_topics/en/hwi.md
@@ -49,6 +49,9 @@ optech_mentions:
   - title: Initial release of Lily Wallet supports HWI
     url: /en/newsletters/2020/07/22/#lily-wallet-initial-release
 
+  - title: "HWI #363 adds support for Bitbox02 hardware wallet"
+    url: /en/newsletters/2020/09/09/#hwi-363
+
 ## Optional.  Same format as "primary_sources" above
 see_also:
   - title: Partially-Signed Bitcoin Transactions (PSBTs)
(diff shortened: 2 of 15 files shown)


_topics/en/eclipse-attacks.md
```diff
@@ -82,6 +82,12 @@ optech_mentions:
   - title: "Bitcoin Core 0.20 released with asmap feature to improve eclipse resistance"
     url: /en/newsletters/2020/06/10/#bitcoin-core-0-20-0
 
+  - title: "Bitcoin Core #19670 reserves inbound slots for block-relay-only peers"
+    url: /en/newsletters/2020/09/09/#bitcoin-core-19670
+
+  - title: "CVE-2018-17145 InvDoS attack could've been used to eclipse nodes"
```
jnewbery, line 88: Is InvDoS common terminology? I don't see it on the mailing list post or in the linked paper.

_topics/en/hwi.md
```diff
@@ -49,6 +49,9 @@ optech_mentions:
   - title: Initial release of Lily Wallet supports HWI
     url: /en/newsletters/2020/07/22/#lily-wallet-initial-release
 
+  - title: HWI adds support for Bitbox02 hardware wallet
```
jnewbery, line 52: Is there a reason to not include the PR number here (we do for other projects)

_topics/en/schnorr-signatures.md
```diff
@@ -98,6 +98,12 @@ optech_mentions:
   - title: Proposal to update BIP340 schnorr signatures to use evenness tiebreaker
     url: /en/newsletters/2020/08/19/#proposed-uniform-tiebreaker-in-schnorr-signatures
 
+  - title: Updates BIP340 to consistently use evenness tiebreaker
```
jnewbery, line 101: Some of the links for BIPs PRs include the PR number and some don't. Should we aim for a consistent style?

jnewbery (commented):
Looks good. Just a couple of comments/questions.

bitschmidty (commented):
tACK 03d7ad1c138cb857b558c42abae8e82abe06ac91

Also used @jnewbery 's terminal command which work quite well for me!

Noticed the PR numbers missing as @jnewbery  pointed out. 

Additionally, minor/internal/nit item, some titles are quoted and others are not. There may be a reason for this.

_topics/en/eclipse-attacks.md
harding: It's not in the paper, but it's in its URL: https://invdos.net/paper/CVE-2018-17145.pdf .  And that's what's in the logo:

jnewbery: 🤦‍♂️

_topics/en/schnorr-signatures.md
harding: Agree that I should be consistent about including PR numbers, but I think we should allow exceptions (in particular, sometimes it can be a challenge to get the number and description to both fit within our max character allowance for titles).

harding:
@bitschmidty 

> some titles are quoted and others are not. There may be a reason for this.

The main reason is that `#` is a quote character in YAML (when it starts a line or follows a space), so if you create a title with:

    title: HWI #363 adds support for Bitbox02 hardware wallet

What you get rendered is:

> - HWI

I also quote when using backtics, although I just did a quick test and they only seem to have special meaning at the beginning of a string (which also applies to some other characters like `@` and `*`).

If it would ease testing, we could add a linter that requires all title strings be quoted.

PR #475, opened 2020-10-14
Update CONTRIBUTING.md with review tips

jnewbery:
Add tip about reviewing topic page updates.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,21 +9,25 @@ Tuesday UTC may not be incorporated due to time constraints.
 
 ## Topics pages
 
-New [topics pages](https://bitcoinops.org/en/topics) are added periodically. Additionally,
+New topics pages are added to https://bitcoinops.org/en/topics periodically. Additionally,
 new links are added to the existing topics pages every week in the same pull request
-as for that week's newsletter. These new topics links can be reviewed on the netlify preview, or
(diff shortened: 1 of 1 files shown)


harding:
This looks fine to me (sans linter failure), but if you want, I think you could make it compatible with both local previews and remote previews.  E.g., for a remote preview:

    BASEURL="https://deploy-preview-474--bitcoinops.netlify.app/" ; git diff master _topics/ | sed "s'url: /'url: ${BASEURL:-http://localhost:4000/}'" | colordiff | less -r

For a local preview, just set BASEURL to empty (or omit it if you've never used it in that shell before):

    BASEURL="" ; git diff master _topics/ | sed "s'url: /'url: ${BASEURL:-http://localhost:4000/}'" | colordiff | less -r

(`${variable_name:-default_value}` appears to be POSIX shell, so it should work even with some non-bash shells, but I haven't 100% confirmed that.)

jnewbery:
Thanks for the review, Dave! I've pushed a fix to the linter issue and updated the shell command.

bitschmidty (approved):
tACK b37d18b20c170fc82c1ff2fce932f9e466cb5476

- Modified a topic md file locally
- Ran local diff command and validated output
- Ran remote/PR command
- `make production` to ensure no build errors

(on macOS)

jnewbery:
Thanks @bitschmidty!

PR #477, opened 2020-10-16
Add additional guidance on adding new topics page

michaelfolkson:


michaelfolkson:
Linter failure. Will fix.

harding:
I don't have any objection to this, but I think we could instead just point to `_data/schemas/topics.yaml` as the definition of the topics format.  Alternatively the topic template `_contrib/new-topic` (called by `make new-topic`) could have its comments updated to include any important limits.

Personally, I don't remember the limits exactly and just let the linter tell me when I need to fix things.

michaelfolkson:
Ok makes sense, I will close this. Linter guidance is [here](https://github.com/bitcoinops/bitcoinops.github.io/pull/100).

I'm still in process of sorting this kind of stuff (build, linter etc) out personally.

michaelfolkson:
Actually excluding the linter guidance makes sense. But some sort of guidance on when new topics should be added makes sense. We don't want external contributors adding new Topics pages unless the topics have appeared in the newsletter multiple times (from my understanding on the purpose of the Topics pages).

jnewbery:
> some sort of guidance on when new topics should be added makes sense.

I don't think we're really soliciting external contributors to write new topic pages. If anything, I'd suggest something like "if you're interested in writing a new topic page, please contact the optech maintainers". I wouldn't want someone to go to the effort of writing a page only for us to reject it for some reason.

michaelfolkson:
Fair enough. Makes sense @jnewbery. Will update.

PR #487, opened 2020-11-26
news125: PR 20305 fixup

jonatack:
During the multiple edits of this writeup, I didn't notice that the unit of the new `fee_rate` was removed, without which it no longer makes sense in the writeup why `fee_rate` was added.

I also didn't notice that "option" was removed in favor of parameter only, which is somewhat incorrect, as `fee_rate` is an option, not a param, in several of the changed RPCs.

- re-add "denominated in sat/vB"

- re-add "option"


jonatack:
(The EOF newline was auto-added by my editor.)

jnewbery:
ACK

PR #488, opened 2020-11-29
Style: update currency abbreviations, including for feerates

harding:
The goal here was to standardize on "sat/vbyte" as our preferred unit for feerates, but this required rewriting the main item for currency units.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -127,12 +127,36 @@ spelled the same.
 
 ## Units
 
-- **Currency:** all amounts should be denominated in BTC.  E.g. "the
-  default minimum relay fee is 0.00000001 BTC/vbyte".  When writing out
-  the name of the currency, pluralize by appending an `s`.  E.g. "Alice
-  owns several bitcoins" not ~~"Alice owns several bitcoin"~~.
+- **Currency:** acceptable units and abbreviations are:
 
-- **Weight:** use vbytes rather than weight units, unless using weight
+    | Unit | Abbreviations | Value | &nbsp; |
+    |-|-|-|-|
+    | bitcoins | **BTC** | |
+    | millibitcoins | mBTC | 0.001 BTC | 1e-3 BTC |
+    | microbitcoins | µBTC, uBTC | 0.000001 BTC | 1e-6 BTC |
+    | satoshis | **sat** | 0.00000001 BTC | 1e-8 BTC |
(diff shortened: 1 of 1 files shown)


STYLE.md
```diff
@@ -127,12 +127,36 @@ spelled the same.
...
+
+    | Unit | Abbreviations | Value |
+    |-|-|-|
+    | bitcoins | **BTC** | |
+    | millibitcoins | mBTC | 0.001 BTC |
+    | microbitcoins | µBTC, uBTC | 0.000001 BTC |
+    | satoshis | **sat** | 0.00000001 BTC |
+    | nanobitcoins | nBTC | 0.000000001 BTC |
+    | millisatoshis | msat | 0.00000000001 BTC |
+    | picobitcoins | pBTC | 0.000000000001 BTC |
```
jonatack, line 140: perhaps add a fourth column in exponential/scientfic notation (I find it helpful for expressing and recalling these units and the conversions between them)

STYLE.md
```diff
@@ -127,12 +127,36 @@ spelled the same.
...
+    dollar/dollars (USD).  Similarly, abbreviations are never
+    pluralized.
+
+- **Feerates:** the preferred unit is "sat/vbyte" (satoshis per vbyte).
+  When used only sporadically within a page or section, it is preferred
+  to write out "sat/vbyte" each time, but when used frequently, you may
+  introduce and use the abbreviation "sat/vB".  See note in
+  *weight/vbytes* item about using SI prefixes.
+
+- **Weight/vbytes:** use vbytes rather than weight units, unless using weight
```
jonatack, line 157: perhaps introduce the abbreviation here: "weight units (WU), unless..."

jonatack (commented):
ACK, good timing--among other items today am just now revisiting the code documentation (and refactoring the code) in `policy/feerate.h`.

STYLE.md
harding: I think the abbreviation for weight units is unambiguous, so it probably doesn't need to be explicitly mentioned.

jonatack: No strong opinion, but I should have mentioned that my comment was motivated by someone recently asking on irc what "WU" meant.

jnewbery:
ACK. Looks good to me.

I agree with @jonatack about adding a column with scientific notation.

jonatack:
LGTM

PR #490, opened 2020-12-01
Topics: replace duplicate entry in PTLCs

harding:
It seems I accidentally added the same link to the PTLCs topic twice (it was that exciting!).  This replaces the redundancy with a different link and adds a test to hopefully stop me from doing it again.


bitschmidty (approved):
It was that exciting! That said, apologies for missing this in the review and publishing process. 

tACK on new entry as well as duplicate topic entry detection check

PR #493, opened 2020-12-17
News97: add label to illustration of attack

harding:
This adds a label to the image in https://bitcoinops.org/en/newsletters/2020/05/13/#request-for-an-additional-taproot-signature-commitment

Adding a label makes it similar to the related attack described in a newsletter a month later: https://bitcoinops.org/en/newsletters/2020/06/10/#fee-overpayment-attack-on-multi-input-segwit-transactions

I want to use both images in a different context and so it'd be nice if they both had labels (my preference) or neither had a label.


jnewbery:
Looks good to me. Thanks @harding!

PR #494, opened 2020-12-18
Spelling: s/discrete/discreet/ for DLCs

harding:
This fixes a couple puny problems.  h/t @Xekyo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,9 @@ test-before-build:
 	## Check for duplicate links in any particular topic file
 	! git grep url:  _topics/ | sed 's/ \+/ /g' | sort | uniq -d | grep .
 
+	## Check for mistakes typical spell checkers can't catch
+	! git --no-pager grep -i '[d]iscrete log contract'
+
 test-after-build:
 	## Check for broken Markdown reference-style links that are displayed in text unchanged, e.g. [broken][broken link]
 	! find _site/ -name '*.html' | xargs grep ']\[' | grep -v skip-test | grep .
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,13 @@
 all: test-before-build build test-after-build
 production: all production-test
 
+## If we call git in our tests without using the --no-pager option
+## or redirecting stdout to another command, fail unconditionally.
+## This addresses an issue where `make` can't be run within an IDE because
+## it tries to paginate the output, see:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -54,6 +54,9 @@ test-before-build:
 	## Check for duplicate links in any particular topic file
 	! git grep url:  _topics/ | sed 's/ \+/ /g' | sort | uniq -d | grep .
 
+	## Check for mistakes typical spell checkers can't catch
+	! git grep -i '[d]iscrete log contract'
```
bitschmidty, line 58: Suggest to append ` | grep .` here to prevent interruption of the build process

bitschmidty (commented):
Tested:

- checked that discrete log contract is caught in a file
- checked that successful build occurs when check passes
- laughed at style guide update
- reviewed previous discreet mentions

One issue I ran into was that the build process was interrupted by the new git grep check and required me to close text editor to continue the build. This occurred whether the check passed or failed. I made a suggestion which appears to resolve the issue on my (macOS) end.

Makefile
harding: Oh, we had this problem before, didn't we?  Can you try: `! git --no-pager grep -i '[d]iscrete log contract'`

bitschmidty: @harding That change works as well (both success and fail behave as expected)

harding:
Force pushed a fix for @bitschmidty's IDE pager issue as well as an additional test that should also produce a failure if anyone tries to write a future test that will have the same issue.

STYLE.md
```diff
@@ -109,6 +109,7 @@ spelled the same.
 | Use | Don't use | Notes |
 |-|-|-|
 | adaptor signatures | adapter signatures | |
+| discreet log contracts | discrete<!-- --> log contracts | Remembering to spell this correctly more often than one time in ten is known as the discrete log problem |
```
jnewbery, line 112: I don't understand the `<!-- -->` here. Is that an html comment?

harding: Yes, an empty comment so that the `git grep` used in the test doesn't match this deliberate violation of the rule.  For the same reason, we use `[d]` instead of `d` in the actual pattern in the Makefile.

jnewbery: ah! Thanks

PR #495, opened 2020-12-19
Update 2020-10-07-newsletter.md

MarcoFalke:
Maybe I am misunderstanding, but this doesn't sound right.

If changing past newsletters is not allowed, please close.


harding:
+100 for correcting the content, but for changes that significantly change the meaning of the text in a document that's already been published, we need to note that we've edited the text.  I'll open a separate PR that includes your commit plus a correction notice.  (Please leave this PR open until then; I'll @ mention you on the other PR.)

Thanks!

jnewbery:
Thanks for reporting this @MarcoFalke!

PR #502, opened 2021-01-11
News88: fix typo

harding:


jnewbery:
Thanks @harding! The `include linkers/issues.md` include at the bottom also needs to be updated.

PR #505, opened 2021-01-18
Typo fix: spelling of Lloyd Fournier's name

harding:
Reported by @ajtowns .  Git grep shows no other incidences of this misspelling.


jnewbery:
Misattributing to Fourier is a very series matter. We should do some analysis on how it happened.

harding:
Three months later, while almost making the same typo again in a draft, I realized I missed an incredible opportunity to title this PR:

> Apply a Fourier transform

PR #510, opened 2021-01-27
Matrix: quick updates to bech32 sending

### LINKED ISSUE #504: bech32 compatibility: Seems outdated
BitMEX definitely supports any v0 bech32 address:
https://blog.bitmex.com/bitmex-enables-bech32-sending-support/

Seems like mycellium supports some level of bech32 since 2018: https://github.com/mycelium-com/wallet-android/issues/425

Jaxx seems to support sending to: https://twitter.com/Jaxx_Support/status/948982835137994754

bitstamp appears to support deposits and sending: https://www.bitstamp.net/article/weve-added-support-bech32-bitcoin-addresses-bitsta/

Abra: no info on this

Binance: No evidence to suggest they support sending to p2wsh. I'll gather more info.

harding:
Closes #504 .  This updates a few of the entries in the table based on evidence provided by @instagibbs in the linked issue.


bitschmidty:
@harding thanks for throwing these together.

Im supportive of the bitstamp and bitmex changes, with a couple fixup commits just pushed.

However, the mycelium changes, I just tested again and neither p2wsh nor p2wpkh addresses are accepted in the UI.

Jaxx changes reference materials before my testing. So I saw with my eyes on those days that bech32 was not supported. But unable to get the wallet working at the moment. Inclined to leave jaxx as is on the site now.

_data/compatibility/jaxx.yaml
```diff
@@ -40,7 +40,7 @@ segwit:
       bech32: "false"
       default: "p2pkh"
     send:
-      bech32: "false"
```
bitschmidty, line 43: I am unable to test this wallet to verify

_data/compatibility/mycelium.yaml
```diff
@@ -39,17 +39,17 @@ rbf:
...
   tested:
-    date: "2019-07-24"
+    date: "2021-01-27"
     platforms:
       - iOS
     version: "14"
   features:
     receive:
       p2sh_wrapped: "false"
-      bech32: "false"
```
bitschmidty, line 49: Cannot see how to do this in the UX currently

_data/compatibility/mycelium.yaml
```diff
@@ -39,17 +39,17 @@ rbf:
...
+      bech32: "true"  # https://github.com/mycelium-com/wallet-android/issues/425#issuecomment-440792004
       default: "p2pkh"
     send:
-      bech32: "false"
+      bech32: "true" # https://github.com/mycelium-com/wallet-android/issues/425#issuecomment-440792004
```
bitschmidty, line 52: Not possible due to validation error in the UX for me

_data/compatibility/mycelium.yaml
```diff
@@ -39,17 +39,17 @@ rbf:
...
       change_bech32: "false"
       segwit_v1: "Fails during address validation."
       bech32_p2wsh: "false"
```
bitschmidty, line 55: Same as above

pointbiz:
Mycelium works for me with bech32. 
Abra App still doesn't support bech32.
Wirex App doesn't support bech32. Should it be added to the list? 

I also heard complaints from multiple people that blockchain.com wallet doesn't support bech32. However, I don't have that wallet to verify.

instagibbs:
Some more updates based on boots on the ground experience:

iOS blockchain.info wallet doesn't support it yet, but they claim Soon(TM)
ZebPay exchange doesn't support(and doesn't seem to want to)
UpBit supports it
ByBit does not.

Xekyo:
Sergej Kotliar [posted](https://twitter.com/ziggamon/status/1393542956322279424) that he was able to send with his Blockchain app to a Bech32 address. Not clear whether that was Android or iOS.

Xekyo:
It appears that Blockchain now also has updated their wallets to receive to P2WPKH addresses by default and produce P2WPKH change outputs.

instagibbs:
@Xekyo can you check on p2wsh support for sending? sometimes wallets don't support both...

Xekyo:
Sorry, I don't have a Blockchain wallet.

harding:
Updated based on the comments here and rebased on master.

_data/compatibility/jaxx.yaml
```diff
@@ -30,7 +30,7 @@ rbf:
         Receiving Replacement Transaction - Replacement transaction shown, original disappears.
 segwit:
   tested:
-    date: "2021-01-27"
+    date: "2019-07-24"
```
Xekyo: Just checking, because it's not mentioned in the commit message, was this rollback intended?

harding: Yes, @bitschmidty said he was unable to verify https://github.com/bitcoinops/bitcoinops.github.io/pull/510#discussion_r568727862 , so I figured it was easier to drop it in this PR in order to get this PR merged.  Sorry for not mentioning that in the commit message.

_data/compatibility/mycelium.yaml
Xekyo: Should this then maybe also be reverted?

Xekyo (commented):
Looks reasonable, just got an open follow-up on Mycelium.

harding:
> Looks reasonable, just got an open follow-up on Mycelium.

Most recent comment on this thread was that it worked. https://github.com/bitcoinops/bitcoinops.github.io/pull/510#issuecomment-772978577

Xekyo:
Ah okay. I thought it was surprising that the date was set to the past. In that case, the update looks good to me. What's would be the next step to ship it?

_data/compatibility/mycelium.yaml
Xekyo: I downloaded and tested Mycelium, added a commit to update the information on Mycelium's segwit compatibility.

Xekyo:
I added a commit to update the Mycelium segwit compatibility data after testing it with Android today.

bitschmidty:
I noticed some of the new images were not optimized, so I optimized those and squashed.

Xekyo:
Thanks for optimizing the images. I'm a bit surprised at the squash. I had left a detailed description in the commit message for the Mycelium update. Squashing the commits dropped that information. Is there a general rule of thumb when I should expect things to be squashed and when we keep the commit history as with the newsletter?

bitschmidty:
Hey @Xekyo that’s a good point. Im happy to keep the commits separate to preserve that additional information in the commit message. Or could pull commit info into a squashed commit? Im open to suggestions on general rule for this sort of thing.

bitschmidty:
I squashed the image optimizations into the mycelium commit.

Xekyo:
I like cleaning up my commit history by squashing tiny commits that belong together to one commit (i.e. a follow-up commit that fixes a single typo shouldn't be a separate commit), but beyond that I prefer keeping all commits as it makes it easier to assess changes in their distinct context later without throwing a bunch of unrelated changes together in one change-set.

For example here, I'd say the "clean way" would be:
* The first commit makes updates to four different services while the following two commits just make corrections to the first commit. Those three commits I'd squash into one.  
* The commit with the blockchain.info update was made much later, and it also reverts the changes to the Jaxx wallet. I'd pick the changes to Jaxx into the first commit (reverting changes to Jaxx altogether rather than introducing them and rolling them back in the same PR in a later commit), while I would leave the blockchain.info update as a distinct commit.  
* I'd leave the Mycelium commit separate because Mycelium was already previously updated in the first commit, and the second update switches the assessment from iOS to Android, and was presumably gathered using a different methodology than the first updates.

In practice, investing that much time would be overkill on this series of commits, so I'd propose that we squash the first four commits and leave the Mycelium commit separate.

bitschmidty:
@Xekyo Good to get your perspective on this. Ive taken your recommendation and squashed the first 4 into one.

Xekyo (approved):
LGTM, I'm an author though.

harding:
Quick skimmed the diff and LGTM.

IMO, if @bitschmidty is happy, that's enough to merge.  The fact that information on these pages can become outdated and so break reader expectations implies to me that us maybe accidentally misprinting some information isn't the end of the world.  So I think we don't need to be overcautious.

PR #513, opened 2021-02-02
Topics: date index: display entries after 2020

harding:
This allows the [topic date index](https://bitcoinops.org/en/topic-dates/) to display entries after 2020.


bitschmidty (commented):
tACK 994970e35454f3db45390857dd4be8b3c2e46528

PR #542, opened 2021-04-01
New topic: DLCs

harding:


jnewbery (commented):
Looks great, Dave!

What do you think about adding a note that the original DLC construction used a trick with schnorr signatures. Later, a version that used adaptor signatures in ECDSA was developed. Is that too much implementation detail?

_topics/en/discreet-log-contracts.md
```diff
@@ -0,0 +1,80 @@
...
+  oracle doesn’t need to know the terms of the contract (or even that
+  a contract was made).
+
+## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
+extended_summary: |
+  The transactions creating and settling the contract can be made
+  indistinguishable from many other Bitcoin transactions or they can be
+  executed within an LN channel. This makes DLCs more private and
+  efficient than other known oracle-based contract methods, and it’s
+  arguably more secure as an oracle that commits to a false result
```
bitschmidty, line 35: Perhaps Im missing something here, but I dont see how the indistinguishable-ness of the contract's txs is related to oracle generating clear evidence of fraud?

harding:
Updated, thanks!

PR #547, opened 2021-04-07
Newsletter-143:Translate into Japanese

azuchi:


bitschmidty (commented):
Wow, very fast @azuchi , thank you!

ACK - checked links, built successfully

Pinging @AkioNak if you have time to review content

PR #548, opened 2021-04-07
Topic: Clarify that Bech32 is case sensitive

Xekyo:
Bech32 only allows uppercase or lowercase, but mixed case is explicitly forbidden. Thus, calling it case insensitive is a bit inaccurate.

### DIFF
--- a/_topics/en/bech32.md
+++ b/_topics/en/bech32.md
@@ -16,8 +16,8 @@ excerpt: >
 
 ## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
 extended_summary: |
-  Using only 32 letters and numbers, the bech32 address format is case
-  insensitive and includes an error-correction code that can catch
+  Using only 32 letters and numbers, the bech32 address format does not use
+  mixed case and includes an error-correction code that can catch
   almost all address typos (and even identify where the typos occur in
   some cases).  Addresses encode a segwit version, making them forward
   compatible with a large range of conceivable upgrades.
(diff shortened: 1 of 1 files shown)


_topics/en/bech32.md
```diff
@@ -16,8 +16,8 @@ excerpt: >
 
 ## Optional.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
 extended_summary: |
-  Using only 32 letters and numbers, the bech32 address format is case
-  insensitive and includes an error-correction code that can catch
+  Using only 32 letters and numbers, the bech32 address format does not use
+  mixed case and includes an error-correction code that can catch
```
jonatack, line 20: maybe "is case insensitive (but mixed case is invalid)"

jnewbery:
LGTM. Thanks @Xekyo!

jonatack (commented):
ACK. Per BIP350, from line 197

```
The following list gives invalid segwit addresses and the reason for
their invalidity.

...

* tb1p0xlxvlhemja6c4dqv22uapctqupfhlxm9h8z3k2e72q4k9hcz7vq47Zagq: Mixed case
```

_topics/en/bech32.md
Xekyo: I was considering phrasing it as "is single case", but I wasn't sure whether that would be as clear.

jonatack: Agree, would've scratched my head with that one.

PR #549, opened 2021-04-07
topics: anchor outputs fixup

jonatack:
Noticed this after tweeting to consult this topic!

https://twitter.com/jonatack/status/1379839115676561408


jnewbery:
This is a nexcellent suggestion

PR #537, opened 2021-03-25
New topic: channel jamming attacks

harding:

PR #551, opened 2021-04-13
Improve markdown rendering of topic content

harding:
Unlike most other content on the site, the content of topic pages is rendered by the Liquid `{{markdownify}}` filter.  This means topic pages don't use the auto anchor link plugin we have and can't `{% include %}` files.

This PR fixes that.  Don't let it's huge diff scare you off---all but a few lines are moves, whitespace changes, and boilerplate.  I suggest reviewing using:

    git checkout <this branch>
    make  # build the site
    cp -a _site _new_site
    git reset --hard HEAD^^
    make  # build the site before these changes
    diff -ruN _site _new_site | colordiff | less -R

Reviewing the above diff, there are a few extraneous whitespace changes where I fixed an odd indentation I came across during the move, but every other change is just the addition of a link anchor bullet.

This PR is fully backwards compatible with the way we previously rendered markdown content, so it doesn't interfere with pending new topics (e.g. #537, etc).  When all those pending topics have been merged, I'll open a cleanup PR to convert them to this update and remove the old mechanism.


_contrib/new-topic
```diff
@@ -68,7 +60,17 @@ optech_mentions:
 # see_also:
 #   - title:
 #     link:
+
+## Optional extended summary.  Use Markdown formatting.  Multiple paragraphs.  Links allowed.
```
jnewbery, line 64: It seems strange to me to have this comment before the end of the front matter section. Perhaps it can just be removed entirely.

jnewbery (commented):
Looks good @harding. I've compared the generated sites and all of the differences are as expected.

One minor suggestion inline.

jnewbery:
> This PR is fully backwards compatible with the way we previously rendered markdown content, so it doesn't interfere with pending new topics (e.g. #537, etc). When all those pending topics have been merged, I'll open a cleanup PR to convert them to this update and remove the old mechanism.

All pending topics pages have now been merged, so you could remove the backwards compatible part in this PR.

harding:
Rebased on master, updated to completely remove `extended_summary`, and addressed feedback.  Thanks @jnewbery!

I re-reviewed the output diff and don't see any unexpected changes.

jnewbery:
ACK be20c1c332d792ec36fe5d626cbe725e1b8c0f0d

PR #552, opened 2021-04-13
compatibility data: trezor update to trezor suite

prusnak:
I've updated the Trezor compatibility data with the latest Suite in mind. Suite will be the default wallet for Trezor in few weeks and has been in open beta since December 2020. It is available from https://suite.trezor.io

I haven't updated all RBF screenshots - 3 out of 5 still need to be replaced. I might get to that later, but the rest is I believe correct.


bitschmidty (approved):
@prusnak thank you for the contribution.

Data looks properly formatted (I did not re-test these to validate, myself). Images look good/optimized.

PR #553, opened 2021-04-16
Newsletter-144:Translate into Japanese

azuchi:


bitschmidty (commented):
Im able to build and all links checked out

bitschmidty:
Thanks @azuchi !

PR #560, opened 2021-04-23
New topic: BIP152 compact block relay

harding:


_topics/en/compact-block-relay.md
```diff
@@ -0,0 +1,113 @@
...
+txid (wtxid).  This can take a multimegabyte block with thousands of
+transactions and compact it down to just a few kilobytes.  This directly
+reduces bandwidth.  Indirectly, it also significantly reduces latency
+because TCP propagates smaller amounts of data faster than larger
+amounts.
+
+For any transactions the generating node doesn't think its peer knows
+about, it sends the full transaction.
+
+The receiving peer hashes the wtxids of all the transactions it is
```
jnewbery, line 74: s/it is/in its/ ?

_topics/en/compact-block-relay.md
```diff
@@ -0,0 +1,113 @@
...
+unconfirmed transactions, which the nodes store in their mempools and
+further relay to their other peers.  Ultimately those transactions
+reach miners, who attempt to include them in a block.  When that new
+block is itself relayed back to nodes, it typically consists almost
+entirely of transactions that the nodes already have in their mempools.
+
+Relaying those transactions a second time is redundant, and it's that
+redundancy that [BIP152][] compact blocks aims to eliminate.  A sending
+node constructs a compact block by replacing each transaction it
+believes its peer has with a 6-byte hash of that transaction's witness
```
jnewbery, line 64: Maybe spend a bit more time explaining that this 6-byte SipHash is entirely different from the txid and wtxid, which are 32 byte SHA256 hashes. The short id hash is not cryptographically secure, but it doesn't need to be.

_topics/en/compact-block-relay.md
```diff
@@ -0,0 +1,113 @@
...
+txid (wtxid).  This can take a multimegabyte block with thousands of
```
jnewbery, line 65: You only mention the hash of the wtxid here, but right at the end of the document, reveal that it could also be the hash of the txid. Perhaps introduce that here.

_topics/en/compact-block-relay.md
```diff
@@ -0,0 +1,113 @@
...
+primary_sources:
+    - title: BIP152
+
+## Optional.  Each entry requires "title", "url", and "date".  May also use "feature:
+## true" to bold entry
+optech_mentions:
+  - title: Bitcoin Core PR Review Club about dropping support for v1 compact blocks
+    url: /en/newsletters/2021/02/10/#bitcoin-core-pr-review-club
+
+  - title: "Bitcoin Core #20764 adds BIP152 high bandwidth indicator to bitcoin-cli"
```
jnewbery, line 30: to bitcoin-cli -netinfo

_posts/en/newsletters/2020-07-29-newsletter.md
```diff
@@ -169,6 +169,7 @@ release candidates.*
        transaction announcements will use more bandwidth than usual
        until everyone has upgraded to wtxid-using nodes.
 
+    {:#dont-relay-taproot}
```
jnewbery, line 172: This causes the ordered list numbers to reset to 1.

bitschmidty (commented):
lgtm @harding Just one small change

_topics/en/compact-block-relay.md
harding: Looking at how to best add this in without going into too much detail, I used "fast", "non-cryptographic", and "per peer".  I think "fast" maintains the reason for using siphash, "non-cryptographic" explicitly mentions the main downside of a fast and short hash, and "per peer" suggests the reason it doesn't need to be cryptographically secure (at worst, an attacker could only cause collisions for one of your peers, or from one of your peers).

_topics/en/compact-block-relay.md
harding: That exceeds the allowed 75 characters.

_posts/en/newsletters/2020-07-29-newsletter.md
harding: Oof, thanks for catching that!

_topics/en/compact-block-relay.md
jnewbery: The compactid of the transaction is not per-peer (see https://github.com/bitcoin/bips/blob/master/bip-0152.mediawiki#short-transaction-ids and https://github.com/bitcoin/bitcoin/blob/eb4df9a628bdcdd1333c7fa4fb23c73df9642902/src/blockencodings.cpp#L31-L45). (I also thought it was per-peer for some reason).

The other changes look good.

_topics/en/compact-block-relay.md
jnewbery: ok!

jnewbery:
All previous review comments have been addressed. I have one additional comment (https://github.com/bitcoinops/bitcoinops.github.io/pull/560#discussion_r636797844).

_topics/en/compact-block-relay.md
harding: BIP152 says:

> We mix in a per-connection 64-bit nonce to obtain independent short IDs on every connection, so that even block creators cannot control where collisions occur, and random collisions only ever affect a small number of connections at any given time.

and (at the point you link, emphasis added):

> Short transaction IDs are used to represent a transaction without sending a full 256-bit hash. They are calculated by:
>
 > 1. single-SHA256 hashing the block header **with the nonce appended **(in little-endian)

The nonce is used in the code you linked (it's initialized in the preceding code block).

jnewbery: Ah, thanks! You're right that a different nonce may be used on each connection.

Note also that the node may choose to use the same nonce for all of its peers for a single block:

> Nodes MAY pick one nonce per block they wish to send, and only build a cmpctblock message once for all peers which they wish to send a given block to. Nodes SHOULD NOT use the same nonce across multiple different blocks.

Which is what we do in Bitcoin Core:

- net_processing is notified of a new block with valid work: https://github.com/bitcoin/bitcoin/blob/7041d256e3a951dc7cc8b10db32e6f21d2071806/src/net_processing.cpp#L1432
- a `CBlockHeaderAndShortTxIDs` object is constructed, which uses a randomly generated nonce: https://github.com/bitcoin/bitcoin/blob/7041d256e3a951dc7cc8b10db32e6f21d2071806/src/net_processing.cpp#L1434
- That `CBlockHeaderAndShortTxIDs` object is serialized and sent to all hb compact block peers: https://github.com/bitcoin/bitcoin/blob/7041d256e3a951dc7cc8b10db32e6f21d2071806/src/net_processing.cpp#L1470
- The `CBlockHeaderAndShortTxIDs` is also cached in the global `most_recent_compact_block`, which is used if a peer requests a compact block of the tip: https://github.com/bitcoin/bitcoin/blob/7041d256e3a951dc7cc8b10db32e6f21d2071806/src/net_processing.cpp#L1782.

I think it's fine either to leave "per-peer" in this description or remove it. I'm happy with whichever you decide.

PR #564, opened 2021-04-28
Newsletter-146:Translate into Japanese

azuchi:


bitschmidty (approved):
tACK (build and links)

Thank you @azuchi !

PR #567, opened 2021-05-04
[build] Delete all index pages in `make preview`

jnewbery:
It sometimes trips me up when I run `make preview` locally and don't see new newsletters/topic pages/blog posts in the index page.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,14 @@ clean:
 	bundle exec jekyll clean
 
 preview:
-	## Don't do a full rebuild (to save time), but always rebuild index.html
-	rm _site/index.html || true
+	## Don't do a full rebuild (to save time), but always rebuild index pages.
+	rm -f _site/index.html \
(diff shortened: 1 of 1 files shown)


harding:
ACK

It looks like maybe the official way to do this is documented here: https://jekyllrb.com/docs/configuration/incremental-regeneration/

> To remedy some of these shortfalls, putting `regenerate: true` in the front-matter of a document will force Jekyll to regenerate it regardless of whether it has been modified. Note that this will generate the specified document only; references to other documents’ contents will not work since they won’t be re-rendered.

I'm not sure how much that will slow down regular incremental rebuilds while I'm editing pages, so I suggest going with your `rm` method for now (which helps reviewers) and I'll look into the the above technique for later.

jnewbery:
Thanks Dave. `regenerate: true` does seem like it's the right way to go. Let me know what you find.

PR #574, opened 2021-05-19
Topics: improve topic descriptions added in Newsletter 149

harding:
This addresses two comments from @jnewbery regarding the topic links in Newsletter 149.  I do leave unaddressed a couple comments from John about the newsletter content itself, but I think of newsletter content as being mostly static once published, while the topic pages are dynamic.


jnewbery:
Looks good. Thanks @harding!

PR #575, opened 2021-05-20
Newsletter-149:Translate into Japanese

azuchi:


bitschmidty:
Build, tested links, looks good!

Thanks @azuchi!

PR #579, opened 2021-05-25
Style Guide: prefer hardware signers over hardware wallets

jnewbery:


jonatack:
ACK! 

Provided that people understand what we are referring to but education here is 👍

IDK if we (informally) need a transitory "hardware signers (hardware wallets)"

bitschmidty:
ACK. I think its good to focus on such distinctions.

Also, agreed with the transitory parentheticals or the like.

Xekyo:
ACK, I'd say that these devices are also more generally referred to as "Personal Hardware Security Modules", but "hardware signing devices" seems like a good improvement over "hardware wallets".

Xekyo (commented):
ACK

Nit: I chafe a bit at describing them as "offline" since they often get plugged into a "hot device" to interact with the wallet. I'd describe them as personal hardware security modules, or dedicated hardware devices to extract the key storage and signing aspect from the wallet software.

harding:
ACK

Maybe s/offline signing devices/dedicated key storage and signing devices/

jnewbery:
> Nit: I chafe a bit at describing them as "offline" ...

> Maybe s/offline signing devices/dedicated key storage and signing devices/

Fixed. Thanks!

PR #580, opened 2021-05-27
Newsletter-150:Translate into Japanese

azuchi:


bitschmidty (approved):
ACK e170d3f153be1d189dbaadd405668c98654e3b28

Thank you @azuchi for your translation!

PR #588, opened 2021-06-10
Newsletter-152:Translate into Japanese

azuchi:


bitschmidty:
LGTM @azuchi , thank you!

PR #593, opened 2021-06-23
News154: fix Bech32m header

jonatack:
Spotted by Adam Gibson / @AdamISZ.


bitschmidty (commented):
Build failure due to broken link since the topics feature uses the anchor link which uses the header text. Pushed a change for that.

jonatack:
Ahh that's why -- thanks @bitschmidty!

PR #597, opened 2021-06-26
Newsletters: add 155 (2021-06-30)

harding:
- [x] Update release/RCs @harding
- [x] Add topic links when all content done @harding
- [x] Stack Exchange section @bitschmidty
- [x] `Bitcoin Core #22154` Add OutputType::BECH32M and related wallet support for fetching bech32m addresses @harding
- [x] `Bitcoin Core #22166` Add support for inferring tr() descriptors @Xekyo
- [x] `BOLTs #877` Remove HTLC amount restriction @dongcarl

### DIFF
--- a/_includes/specials/taproot/en/01-single-sig.md
+++ b/_includes/specials/taproot/en/01-single-sig.md
@@ -1,0 +1,75 @@
+Using Optech's [transaction size calculator][], we can compare the sizes
+of different types of single-sig transactions.  As expected,
+transactions using P2WPKH inputs and outputs are much smaller than those
+using P2PKH inputs and outputs---but, perhaps surprisingly, P2TR
+transactions are slightly larger than equivalent P2WPKH transactions.
+
+|                    | P2PKH (legacy) | P2WPKH (segwit v0) | P2TR (taproot/segwit v1) |
+|--------------------|----------------|--------------------|--------------------------|
+| **Output**         |  34            | 31                 | 43                       |
+| **Input**          |  148           | 68                 | 57.5                     |
+| **2-in, 2-out tx** |  374           | 208.5              | 211.5                    |
+
+That may make it seem counterproductive for single-sig wallets to
+implement taproot spending in preparation for block {{site.trb}}, but a
+closer look reveals that there are a number of advantages to using P2TR
+for single-sigs, both for wallet users and for the network as a whole.
+
+- **Cheaper to spend:** it costs about 15% less at the input level to
+  spend a single-sig P2TR UTXO than it does to spend a P2WPKH UTXO.  An
+  overly simple analysis like the table above hides the detail that the spender
+  can't choose which addresses they're asked to pay, so if you stay on
+  P2WPKH and everyone else upgrades to P2TR, the actual typical size of
+  your 2-in-2-out transactions will be 232.5 vbytes---while all-P2TR
+  transactions will still only be 211.5 vbytes.
+
+- **Privacy:** although some privacy is lost when early adopters change
+  to a new script format, users switching to taproot also immediately
+  receive a privacy boost.  Your transactions will be able to look
+  indistinguishable from people working on new LN channels, more
+  efficient [DLCs][topic dlc], secure [multisignatures][topic
+  multisignature], various clever wallet backup recovery schemes, or a
+  hundred other pioneering developments.
+
+    Using P2TR for single-sig now also allows your wallet to upgrade to
+    multisignatures, tapscripts, LN support, or other features later on
+    without affecting the privacy of your existing users.  It won't
+    matter whether a UTXO was received to an old version or a new
+    version of your software---both UTXOs will look the same onchain.
+
+- **More convenient for hardware signing devices:** since the
+  rediscovery of the [fee overpayment attack][news101 fee overpayment
+  attack], several hardware signing devices have refused to sign a
+  transaction unless each UTXO spent in that transaction is accompanied
+  by metadata containing a copy of significant parts of the entire
+  transaction which created that UTXO.  This greatly increases the
+  worst-case processing that hardware signers need to perform and is
+  especially problematic for hardware signers using limited-size QR
+  codes as their primary communication medium.  Taproot eliminates the
+  vulnerability underlying the fee overpayment attack and so can
+  significantly improve the performance of hardware signers.
+
+- **More predictable feerates:** ECDSA signatures for P2PKH and P2WPKH
+  UTXOs can vary in size (see [Newsletter #3][news3 sig size]).  Since
+  wallets need to choose a transaction's feerate before creating the
+  signature, most wallets just assume the worst case signature size and
+  accept that they'll slightly overpay the feerate when a smaller
+  signature is generated.  For P2TR, the exact size of the signature is
+  known in advance, allowing the wallet to reliably choose a precise
+  feerate.
+
+- **Help full nodes:** the overall security of the Bitcoin system
+  depends on a significant percentage of Bitcoin users verifying every
+  confirmed transaction with their own nodes.  That includes verifying
+  the transactions your wallet creates.  Taproot's [schnorr
+  signatures][topic schnorr signatures] can be efficiently batch
+  verified, [reducing by about 1/2][batch graph] the number of CPU
+  cycles nodes need to expend when verifying signatures during the
+  process of catching up on previous blocks.  Even if you rejected every
+  other point on this list, consider preparing to use taproot for the
+  benefit of people running full nodes.
+
+[transaction size calculator]: /en/tools/calc-size/
+[news3 sig size]: /en/newsletters/2018/07/10/#unrelayable-transactions
+[news101 fee overpayment attack]: /en/newsletters/2020/06/10/#fee-overpayment-attack-on-multi-input-segwit-transactions
+[batch graph]: https://github.com/jonasnick/secp256k1/blob/schnorrsig-batch-verify/doc/speedup-batch.md
--- a/_posts/en/newsletters/2021-06-30-newsletter.md
+++ b/_posts/en/newsletters/2021-06-30-newsletter.md
@@ -46,7 +46,51 @@ answers posted since our last update.*
 {% comment %}<!-- https://bitcoin.stackexchange.com/search?tab=votes&q=created%3a1m..%20is%3aanswer -->{% endcomment %}
 {% assign bse = "https://bitcoin.stackexchange.com/a/" %}
 
-FIXME:bitschmidty
+- [What are the downsides to enabling potentially suboptimal or unused opcodes in a future soft fork?]({{bse}}106851)
+  G. Maxwell outlines many considerations for enabling any opcodes that affect
+  consensus including:
+
+    * upfront as well as ongoing maintenance costs
+
+    * potential risks to the user of the opcode but also the entire network
+
+    * additional complexity acting as a disincentive for customizing or reimplementing node software
+
+    * partial or inefficient features crowding out better future alternative ones
+
+    * accidentally creating perverse incentives
+
+- [Why does blockwide signature aggregation prevent adaptor signatures?]({{bse}}107196)
+  Pieter Wuille explains why cross-input signature aggregation interferes
+  with techniques like [adaptor signatures][topic adaptor signatures] or
+  scriptless scripts, noting: "In case of block-wide signature aggregation, there
+  is just a single signature for the entire block. There is simply no space for
+  that single signature to reveal multiple independent secrets to multiple
+  independent parties."
+
+- [Should the Bitcoin Core wallet (or any wallet) prevent users from sending funds to a Taproot address pre activation?]({{bse}}107186)
+  Murch makes the case for why wallet software should allow users to send to any future BIP173
+  segwit output types. By putting the onus on the receiver to provide a spendable
+  address, the ecosystem can take advantage of the forward-compatibility of
+  [bech32/bech32m][topic bech32] and instantly utilize new output types.
+
+- [Why are the witnesses segregated with schnorr signatures?]({{bse}}106930)
+  Dalit Sairio asks why, since [schnorr signatures][topic schnorr signatures] do
+  not suffer from the same malleability that ECDSA signatures do, schnorr
+  signatures will still be segregated? Darosior points out that malleability is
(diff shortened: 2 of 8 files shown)


_posts/en/newsletters/2021-06-30-newsletter.md
```diff
@@ -0,0 +1,167 @@
...
+    PSBT can omit copies of previous transactions because taproot fixes
+    the fee overpayment attack against v0 segwit inputs (see
+    [Newsletter #101][news101 fee overpayment attack]).
+
+- **Key derivation path for single-sig P2TR:** Andrew Chow also
+  [posted][chow taproot path] a second [proposed BIP][bip-taproot-bip44]
+  to the Bitcoin-Dev mailing list suggesting a [BIP32][] derivation path
+  to use for wallets creating single-sig taproot addresses.  Chow notes
+  that the BIP is very similar to [BIP49][] for P2SH-wrapped P2WPKH
+  addresses and [BIP89][] for native P2WPKH addresses.
```
bitschmidty, line 37: s/BIP89/BIP84

https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki

_includes/specials/taproot/en/01-single-sig.md
```diff
@@ -0,0 +1,75 @@
...
+  the transactions your wallet creates.  Taproot's [schnorr
+  signatures][topic schnorr signatures] can be efficiently batch
+  verified, [reducing by about 1/2][batch graph] the number of CPU
+  cycles nodes need to expend when verifying signatures during the
+  process of catching up on previous blocks.  Even if you rejected every
+  other point on this list, consider preparing to use taproot for the
+  benefit of people running full nodes.
+
+[transaction size calculator]: /en/tools/calc-size/
+[news3 sig size]: http://devbcop/en/newsletters/2018/07/10/#unrelayable-transactions
```
bitschmidty, line 73: s/http://devbcop/

_posts/en/newsletters/2021-06-30-newsletter.md
```diff
@@ -0,0 +1,172 @@
...
+  descriptors (see [Newsletter #152][news152 p2tr descriptors]).  If a
+  transaction includes any bech32m addresses after taproot activation,
+  the descriptor wallet will also use a P2TR change output.
+
+- [Bitcoin Core #22166][] adds support for inferring tr() descriptors
+  from outputs, completing basic taproot descriptor support. While
+  descriptors are a compact encoding for a set of addresses, this allows
+  the wallet to look up the descriptor from an address and the
+  descriptor expansion. Descriptor inference is used to provide more
+  accurate information in responses to RPC calls such as `listunspent`.
```
Xekyo, line 134: Ran it past Pieter, who corrected my description for accuracy, but perhaps it is needlessly fine-grained and it would be kinder to our readers to just give the need-to-know version:

```suggestion
- [Bitcoin Core #22166][] adds support for inferring tr() descriptors
  from outputs, completing basic taproot descriptor support.
```

harding: I'm splitting the difference, including all the text from your short version and some of the text from your long version.

_posts/en/newsletters/2021-06-30-newsletter.md
```diff
@@ -0,0 +1,161 @@
...
+
+- [Bitcoin Core #21056][] adds a new `-rpcwaittimeout` parameter to
+  `bitcoin-cli`.  The existing `-rpcwait` parameter will delay sending a
+  command (RPC call) until the `bitcoind` server has started.  The new
+  parameter stops the waiting after the indicated number of seconds,
+  returning an error.
+
+- [C-Lightning #4606][] allows creating invoices over about 0.043 BTC,
+  following a similar change in LND (see [Newsletter #93][news93
+  lnd4075]) and a change to the specification described in the next item.
```
Xekyo, line 140: Seems more concrete, "the next item" came a bit as a surprise after "a"
```suggestion
  lnd4075]) and the change to the specification described in the next item.
```

_topics/en/taproot.md
```diff
@@ -182,6 +182,9 @@ optech_mentions:
   - title: "Bitcoin Core #21365 allows the wallet to create signatures for P2TR spends"
     url: /en/newsletters/2021/06/23/#bitcoin-core-21365
 
+  - title: Proposed BIP to standardized a wallet path for single-sig P2TR addresses
```
Xekyo, line 185:
```suggestion
  - title: Proposed BIP to standardize a wallet path for single-sig P2TR addresses
```

Xekyo (commented):
I especially love the taproot series! Also checked all the links to Stackexchange (but not the others).

_includes/specials/taproot/en/01-single-sig.md
```diff
@@ -0,0 +1,75 @@
...
+of different types of single-sig transactions.  As expected,
+transactions using P2WPKH inputs and outputs are much smaller than those
+using P2PKH inputs and outputs---but, perhaps surprisingly, P2TR
+transactions are slightly larger than equivalent P2WPKH transactions.
+
+|                    | P2PKH (legacy) | P2WPKH (segwit v0) | P2TR (taproot/segwit v1) |
+|--------------------|----------------|--------------------|--------------------------|
+| **Output**         |  34            | 31                 | 43                       |
+| **Input**          |  148           | 67.75              | 57.25                    |
+| **2-in, 2-out tx** |  374           | 208.25             | 211.25                   |
```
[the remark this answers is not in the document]
harding: Wow, thanks for the amazing fact checking!  Looking at the source code for https://bitcoinops.org/en/tools/calc-size/ and comparing it to your B.SE answer, it appears I omitted the `count witness items` from all the segwit types, which explains both discrepancies above.  Will fix!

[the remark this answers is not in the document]
Xekyo: Oh, good. I think I had looked at that before and hadn't noticed. :flushed:
It's one of those seldom but gratifying instances where intimate familiarity with input sizes is actually useful. :laughing:

_includes/specials/taproot/en/01-single-sig.md
```diff
@@ -0,0 +1,75 @@
...
+  transaction unless each UTXO spent in that transaction is accompanied
+  by metadata containing a copy of significant parts of the entire
+  transaction which created that UTXO.  This greatly increases the
+  worst-case processing that hardware signers need to perform and is
+  especially problematic for hardware signers using limited-size QR
+  codes as their primary communication medium.  Taproot eliminates the
+  vulnerability underlying the fee overpayment attack and so can
+  significantly improve the performance of hardware signers.
+
+- **More predictable feerates:** ECDSA signatures for P2PKH and P2WPKH
```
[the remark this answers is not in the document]
harding: Grinding increases the likelihood that your predicted feerate will be your obtained feerate, but it doesn't guarantee it, since you could always end up with an even smaller R than you expected (e.g. you have a `1 / 2 / 256` chance at a 70-byte signature, a `1 / 2 / 256^2` chance at 69 bytes, etc..).

Additionally, while I think signature grinding is probably overall a good idea, it also feels to me a bit like rolling your own crypto.  Andrew Chow's PR to grind in Bitcoin Core had review by experienced cryptographers; but I wouldn't blame a cautious solo wallet dev about being reluctant to do something that might risk introducing bias into his signatures or otherwise risking problems the dev didn't fully understand.

_includes/specials/taproot/en/01-single-sig.md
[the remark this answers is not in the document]
harding: Oh, when I went to edit the code, I discovered it does seem to be correct.  We count `count witness items` as part of the transaction overhead (since it's only used once per tx, not per input).  So I think both the P2WPKH and P2TR examples are correct as written.

harding:
Thanks @bitschmidty @dongcarl @Xekyo for the writing and @bitschmidty and @Xekyo for the reviews!

_includes/specials/taproot/en/01-single-sig.md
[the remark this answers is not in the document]
Xekyo: No, `count witness item` is definitely per input, not per transaction. For P2WPKH it is two (signature and public key), for P2TR it is one (signature only).

Interesting thing I learned some time ago, later than I should have: there must be a witness stack for every input as soon there is a single segwit input. For non-segwit inputs this is just a 0 for the `count witness items`. This actually causes an input size non-linearity that had baffled me for a while then, since adding the first segwit input increases the whole transaction weight by more than just the segwit input's weight.

_includes/specials/taproot/en/01-single-sig.md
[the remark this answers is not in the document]
Xekyo: As far as I'm aware issues with incorrect fees are caused by underpaying fees (like conflicting with the `minRelayTxFeeRate`, or Bitpay's inane minimum fees), so getting a smaller `r` simply increases the feerate of your transaction beyond the intended, which happens all the time with conservative estimates anyway.

The "roll your own" crypto makes sense to me, though, although, I would suspect that most wallets use some standard library like `libsecp`, or `bitcoinjs`.

_includes/specials/taproot/en/01-single-sig.md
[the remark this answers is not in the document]
harding: > No, count witness item is definitely per input, not per transaction.

Huh, today I learned.  It looks like I only skimmed BIP141, which gives a misleading impression with:

> `[nVersion][marker][flag][txins][txouts][witness][nLockTime]`

(How come `txins` and `txouts` are plural, but `witness` is singular?)

But reading a bit further, it does say:

> Each txin is associated with a witness field.

I have some asserts to actual transactions in the code for the tool, but it looks like I only selected 1-input examples and so they didn't catch this.  I'll find a 2-input example, add it, and update the code.  Thanks so much for (1) discovering the mistake, (2) putting up with my vacillating, (3) continuing to try to show me the truth until I saw the light!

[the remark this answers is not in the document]
harding: Should be fixed now (two separate commits; one for the tx tool; one for the newsletter text).  Thanks again!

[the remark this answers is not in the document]
Xekyo: All good, glad to return the favor after learning so much from your writings. :+1:

en/tools/calc-size.md
```diff
@@ -119,9 +114,14 @@ sections are [vbytes][].  Sizes in the *common elements* section are bytes.
   input.  Used by [BIP68][] and [BIP125][], with other values having no
   defined meaning
 
-- **Witness data** (varies) The source of witness data for in segwit
+- **Witness item count** ([compactSize][]/4) The number of witness
+  elements included in transactions containing at least one segwit
+  input. 0.25 vbytes for up to 252 elements
```
Xekyo, line 119: Feels like there is something missing between elements and included. Maybe:

```suggestion
- **Witness item count** ([compactSize][]/4) The number of witness
  elements. This field is included for each input if the transaction contains at least one segwit
  input. 0.25 vbytes for up to 252 elements
```

en/tools/calc-size.md
```diff
@@ -119,9 +114,14 @@ sections are [vbytes][].  Sizes in the *common elements* section are bytes.
...
+
+- **Witness items** (varies) The source of witness data for in segwit
```
Xekyo, line 121:
```suggestion
- **Witness items** (varies) The source of witness data in segwit
```

en/tools/calc-size.md
```diff
@@ -274,7 +274,7 @@ function calculateTotal(type, inputs, outputs) {
 
   // Calculate the transaction's overhead (the size independent of inputs and outputs)
   if (types[type].segwit == true) {
-    witness_flag = {{page.size.marker_and_flag}} + {{page.size.nelements}};  // segwit marker, segwit flag, segwit # of witness elements
+    witness_flag = {{page.size.marker_and_flag}};  // segwit marker, segwit flags
```
Xekyo, line 277:
```suggestion
    witness_flag = {{page.size.marker_and_flag}};  // segwit marker, segwit flag
```

Xekyo (commented):
Ah nevermind, the calculator only uses a single type at a time.

_includes/specials/taproot/en/01-single-sig.md
```diff
@@ -0,0 +1,75 @@
...
+of different types of single-sig transactions.  As expected,
+transactions using P2WPKH inputs and outputs are much smaller than those
+using P2PKH inputs and outputs---but, perhaps surprisingly, P2TR
+transactions are slightly larger than equivalent P2WPKH transactions.
+
+|                    | P2PKH (legacy) | P2WPKH (segwit v0) | P2TR (taproot/segwit v1) |
+|--------------------|----------------|--------------------|--------------------------|
+| **Output**         |  34            | 31                 | 43                       |
+| **Input**          |  148           | 68                 | 57.5                     |
+| **2-in, 2-out tx** |  374           | 208.5              | 211.5                    |
```
shesek, line 11: Maybe worth to also list the sizes for a 2-in, 3-out transaction, where Taproot does become favorable?

harding: @shesek do you mean 3-in, 2-out?

I think I prefer to stick with the apples-to-apples comparison and link to the calculator so that anyone who wants to can play around with the numbers however they'd like.  Including more examples in the table might confuse readers and we directly address the issue of taproot inputs being cheaper in the first bullet point, so hopefully nobody will be missing the overall point.

Xekyo: "2-in, 2-out" is a good example case, no need to sugarcoat. The important parts regarding Taproot's weight increase are that the overall cost increase is minuscule, the input to output costs have better incentive alignment (increasing the UTXO set size is more expensive than for other single-sig types, but reducing it is cheaper), and, as @harding writes in this section, _the receiver picks the destination address_, but the sender pays for the output that pays to it and the receiver pays for the subsequent input. Hence, the receiver is incentivized to pick P2TR as their address to get paid to.

bitschmidty:
Great newsletter, team!

Pushed a small commit to remove the check-for-topics comment.

@harding @Xekyo @dongcarl @shesek 🙏

jnewbery:
Great work this week, team! Very happy to see that I wasn't missed at all :grinning:

PR #598, opened 2021-07-01
Newsletter-155:Translate into Japanese

azuchi:


bitschmidty (approved):
Looks great, thanks @azuchi !

PR #601, opened 2021-07-08
Add Murch to associates

Xekyo:
Am I doing this right? :innocent:


harding:
Strong ACK!  Checked netlify preview, LGTM.  (And now, after five years (?), I know what Murch looks like!)

jnewbery:
ACK. Welcome aboard

PR #609, opened 2021-07-21
P4TR: columns about sig adaptors and PTLCs

harding:
**Be careful merging:** the commit labeled for deletion needs to be removed before merge.  It's only added to this branch to allow previewing the columns.

This adds ~~three~~ two columns for our Preparing For Taproot series, one about signature adaptors and one about PTLCs.  The columns are still a bit rough and need illustrations (and, certainly, technical review) but I think they're in good enough shape for people to get an idea of what we'll be covering in those three columns.


harding:
> I think they're in good enough shape for people to get an idea of what we'll be covering in those three columns.

The above is useful because, I'm very pleased to report, @zmnscpxj and @darosior have each graciously agreed to guest write one of the columns for our series, respectively about the taproot-inspired upgrades for LN and vaults.

Xekyo:
Would you like review on these at this itime, @harding?

harding:
@Xekyo I should have time to finish the illustrations today, so a review anytime after that update would be great.

harding:
Note for myself next time I update this PR, we already have a graph of multisig vs multisignature costs that looks way better than what I made in this PR.  I should use that instead.  https://bitcoinops.org/img/posts/2019-05-taproot-multisig-size.png

harding:
Dropped the column about multisignatures since I'm including that in next week's newsletter.

_includes/specials/taproot/en/xx-ptlcs.md
```diff
@@ -0,0 +1,152 @@
...
+he receives from Alice.  Using the resultant EC point (which Bob doesn't
+know is now just the point Alice originally received from Carol), Bob
+constructs the adaptor he gives to Carol.  Carol knows the scalar for
+that final point and so converts Bob's adaptor into a valid signature.
+As before, Bob recovers Carol's scalar from her signature and uses it
+and his own scalar to convert Alice's adaptor into a valid signature.
+
+In the two hops in this path, Alice→Bob and Bob→Carol, two different
+EC points and scalars were used, eliminating linkability.  We can apply
+this to the same longer paths we examined when considering HTLC privacy.
```
Xekyo, line 122: It's not clear to me which longer paths are being referred to here. Also "same longer" seems contradictory.

Xekyo (commented):
Only reviewed the article on PTLCs, I think the other article was part of the last newsletter.

_includes/specials/taproot/en/xx-ptlcs.md
harding: I think it might be unclear to you because you only read the text in the diff.  In the preview, there are two images (one earlier, one immediately after this paragraph) that illustrate a longer path.  I'll change the full stop at the end of this sentence to a colon to maybe help make that a bit clearer.

harding:
Moved to #630 , with feedback from @Xekyo addressed in that PR (thanks Murch!).

PR #611, opened 2021-07-22
Template: update translations navigation bar

### LINKED ISSUE #595: Improve discoverability of translations
When trying to reply to [this comment](https://www.reddit.com/r/Bitcoin/comments/o144wf/bitcoin_optech_newsletter_153/h20orry/) on Reddit, it took me a while to find links to the Japanese translations---and I knew they were there.  I don't have any specific suggestions, but I think we could certainly improve discoverability.

Related: #285 for a Japanese RSS feed.

bitschmidty:
Closes #595

Adds better visualizations about translations and availability.

If someone has a solution using the existing localization-finder.md helper (this solution removed that), let me know.


_layouts/newsletter.html
```diff
@@ -3,18 +3,22 @@
...
-        {% endfor %}
-      </div>
-    {% endif %}
+    <div class="localization">
+      <a href="{{ '/en' | append:baseurl }}">en</a>
+      {% for lang in site.languages %}
+        {% assign localization = "/" | append: lang | append:baseurl %}
+        {% assign locale = site.posts | where:"permalink", localization %}
+        {% if locale.size == 0 %}
+          | <span title="No {{lang}} translation currently available">{{ lang }}</span>
```
jnewbery, line 16: Is there a way to display the full name of the language in this pop up, eg:

"No Spanish translation currently available"

instead of:

"No es translation currently available"

_layouts/newsletter.html
```diff
@@ -3,18 +3,22 @@
...
-      <div class="localization">
-        <a href="{{ '/en' | append:baseurl }}">en</a>
-        {% for locale in localizations %}
-          | <a href="{{ locale.url }}">{{locale.lang}}</a>
-        {% endfor %}
-      </div>
-    {% endif %}
+    <div class="localization">
+      <a href="{{ '/en' | append:baseurl }}">en</a>
+      {% for lang in site.languages %}
```
jnewbery, line 12: Is it possible to keep this code as a function in localization-finder.md rather than duplicate it on multiple layout templates?

jnewbery (commented):
Looks good to me. A couple of small suggestions inline

Did you see the comment https://github.com/bitcoinops/bitcoinops.github.io/issues/595#issuecomment-871430440:

> The greyed out link for languages that don't yet have a translation could direct to a page that says something like "The {language} translation for this newsletter is not available yet. If you'd like to help translate or review, please contact the optech team/go to the github/etc. In the mean time, here are the previous newsletters that have been translated into {language}".

Any thoughts about that?

bitschmidty:
@harding any thoughts on this approach?

_layouts/newsletter.html
harding: @bitschmidty If you put a dict in `_config.yml` like:

    langs:
      en: English
      es: Spanish
      etc...

You should be able to do something like `"No {{site.langs[lang]}} translation currently available"`

harding (commented):
Concept ACK.  Looked at the code and a local preview.  For @jnewbery comment:

> The greyed out link for languages that don't yet have a translation could direct to a page that says [...]

I think that's a good idea, but I also think that if we add more than one or two more languages, we'll probably want to change the pipe-separated list of language codes into a drop-down selector (possibly with full language names).  It might be worth implementing the drop-down before optimizing other stuff too much.

As another unrelated optimization, I don't think we need to display the link and title for the current language, e.g. "en" on the English page.

PR #616, opened 2021-07-29
Layout: Add auto bullet linking to included files

### LINKED ISSUE #604: Automatic bullet linking doesn't work for included files
See https://github.com/bitcoinops/bitcoinops.github.io/pull/603#discussion_r668161521

It would be *really* nice to make it work.  I'll see what I can do.

harding:
Closes #604 

Apparently it's a known limitation of Jekyll that Liquid template processing doesn't work on `{% include %}` files, which is why the automatic bullet linking isn't working.  This PR adds a `{% auto_anchor %}...{% endauto_anchor %}` block filter that we can put in included files in order to run the anchoring code manually (including calling Liquid).

To test this by diffing the site, I found it necessary to temporarily disable the SEO plugin which outputs a non-deterministic entry in the header of every page.  There's already a bug for that upstream.  (I don't know why this just became a problem now; I've diffed plenty of times before without problems.)  Temporarily disabling the SEO plugin required adding the template from Jekyll Minima to our repository, which is the first commit in this PR.  (To be clear, this PR only adds the ability to disable the plugin, it doesn't actually change anything.)

Diff instructions are:

- Checkout this PR
- Edit `_includes/head.html` and delete the `{%- seo -%}` line
- Build the site: `make clean build`
- Backup the output: `cp -a _site _new_site`
- Drop the two real commits: `git reset --hard HEAD^^`
- Make sure the SEO tag is still deleted
- Build the site again: `make clean build`
- Diff the output: `diff -ruN _site _new_site | colordiff | less -R`

Note: I didn't really check the translations.

### DIFF
--- a/_includes/specials/bech32/23-compat.md
+++ b/_includes/specials/bech32/23-compat.md
@@ -1,4 +1,4 @@
-As of this writing, here are what we think are some of the most
+{% auto_anchor %}As of this writing, here are what we think are some of the most
 significant bech32-related insights we've gleaned from creating and
 reviewing the [Compatibility Matrix][].
 
@@ -40,3 +40,4 @@ reviewing the [Compatibility Matrix][].
   of change output to the type of payment output) but, in most cases,
   this seems like a missed opportunity for wallets to send change to
   their own bech32 addresses for increased fee savings.
+{% endauto_anchor %}
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -1,0 +1,11 @@
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  {%- seo -%}
+  <link rel="stylesheet" href="{{ "/assets/main.css" | relative_url }}">
+  {%- feed_meta -%}
+  {%- if jekyll.environment == 'production' and site.google_analytics -%}
+    {%- include google-analytics.html -%}
+  {%- endif -%}
+</head>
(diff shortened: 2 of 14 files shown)


_includes/specials/bech32/23-compat.md
```diff
@@ -1,4 +1,4 @@
-As of this writing, here are what we think are some of the most
+{% auto_anchor %}As of this writing, here are what we think are some of the most
```
harding, line 1: Note: the lack of newline here is deliberate.  I did a weird thing when adding this to [Newsletter 60](https://bitcoinops.org/en/newsletters/2019/08/21/#bech32-sending-support) and so we can't have a newline here.

harding:
> Note: I didn't really check the translations.

Glad I added that disclaimer since they're all broken.  🤦‍  I'll look into fixing them, probably tomorrow.

jnewbery:
Concept ACK. Will review when the translations are fixed!

harding:
The problem with the translations was the limitations with Kramdown previously described in https://github.com/bitcoinops/bitcoinops.github.io/pull/349#issue-377248256 , so in this PR I just dropped the additional processing of the Japanese files.  If the bullet titles in translations are updated to add some unique latin characters to each bullet, it should be safe to add the `{% auto_anchor %}...{% endauto_anchor %}` back.

jnewbery:
ACK 04c7452025ebb29310e5fd54bc9bcfec73d78256

Do you want the first commit (_Templates: add head.html from jekyll-minima project (MIT license)_) to be merged with this PR, or should it be removed first? It's not clear to me whether it was intended to be just temporary for testing the PR.

harding:
@jnewbery Unless you or anyone else objects, I'd like that first commit merged as it'll make future diffing of the site build contents easier.  However, it isn't needed for this PR and so can be dropped without issue.

jnewbery:
Thanks Dave. No problem including the first commit, so I merged it as is.

bitschmidty:
Post merge ACK. Nice workaround.

PR #572, opened 2021-05-14
Newsletters: add 149 (2021-05-19)

harding:
- [x] Update releases/RCs @harding 
- [x] Add topic links @harding 
- [x] Services/clients section @bitschmidty 
- [x] `Bitcoin Core #21462` guix: Add guix-{attest,verify} scripts @Xekyo 
- [x] `LND #5253` Support paying AMP invoices via SendPaymentV2 @dongcarl

PR #619, opened 2021-08-04
Topics: CPFP remove hyphens

bitschmidty:
Motivation: https://github.com/bitcoinops/bitcoinops.github.io/pull/572#discussion_r635159250 
and https://github.com/bitcoinops/bitcoinops.github.io/blob/master/STYLE.md 

I decided to change the topics and leave newsletters as it seems to be our approach to avoid newsletter updates except for more drastic corrections.


harding:
ACK.

PR #620, opened 2021-08-04
News159 fixup

bitschmidty:
News159 fixup. The missing closing brackets caused a non-linkable bulleted list to be rendered (en and ja affected).


harding:
ACK, thanks!

PR #622, opened 2021-08-05
News160: update MuSig2 description

harding:
As with the other rare occasions where we've made a significant post-publication edit, I'll add a Corrections section to next week's newsletter linking back here.

### DIFF
--- a/_includes/specials/taproot/en/06-multisignatures.md
+++ b/_includes/specials/taproot/en/06-multisignatures.md
@@ -56,11 +56,12 @@ family:
 
 - **MuSig2**, also simple to implement.  It eliminates one round of
   communication and allows another round to be combined with key
-  exchange.  That allows using a somewhat similar signing
+  exchange.  That can allow using a somewhat similar signing
   process to what we use today with script-based multisig.  This does
-  require storing extra data and ensuring your signing software or
+  require storing extra data and being very careful about ensuring your signing software or
   hardware can't be tricked into unknowingly repeating part of the
-  signing session.
+  signing session.  We'll examine the tradeoffs in more detail in next
+  week's *preparing for taproot* column.
 
 - **MuSig-DN** (Deterministic Nonce), significantly more complex to
   implement.  Its communication between participants can't be combined
@@ -76,7 +77,7 @@ utility.  <!-- "[...] there is no reason to prefer MuSig1 over MuSig2
 [...]" -->
 
 There's an open and actively-developed [PR][-zkp 131] to the libsecp256k1-zkp
-project to add MuSig2 support.  We expect the basic multisignature workflow will look
+project to add MuSig2 support.  We expect the basic multisignature workflow for most software will look
 something like the following:
 
 1. The wallet for each participant generates a [BIP32][] xpub that is shared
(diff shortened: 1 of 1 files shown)


jnewbery:
Concept ACK. The relevant discussion is here: https://gnusha.org/secp256k1/2021-08-04.log

Sorry I didn't catch this in review. Maybe we could have asked a libsecp expert for a quick review since this stuff is still moving quite quickly.

harding:
> Maybe we could have asked a libsecp expert for a quick review since this stuff is still moving quite quickly.

Agree, that was my fault for not doing so.

_includes/specials/taproot/en/06-multisignatures.md
```diff
@@ -85,28 +91,30 @@ something like the following:
...
    [PSBT][topic psbt]-based workflow similar to what it would use with
-   script-based multisig.  Unlike multisig, the wallet uses its next
-   nonces and the next nonces of all the other participants to create a
-   shared nonce according to the MuSig2 algorithm; it then creates a
-   partial signature over that nonce and the transaction.
-
-5. When the other wallets receive the PSBT, they use the same procedure.
+   script-based multisig, but now two rounds of communication between
+   signers are required.  In the first round, the proposer creates the
+   unsigned transaction and includes a pair of randomly-generated
```
jonasnick, line 103: nit: imo it's a technical detail whether it's one or two nonces. To the user it should just be a some byte array (of length 66).

_includes/specials/taproot/en/06-multisignatures.md
```diff
@@ -85,28 +91,30 @@ something like the following:
...
+   nonces.  It's important that the nonces not be derived in an entirely
+   deterministic way that could lead to the same nonce being used again
```
jonasnick, line 105: This point could be stressed (it's more than just important). There are ways to rationalize ignoring this advice. I've heard for example: "When I generate the nonces deterministically, I obtain the same signature in the end. So why would I have to add randomness". Well you only get the same signature if the other signers generate their nonces deterministically, but they don't have to.

jnewbery:
> Agree, that was my fault for not doing so.

Nope. I could easily have suggested that during review.

_includes/specials/taproot/en/06-multisignatures.md
```diff
@@ -56,11 +56,17 @@ family:
...
+  exchange.  That can allow using a somewhat similar signing
   process to what we use today with script-based multisig.  This does
-  require storing extra data and ensuring your signing software or
+  require storing extra data and being very careful about ensuring your signing software or
   hardware can't be tricked into unknowingly repeating part of the
-  signing session.
+  signing session.  Proper state management that works even when
+  software and hardware fails is a challenging problem, so it may only be
+  advisible to use the most optimized form of MuSig2 in protocols that
+  already depend on careful state management, such as LN.
```
jonasnick, line 66: This doesn't seem to be quite right. You need to be just as careful with your state management if you don't share the nonces before knowing the message. The only advantage is that you have some defense in depth when you mess up the randomness (or counter). Only MuSig-DN solves the state management issue.

What was a bit problematic in the optech newsletter was the implication that a hardware wallet could share nonces and then any time later sign with these nonces. That's very hard to implement securely compared to using MuSig2 in Lightning because 1) the session state (in particular the nonces) exist only in memory and 2) the memory is assumed to be not tampered with.

_includes/specials/taproot/en/06-multisignatures.md
harding: Gonna leave this in because:

1. I've already seen a case where someone who knew it was two nonces was "corrected" by someone else for only saying "nonce"
2. Maybe someone will wonder "why two nonces" and be motivated to read more about it
3. I find it a nice mnemonic that MuSig**2** is the version with 2 nonces :-)

_includes/specials/taproot/en/06-multisignatures.md
harding: > This doesn't seem to be quite right.

Ok, dropping this whole text in favor of "We'll examine the tradeoffs in more detail in next week's *preparing for taproot* column."  We were scheduled to cover adaptor sigs next week (see #609 ) but I think the discussions from IRC about presharing versus pipelining with musig2 are interesting and useful enough to cover, and there's no better time than right after we first described musig2.

_includes/specials/taproot/en/06-multisignatures.md
jnewbery: Wording changed from "important" to "absolutely essential"

jnewbery:
This looks ready to merge to me. Thanks for the review, @jonasnick!

PR #623, opened 2021-08-08
Newsletter-160:Translate into Japanese with update  MuSig2 description

azuchi:
Based on the update of #622 , the Japanese translation has also been updated.


bitschmidty (approved):
ACK 5b7af30d450de3c8dea7085545c13a3d288aff64

bitschmidty:
Thanks @azuchi

PR #626, opened 2021-08-12
Newsletter-161:Translate into Japanese

azuchi:


bitschmidty (approved):
ACK

Thank you @azuchi !

PR #628, opened 2021-08-13
Newsletters: add 162 (2021-08-18)

harding:
- [x] Update releases & RCs, add topic links @harding 
- [x] Client & Services section @bitschmidty 
- [x] `Bitcoin Core #21800` mempool/validation: mempool ancestor/descendant limits for packages @jnewbery 
- [x] `Rust-Lightning #1009` forward dust limit @dongcarl

### DIFF
--- a/_includes/specials/taproot/en/08-signature-adaptors.md
+++ b/_includes/specials/taproot/en/08-signature-adaptors.md
@@ -1,0 +1,115 @@
+Imagine someone offers to donate 1,000 BTC to a particular charity if
+anyone can guess that person's favorite very large number.  An easy way for the donor
+to do this is to create an unsigned transaction paying the 1,000 BTC and
+then publish an encrypted copy of their signature for the transaction,
+with the favorite number being the decryption key.
+
+In theory, anyone who guesses the number can decrypt the signature and
+then broadcast the transaction, paying the charity.  But if the donor
+uses a standard encryption scheme like AES, there's no easy way for
+third parties to know before decryption that the signature is actually
+valid for that transaction.  Anyone who wants to put effort into
+number guessing has to trust that the donor is sincere and not a
+troll.
+
+Let's extend this problem a bit further.  Third parties Alice
+and Bob want to bet on whether or not the signature is revealed.
+They could perhaps ask the signer for the hash of the signature and use
+that as the hash in an [HTLC][topic htlc] function, but that again
+requires trusting the donor to act honestly.  Even if the signature was
+eventually revealed, the donor could sabotage Alice and Bob's contract
+by giving them an incorrect hash.
+
+### Adaptor magic
+
+[Signature adaptors][topic adaptor signatures], also commonly called
+*adaptor signatures* and *one-time verifiably encrypted signatures*, are
+a solution to these problems---and to many other problems actually faced
+today in production systems built on Bitcoin.  Although
+usable with Bitcoin's existing ECDSA signature scheme, it's much easier
+to use adaptors privately and costlessly in combination with the
+[BIP340][] implementation of [schnorr signatures][topic schnorr
+signatures] for taproot.  Let's see how our example above changes if we
+use adaptors.
+
+As before, the donor prepares a 1,000 BTC transaction.  They sign in
+almost the normal way, with the one difference being that they
+essentially generate their nonce in two parts: a true random nonce that
+they will forever keep secret, and their favorite number---which they'll
+initially keep secret but which is safe for other people to discover.
+The donor generates a fully valid signature using both of these values,
+adding them together as if they were a single nonce.
+
+BIP340 signature commitments use the nonce in two forms: a numeric
+representation (called a *scalar*), which normally only the signer
+knows, and as a *point* on the Elliptic Curve (EC), which is published
+to enable verification.
+
+The donor takes the commitment part of their valid signature and
+subtracts out the hidden scalar.  This makes the signature incomplete
+(and thus invalid) but allows the donor to share the (invalid) signature
+commitment, the (valid) point for the complete nonce, and the (valid)
+point for the hidden number.  Together these three pieces of information
+are a *signature adaptor*.
+
+Using a slight variant on the BIP340 signature verification algorithm,
+anyone can verify that the signature adaptor would provide a valid signature
+if the hidden scalar was simply added back in to the (currently invalid)
+signature commitment.  This is possible to verify even without knowing
+what that hidden number is.  In short, it is now possible for users to
+trustlessly begin making guesses about the value of hidden scalar,
+secure in the knowledge that a correct guess will allow them to get the
+signature and send the transaction.
+
+Like everyone else who received the donor's signature adaptor, Alice and
+Bob now have a copy of the EC point for the hidden number.  Also like
+everyone else, they don't know the actual scalar.  But, if you recall,
+all the donor did to turn their valid signature into an invalid
+signature is subtract out the hidden number from their signature
+commitment while continuing to have the signature commit to the point
+of the hidden number.  Alice can just as easily create an invalid
+signature by not committing to the scalar she doesn't know but still
+committing to the EC point she does know.  She does this by creating her
+own nonce pair, using the private form when creating her (invalid)
+signature but commiting to the aggregation of the public form of her
+nonce and the EC point from the donor's signature adaptor.
+This produces a
+signature adaptor for a transaction that pays Bob.  If Bob learns the
+scalar, he can convert that adaptor into a valid signature and send the
+transaction, winning the bet.
+
+But how does Bob learn the winning number?  Does he have to wait for
+someone else who guesses it to publish a press release?  Nope.  Recall
+one more time that the signature adaptor the donor published was their
+actual signature minus the scalar.  When the hidden number is discovered
+and somebody sends the 1,000 BTC transaction, they must publish the
+original (valid) signature commitment.  Bob can take that (valid)
+signature commitment and subtract from it the (invalid) signature
+commitment in the original signature adaptor to get the scalar.  He then
+uses that scalar to convert Alice's adaptor into a valid signature.
+
+### Multisignature adaptors
+
+The previous section shows individual users modifying how they
+create their signatures to produce signature adaptors.  It's also
+possible for the parties to a [multisignature][topic multisignature] to
+use the same trick.  That's extraordinarily useful, as many cases where
+signature adaptors will be used require the cooperation of two users.
+
+For example, when Alice and Bob make the bet above, they might start by
+depositing funds into a script only spendable by a multisignature
+between them.  Then Alice can produce her partial signature in the form
+of a signature adaptor; if Bob learns the hidden number, he can
+transform Alice's adaptor into her valid partial signature and then
+provide his partial signature to produce a full signature spending the
+money.
+
+This gives signature adaptors all the same advantages of multisignatures
+in general: they look like and use the same amount of space as a single
+signature, minimizing fees and maximizing privacy and fungibility.
+
+In next week's *preparing for taproot* column, we'll explore one of the
+main ways we expect to see signature adaptors used: Point Time Locked
+Contracts ([PTLCs][topic ptlc]), an upgrade for the venerable Hash Time
+Lock Contracts ([HTLCs][topic htlc]) used extensively in LN, coinswaps,
+and a number of other protocols.
--- a/_posts/en/newsletters/2021-08-18-newsletter.md
+++ b/_posts/en/newsletters/2021-08-18-newsletter.md
@@ -1,0 +1,110 @@
+---
+title: 'Bitcoin Optech Newsletter #162'
+permalink: /en/newsletters/2021/08/18/
+name: 2021-08-18-newsletter
+slug: 2021-08-18-newsletter
+type: newsletter
+layout: newsletter
+lang: en
+---
+This week's newsletter summarizes a discussion about the dust limit and
+includes our regular sections with descriptions of changes to services
+and client software, how you can prepare for taproot, new releases and
+release candidates, and notable changes to popular Bitcoin
+infrastructure software.
+
+## News
+
+- **Dust limit discussion:** Bitcoin Core and other node software
+  refuses by default to relay or mine any transaction with an output
+  value below a certain amount, the [dust limit][topic uneconomical
+  outputs] (the exact amount varies by output type).  This makes it more
+  difficult for users to create *uneconomical outputs*---UTXOs that
+  would cost more in fees to spend that they hold in value.
+
+    This week, Jeremy Rubin [posted][rubin dust] to the Bitcoin-Dev
+    mailing list a five-point argument for removing the dust limit and
+    stated a belief that the reason for the limit is to prevent "spam"
+    and "[dust fingerprint attacks][topic output linking]".  Others
+    [replied][corallo dust] with [counterarguments][harding dust] and
+    noted that the limit exists not to prevent spam but to prevent
+    users from permanently wasting the resources of full node operators
+    by creating UTXOs that the users will have no financial incentive to
+    ever spend.  Parts of the discussion also [described][riard dust]
+    the [impact][towns dust] of both the dust limit and uneconomical
+    outputs on parts of LN.
+
+    As of this writing, it did not appear any agreement was likely to be
+    reached.  At least for the short term, we expect the dust limit to
+    remain.
+
+## Changes to services and client software
+
+*In this monthly feature, we highlight interesting updates to Bitcoin
+wallets and services.*
+
+FIXME:bitschmidty
+
+## Preparing for taproot #9: signature adaptors
+
+*A weekly [series][series preparing for taproot] about how developers
+and service providers can prepare for the upcoming activation of taproot
+at block height {{site.trb}}.*
+
+{% include specials/taproot/en/08-signature-adaptors.md %}
+
+## Releases and release candidates
+
+*New releases and release candidates for popular Bitcoin infrastructure
+projects.  Please consider upgrading to new releases or helping to test
+release candidates.*
+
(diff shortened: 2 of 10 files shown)


harding:
@jeremyrubin This newsletter briefly summarizes the dust limit discussion from the mailing list.  I have strong feelings about that, so it's unfair of me to be writing summaries about conflicting opinions.  Would you be able to take a moment to review and let me know if I'm misrepresenting your position, omitting important information, or otherwise being injudicious?  There's a preview here: https://deploy-preview-628--bitcoinops.netlify.app/

JeremyRubin:
Add sentences saying something like?

"When adopted, technologies like Utreexo would make it very cheap for the network to store outputs unlikely to be redeemed and so the network would likely be able to remove this restriction. 

"The maximum growth rate of UTXO creation is 144*365*1mb = 50MB (assuming every block in a year is all outputs, which is a bad assumption)."

harding:
@jeremyrubin

I don't believe the Utreexo claim is as rosy as that (and I previously disputed that on list: https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2021-August/019321.html ).  I personally don't see a reason to ever eliminate the limit unless we truly need to drop it for some desired technological innovation (e.g. CT if CT really required it, which it doesn't).  If people plan to spend a UTXO, they can fill it with enough value to pay its fee at spend time.  If they don't plan to spend it, then they don't need a UTXO.

How about:  "It was [argued][] that, if every Bitcoin full node began using Utreexo, it would make the cost of storing the UTXO set  constant regardless of its size; but others [countered][] that some people (such as miners) will still want to store the whole UTXO set and access it quickly, and that larger UTXO set sizes will still increase the bandwidth requirements on all full nodes."

> "The maximum growth rate of UTXO creation is 144_365_1mb = 50MB (assuming every block in a year is all outputs, which is a bad assumption)."

You mean 50 GB?  I guess I don't mind printing that, but (1) more than 10xing the current UTXO set size in a year doesn't sound like a persuasive argument and (2) was this worst case discussed in the mailing list?

JeremyRubin:
It's very possible for utreexo to bin dust-y UTXOs together if they're unlikely to be spent. In fact this might be a good integration plan for utreexo -- only put into it at first things you think are not worth spending @adiabat. Utreexo is indeed log(n) but you can have a cache  of say the last year of unspent coins at any given time. I also don't really know what n is specifically in utreexo -- depending on the exact design, N could be just 25,000 (~max utxos per block) if each block has it's own utreexo hash root.

Further, if you want to have a whole index of coins accessible, maybe you can run on better hardware specs. Perhaps there's an issue that miners currently can put the whole thing in RAM but at 100's GBs may not be able to on a given machine. This is a good concern; but I think with NVMEs it wouldn't be the bottleneck.

50GB (i edited immediately after posting, you must be quoting from email) is AFAICT the worst case. I think we should support it because even short of dust because it's what's required to keep a node in consensus. Further, that's about 1.3B UTXOs total, or about 1/10th the population receiving a coin in a given year that they don't spend quickly.

This is also all presupposing that there's actually demand to create these. I don't think there would be much, and if there is, then it's good for Bitcoin because that will overall raise fees as more users use the chain or other use cases would outcompete. And if there is more fee demand to create dust outputs than anything else, maybe the market has just discovered a better use for Bitcoin than we have now (I doubt it!.

I think what the main issue is that people use the UTXO set as an DHT and expect some sort of free retreival from everyone else. This is already the case to the extent anyone would serve you a block, so it's not really different (if you don't have a block, then you can't verify a utxo anyways w/o utxo commitments).

harding:
> It's very possible for utreexo to bin dust-y UTXOs together if they're unlikely to be spent.

Sure, but any such optimization on the normal case increases costs in the abnormal case.  And if UTXOs are unlikely to be spent then they're unlikely to be needed---since the only reason to create a UTXO is to spend it---so why should we complicate the design of systems like Utreexo to give dust special treatment?  A far simpler solution is to continue with the rule we have now: if you want a UTXO, demonstrate your intention to spend it in the future by funding it with enough value to (hopefully) cover the fee to spend it.

> (i edited immediately after posting, you must be quoting from email)

Opening the PR and commenting on it was the last thing I did before going to bed.  Your comment was displayed when I woke up; I guess GitHub displays new comments but doesn't update the display when they're edited (which seems sane to me).

> 50GB is AFAICT the worst case

I think that's about right in the worst case with Bitcoin Core's current design.  However, it's worth noting that the amount of data currently required to store a UTXO is greater than the amount of data it adds to a transaction because it needs to stored with its outpoint and block height in addition to its nValue and scriptPubKey.  IIRC, leveldb allows bitcoind to prefix compress identical bytes in different UTXOs, so the worst case block would only have one txid and one block height, so the only differences between UTXOs would be the vout, nvalue, and scriptpubkey---which is pretty close to the actual transaction data.

In other words, creating a million dust UTXOs hurts full nodes but creating a single dust UTXO hurts worse, proportionately.

> 50GB [UTXO set size] we should support it [...] that's about 1.3B UTXOs total, or about 1/10th the population receiving a coin in a given year that they don't spend quickly.

That's an interesting argument, but it also seems like a different argument to me.

> This is also all presupposing that there's actually demand to create these. I don't think there would be much, and if there is, then it's good for Bitcoin because that will overall raise fees as more users use the chain or other use cases would outcompete.

It's good for Bitcoin for increased fees to increase PoW, but it's not good for Bitcoin for running a full node become more burdensome.

> And if there is more fee demand to create dust outputs than anything else, maybe the market has just discovered a better use for Bitcoin than we have now (I doubt it!).

That doesn't even make sense.  How can you create dust without paying bitcoin in fees, and if you're paying bitcoin in fees, you (and the miner receiving those fees) are using Bitcoin as a payment system (and probably also a store of value), which is how we have it now.

> I think what the main issue is that people use the UTXO set as an DHT and expect some sort of free retreival from everyone else.

I don't understand what you're trying to say here.

---

To the main point, unless you object, I'll add something to the article about Utreexo similar to the text I suggested above.  I'd prefer to leave the 50GB bit out, but if you really think it's important, please say so and I'll do my best to work it in.

jnewbery:
LGTM so far. I need to review the signature adaptors section again tomorrow, and suggest we also ask one of our friends in the #secp256k1 channel to look over it.

> To the main point, unless you object, I'll add something to the article about Utreexo similar to the text I suggested above. I'd prefer to leave the 50GB bit out, but if you really think it's important, please say so and I'll do my best to work it in.

> How about: "It was [argued][] that, if every Bitcoin full node began using Utreexo, it would make the cost of storing the UTXO set constant regardless of its size; but others [countered][] that some people (such as miners) will still want to store the whole UTXO set and access it quickly, and that larger UTXO set sizes will still increase the bandwidth requirements on all full nodes."

@harding - I think the write-up of the dust discussion is absolutely fine as it is currently, and wouldn't be improved by including that text. We shouldn't aim to include every point and counter-point in a discussion thread, and the text that you've already written gives a fair summary of what was discussed. Anyone who wants more detail can refer to the mailing list thread.

_includes/specials/taproot/en/08-signature-adaptors.md
```diff
@@ -0,0 +1,111 @@
...
+transaction, winning the bet.
+
+But how does Bob learn the winning number?  Does he have wait for
+someone else who guesses it to publish a press release?  Nope.  Recall
+one more time that the signature adaptor the donor published was their
+actual signature minus the scalar.  When the hidden number is discovered
+and somebody sends the 1,000 BTC transaction, they must publish the
+original (valid) signature commitment.  Bob can take that (valid)
+signature commitment and subtract from it the (invalid) signature
+commitment in the original signature adaptor to get the scalar.  He then
```
Xekyo, line 84: The donor encumbered the signature by deducting the hidden scalar "hs": `sig - hs = invalid_sig`
`invalid_sig - sig = (sig - hs) -sig = -hs`

=> Shouldn't this be "valid signature - invalid signature": `sig - invalid_sig = sig - (sig - hs) = hs`?

_posts/en/newsletters/2021-08-18-newsletter.md
```diff
@@ -0,0 +1,121 @@
...
+repo], [Hardware Wallet Interface (HWI)][hwi repo],
+[Rust Bitcoin][rust bitcoin repo], [BTCPay Server][btcpay server repo],
+[Bitcoin Improvement Proposals (BIPs)][bips repo], and [Lightning
+BOLTs][bolts repo].*
+
+- [Bitcoin Core #22642][] updates Bitcoin Core's release process for
+  upcoming version 22.0 to concatenate the GPG signatures of everyone
+  who [reproducibly built][topic reproducible builds] the binaries into a single file which
+  can be batch verified ([example][gpg batch]).  Signatures from
+  determinsitic builders have been available for years, but this should
```
Xekyo, line 83:
```suggestion
  deterministic builders have been available for years, but this should
```

_topics/en/uneconomical-outputs.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Optional.  Force the display (true) or non-display (false) of stub
+## topic notice.  Default is to display if the page.content is below a
+## threshold word count
+#stub: false
+
+## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
+## Should be less than 500 characters
+excerpt: >
+  **Uneconomical outputs** are transaction that are worth less
```
Xekyo, line 49: This sentence seems to miss a word

```suggestion
  **Uneconomical outputs** are transaction outputs that are worth less
```

_posts/en/newsletters/2021-08-18-newsletter.md
```diff
@@ -0,0 +1,121 @@
...
+  configuration option which limits the total balance of pending "dusty HTLCs"
+  whose amounts are below the dust limit.
+
+  This change is in preparation for a [proposed][BOLTs #873]
+  `option_dusty_htlcs_uncounted` feature bit, which advertises that the node
+  does not wish to count "dusty HTLCs" against `max_accepted_htlcs`. Node
+  operators would likely want to adopt this feature bit since
+  `max_accepted_htlcs` is mainly used to limit the potential size of the onchain
+  transaction if a force-close were to happen and "dusty HTLCs" are unclaimable
+  onchain and would never affect the final transction size.
```
adamjonas, line 105:
```suggestion
  onchain and would never affect the final transaction size.
```

kcalvinalvin:
> larger UTXO set sizes will still increase the bandwidth requirements on all full nodes."

In general yes but since Utreexo is a collection of many merkle trees, utxos that just stick around forever and just increase the UTXO set doesn't have that much effect on the proof size (thus not having that much effect on bandwidth).

```
28
|---------------\
24              25              26
|-------\       |-------\       |-------\
16      17      18      19      20      21      22
|---\   |---\   |---\   |---\   |---\   |---\   |---\
00  01  02  03  04  05  06  07  08  09  10  11  12  13  14
```

In the above example, if UTXO numbered 08 is to be spent, the proof of existence for that would be:

```

|-------\
        21
|---\   |---\
08  09
```

UTXOs `00` - `07` had no effect on the proof size.

> How about: "It was [argued][] that, if every Bitcoin full node began using Utreexo, it would make the cost of storing the UTXO set constant regardless of its size; but others [countered][] that some people (such as miners) will still want to store the whole UTXO set and access it quickly, and that larger UTXO set sizes will still increase the bandwidth requirements on all full nodes."

I think just mentioning that there's an existing proposal out there (that may be of help) would be beneficial for people reading the summary. But of course I'm biased so I'd be happy if the Utreexo project gets a mention with it's potential benefits listed.

Also, some might have a problem claiming that Utreexo is constant. Specifically saying O(log n) would be better.

harding:
@kcalvinalvin 

> Also, some might have a problem claiming that Utreexo is constant. Specifically saying O(log n) would be better.

Can't utreexo storage be reduced down to a single merkle root?  Obviously there's a tradeoff between storage and bandwidth (and other things), but I thought the absolute best case on storage for a utreexo-using node would be just a merkle root.

kcalvinalvin:
> Can't utreexo storage be reduced down to a single merkle root?

Depends. If the number of UTXOs in existence is a power of 2, then yes. Otherwise there's gonna be multiple merkle roots.

Side note but an eeasy way of calculating this is looking at the binary representation of the UTXO count. If there's exactly 100 UTXOs in existence, there would be 3 merkle roots since there's 3 bits turned on: 1100100. If there's 128 UTXOs in existence then yes there would only be a single merkle root since: 10000000.

So like in absolute best case yes but it's rare. Tadge did mention that he scanned through the mainnet data a while ago and he couldn't find any blocks where there was only a single root.

_includes/specials/taproot/en/08-signature-adaptors.md
```diff
@@ -0,0 +1,111 @@
...
+
+Like everyone else who received the donor's signature adaptor, Alice and
+Bob now have a copy of the EC point for the hidden number.  Also like
+everyone else, they don't know the actual scalar.  But, if you recall,
+all the donor did to turn their valid signature into an invalid
+signature is subtract out the hidden number from their signature
+commitment while continuing to have the signature commit to the point
+of the hidden number.  Alice can just as easily create an invalid
+signature by not committing to the scalar she doesn't know but still
+committing to the EC point she does know.  This allows Alice to create a
```
jnewbery, line 72: I find "not committing to the scalar she doesn't know but still committing to the EC point she does know" just a little difficult to understand. What does it mean that Alice's signature is _not committing to the scalar she doesn't know_?

jnewbery (commented):
Reread everything and still looks good to me!

_posts/en/newsletters/2021-08-18-newsletter.md
```diff
@@ -0,0 +1,140 @@
...
+  on the project's lead maintainer signing the release binaries.
+
+- [Bitcoin Core #21800][] implements ancestor and descendant limits for
+  mempool package acceptance. Bitcoin Core limits the number of
+  related transactions in its mempool as a protection against DoS
+  attacks and so that block construction is tractable for miners. By default,
+  those [limits][bitcoin core mempool limits] ensure that no transaction in
+  the mempool can have  more than 25 ancestors or descendants, and that no
+  transaction can have ancestors or descendants that have a combined weight of
+  101kVb.
```
glozow, line 94: The limits are inclusive of the tx itself. Also, it's kind of hard to word this in a clear way, but I hope to articulate that there are a total of 4 limits:

```suggestion
- [Bitcoin Core #21800][] implements ancestor and descendant limits for
  mempool package acceptance. Bitcoin Core limits the number of
  related transactions in its mempool as a protection against DoS
  attacks and so that block construction is tractable for miners. By default,
  those [limits][bitcoin core mempool limits] ensure that no transaction
  in the mempool, combined with its mempool ancestors, can exceed 25
  transactions or 101KvB in weight. The same rules apply to the transaction
  combined with its mempool descendants.
```

_posts/en/newsletters/2021-08-18-newsletter.md
```diff
@@ -0,0 +1,140 @@
...
+
+    Those ancestor and descendant limits are enforced when a transaction is
+    considered for addition to the mempool. If adding the transaction would
+    cause one of those limits to be exceeded, then the transaction is rejected.
+    [#21800][bitcoin core #21800] adds ancestor and descendant limit checks for
+    mempool package acceptance (i.e. when multiple transactions are considered
+    for addition to the mempool at the same time). Mempool package acceptance
```
glozow, line 101: The main thing about #21800 is it's safe for any arbitrary package. I'll optimize it in future PRs when stuff is more specific (also I think harding mentioned something about looking for feedback on package limits?) - I'm pretty sure it's going to be 2 generation, multi parent + 1 child, but probably good idea to say it's not finalized and maybe interested people will know where to make suggestions?

```suggestion
    While package semantics have not been finalized, [#21800][bitcoin core #21800]
    implements ancestor and descendant limit checks for validating
    arbitrary packages (i.e. when multiple transactions are considered
    for addition to the mempool at the same time). Mempool package acceptance
```

glozow (commented):
Thanks for 21800 description @jnewbery!

jnewbery:
Thanks @glozow. I've taken both of your suggestions.

_includes/specials/taproot/en/08-signature-adaptors.md
harding: I think that's what it says?  "take ... valid ... and subtract from it ... invalid"

_posts/en/newsletters/2021-08-18-newsletter.md
dongcarl: Fixed!

_includes/specials/taproot/en/08-signature-adaptors.md
Xekyo: Oh, oops. Sorry, I guess I must have misread that as "take the valid signature commitment and subtract it from the invalid", i.e. switched the order reading of "from" and "it" when reading it.

harding: Yeah, the writing here is a bit stilted.  In English, we usually write "subtract b from a" but I've always found that confusing since we write "a - b".  I guess by trying to normalize things, I may have made it more confusing.  I think I'll leave it this way, though.  (Maybe next time I should try phrasing it in RPN: "take the invalid_sig and the valid_sig and perform subtraction"  :-)

Xekyo:
Re-reviewed, LGTM, no additional comments at this time.

harding:
Made all suggested edits (I hope), added topics, and added Bitcoin Core 0.21.2rc1.  I also reviewed the client/services section by @bitschmidty and PR descriptions by @dongcarl and @jnewbery, they all look good, thanks!  Special thanks @dongcarl for really digging into that PR and the BOLT update behind it, and to @glozow for the review!

jnewbery:
Re-reviewed. ACK. Thanks all!

bitschmidty:
Squashed and merged. 

Thank you to @harding @dongcarl @jnewbery @glozow @adamjonas @kcalvinalvin @JeremyRubin @Xekyo 👏

PR #629, opened 2021-08-19
Newsletter-162:Translate into Japanese

azuchi:


bitschmidty (approved):
ACK

Thank you @azuchi !

PR #631, opened 2021-08-20
Correct symbol for kibibyte

Xekyo:
The symbol for kibi is "Ki" with a capital k. Since the International
System of Units restricts the use of SI prefixes strictly to powers of
10, the description of that line is updated to not mention SI in the
context.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -53,7 +53,7 @@ following abbreviations we assume readers will already know.
 - DoS (Denial-of-service)
 - ECDSA (Elliptic Curve Digital Signature Algorithm)
 - kB, MB, GB, TB, etc... (SI-prefixed byte sizes)
-- kiB, MiB, GiB, TiB, etc... (SI-prefixed binary byte sizes)
+- KiB, MiB, GiB, TiB, etc... (Prefixes denoting byte sizes in powers of 1024)
 - LN (Lightning Network)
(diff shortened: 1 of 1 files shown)


harding:
ACK.  Thanks!

For anyone else who is confused as I am, `k` is the generic SI prefix for kilo but they decided to use `K` for kibi 'cause reasons, or something:

> (For consistency with the other prefixes for binary multiples, the symbol Ki is used for 2^10 rather than ki.)

[source](https://physics.nist.gov/cuu/Units/binary.html)

Xekyo:
I grepped the repository and could not find a single instance of `kiB` beyond the one in STYLE.md.

jnewbery:
Thanks @Xekyo. TIL!

Xekyo:
> For anyone else who is confused as I am, `k` is the generic SI prefix for kilo but they decided to use `K` for kibi 'cause reasons, or something:

I think the general pattern for metric prefixes was supposed to be majuscules for prefixes denominating greater powers than 1000^0 (k, M, G, T, P, E, Z, Y) and minuscules for prefixes denominating fractions (m, μ, n, p, f, a, z, y). From what I gather, kilo got the symbol "k" because the "K" was already the symbol for Kelvin, one of the seven base units. That collision doesn't exist in that form for kibi (Ki), so they could stay consistent. :)

PR #642, opened 2021-09-09
Newsletter-165:Translate into Japanese

azuchi:


bitschmidty (approved):
tACK

Thanks @azuchi !

PR #652, opened 2021-09-22
Newsletter-167:Translate into Japanese

azuchi:


bitschmidty (approved):
Very fast, thanks @azuchi !

PR #656, opened 2021-09-27
Remove links to https://github.com/bitcoinops/scaling-book/blob/master/x.payment_batching/payment_batching.md

jnewbery:
See https://github.com/bitcoinops/scaling-book/pull/30


harding:
Untested ACK

PR #657, opened 2021-09-29
News168: update Eclair#1942 description

harding:
Correction suggested by @thomash-acinq (thanks!)


jnewbery:
Thanks @harding and @thomash-acinq!

PR #659, opened 2021-09-30
Newsletter-168:Translate into Japanese

azuchi:


bitschmidty (approved):
Looks good, thank you @azuchi !

PR #667, opened 2021-10-18
Newsletters: add 171 (2021-10-20)

harding:


_posts/en/newsletters/2021-10-20-newsletter.md
```diff
@@ -0,0 +1,205 @@
...
+seems worth publicizing the decision to use this value -->
+- [Bitcoin Core #22863][] documents a decision to use the same minimum
+  output amount ("dust amount") of 294 sat for P2TR outputs as for
+  P2WPKH outputs.  This is done even though it costs less to spend a
+  P2TR output than a P2WPKH output because some developers were
+  [opposed][bitcoin core #22779] to lowering the dust limit at this
+  time.
+
+- [Bitcoin Core #23093][] adds a new `newkeypool` RPC that will mark all
+  pre-generated addresses in the wallet as used so that a new set addresses
```
adamjonas, line 155:
```suggestion
  pre-generated addresses in the wallet as used so that a new set of addresses
```

_includes/specials/taproot/en/18-trivia.md
```diff
@@ -0,0 +1,83 @@
+{% auto_anchor %}
+
+- **What is a taproot?** Wikipedia [says][wikipedia taproot], "A taproot
+  is a large, central, and dominant root from which other roots sprout
+  laterally. Typically a taproot is somewhat straight and very thick, is
+  tapering in shape, and grows directly downward.  In some plants, such
+  as the carrot, the taproot is a storage organ so well developed that
+  it has been cultivated as a vegetable."
+
+    Bitcoin's [taproot][topic taproot], similar to the plant root, hides
```
sipa, line 10: I always assumed the origin of the name was "taps into the Merkle root", but I don't actually know what @gmaxwell's thinking was.

_includes/specials/taproot/en/18-trivia.md
```diff
@@ -0,0 +1,83 @@
...
+  Pieter Wuille writes that the idea was developed by Andrew Polestra
+  and Gregory Maxwell "while I briefly left the table... !$%@" [sic].
+  <!-- personal correspondence with Wuille -harding -->
+
+- **<!--x-->2.5 years in 1.5 days:** choosing the optimal constant for
+  [bech32m][topic bech32] required [about][wuille matrix elimination]
+  2.5 years of CPU time, which was performed in just 1.5 days mostly
+  using a CPU cluster belonging to Gregory Maxwell.
+
+*We thank Gregory Maxwell and Pieter Wuille for enjoyable conversations
```
sipa, line 69: Maybe @ajtowns, @jonasnick or @real-or-random have some more trivia about the history of the proposal?

_includes/specials/taproot/en/18-trivia.md
```diff
@@ -0,0 +1,83 @@
...
+  schnorr signatures] as an upgrade on Bitcoin's original ECDSA
+  signatures because they make it easier to implement various
+  cryptographic tricks, but the schnorr signature algorithm predates the
+  DSA algorithm that ECDSA is based upon.  Indeed, DSA was created in
+  part to circumvent the [patent on schnorr signatures][schnorr patent]
+  and it was [reported][gov report] that "there were allegations of
+  infringement from Mr. Claus Schnorr and Public Key Partners (PKP)"
+  that DSA violated Schnorr's patent.  That patent expired in 2008.
+
+- **Un-Schnorr what name to use:** although it didn't work out, there
```
sipa, line 24: In  fact, I think we did consider calling BIP340 "DLS" for "Discrete Logarithm Signatures", but I think we eventually reneged on that because the name Schnorr was already so much talked about.

_includes/specials/taproot/en/18-trivia.md
ajtowns: I originally had to look the word up; but I took it as the key path being the "taproot" because that's the tasty one that you make carrot soup out of, and the merklized scripts would be the other lesser roots that you hope to ignore. The BIP-116/117 approach didn't have a "root" to tap into per se, and the original proposal was pretty explicitly compatible with that I think.

_includes/specials/taproot/en/18-trivia.md
```diff
@@ -0,0 +1,83 @@
...
+    but, instead of committing to the terms of an offchain contract, the
+    output creator commits to consensus-enforced terms chosen by the
+    receiver for how they can spend the received bitcoins onchain.
+
+- **A Good Morning:** the idea to use P2C so that payments to scripts
+  can look identical onchain to paying public keys was invented in Los
+  Altos, California, at the diner "A Good Morning" on 22 January 2018.
+  Pieter Wuille writes that the idea was developed by Andrew Polestra
+  and Gregory Maxwell "while I briefly left the table... !$%@" [sic].
+  <!-- personal correspondence with Wuille -harding -->
```
ajtowns, line 62: Oh, I get it now! When we were discussing whether to keep calling it "metas" (MErkle/TAproot/Schnorr afaik) or pick an alternative name ("segwit v1", "taproot"), `codename "good morning" ?` was a suggestion... Would've been amusing with the current "ngmi" "gm!" slang...

_posts/en/newsletters/2021-10-20-newsletter.md
```diff
@@ -0,0 +1,205 @@
...
+  mobile] on the Lightning-Dev mailing list for brainstorming how
+  frequently-offline LN nodes (such as those on mobile devices) can
+  receive payments without requiring a custodial solution or locking up
+  channel liquidity for prolonged periods of time.  Corallo believes
+  there's a reasonably good solution involving an untrusted third
+  party once LN is [upgraded][zmn taproot] to support [PTLCs][topic
+  ptlc], but he's also seeking suggestions for any alternative solutions
+  that could be deployed even before PTLC support is added.
+
+- **FIXME:harding needs a better title:** a few weeks apart, developers
```
ajtowns, line 24: "A probing question", "Probe-able flaws", "Towards more probable probe-ability" ?

_includes/specials/taproot/en/18-trivia.md
```diff
@@ -0,0 +1,83 @@
...
+  laterally. Typically a taproot is somewhat straight and very thick, is
+  tapering in shape, and grows directly downward.  In some plants, such
+  as the carrot, the taproot is a storage organ so well developed that
+  it has been cultivated as a vegetable."
+
+    Bitcoin's [taproot][topic taproot], similar to the plant root, hides
+    the details of its scripts below the surface, away from public view,
+    and is very efficient at storing those scripts.
+
+- **Schnorr predates ECDSA:** we talk about [schnorr signatures][topic
```
real-or-random, line 14: nit (also a few times below): I think I point this out at least once per month somewhere but I prefer to call this "Schnorr signatures" (or at least "Schnorr sigs") in formal writing. "Schnorr" is an alive human being and not a thing, so I think using his name without any suffixes for a signature scheme is a little bit disrepectful, and everyone deserves respect (even though he patented the scheme. ;)).

_includes/specials/taproot/en/18-trivia.md
```diff
@@ -0,0 +1,83 @@
...
+  schnorr signatures] as an upgrade on Bitcoin's original ECDSA
+  signatures because they make it easier to implement various
+  cryptographic tricks, but the schnorr signature algorithm predates the
+  DSA algorithm that ECDSA is based upon.  Indeed, DSA was created in
+  part to circumvent the [patent on schnorr signatures][schnorr patent]
+  and it was [reported][gov report] that "there were allegations of
+  infringement from Mr. Claus Schnorr and Public Key Partners (PKP)"
+  that DSA violated Schnorr's patent.  That patent expired in 2008.
```
real-or-random, line 22: Just for my own curiosity: Everyone claims it's 2008 but google patents says 2010. I believe 2008 is right but is there a definitive source?

_includes/specials/taproot/en/18-trivia.md
```diff
@@ -0,0 +1,83 @@
...
+
+- **Un-Schnorr what name to use:** although it didn't work out, there
+  was a [suggestion][dryja bn sigs] early in the development of adapting
+  schnorr signatures for Bitcoin that Claus Schnorr's name shouldn't be
+  used in association with them because his patent on them prevented the
+  widespread use of a valuable cryptographic technique for over 20
+  years.
```
real-or-random, line 29: more a comment, no need to change anything: I don't know him personally but in his defend, those were just different times. I think nowadays the idea that patents are not a good idea is much more wide-spread in the applied crypto community. Even Phil Rogaway who wanted to use them for good (no military use etc.) apparently has moved away from the idea https://mailarchive.ietf.org/arch/msg/cfrg/qLTveWOdTJcLn4HP3ev-vrj05Vg/

_includes/specials/taproot/en/18-trivia.md
real-or-random: In light of my comment above, "Un-Schnorr" is another term I would personally avoid.

_includes/specials/taproot/en/18-trivia.md
```diff
@@ -0,0 +1,83 @@
...
+
+- **Schnorr for twisted Edwards curves:** an application of schnorr
+  signatures using elliptic curves was written in 2011 <!-- guessing
+  from URL; Wikipedia also says this without a source --> and
```
real-or-random, line 33: the paper dates back to 2011, see https://ed25519.cr.yp.to/papers.html

_includes/specials/taproot/en/18-trivia.md
real-or-random: If the diner will be famous for this in 30 years, we either did everything right or very wrong.

_includes/specials/taproot/en/18-trivia.md
```diff
@@ -0,0 +1,83 @@
...
+  [published][ed25519] in 2012<!-- verified -->.  The scheme,
+  [EdDSA][], is now the basis of several standards.  Although not used
+  in Bitcoin consensus, references to it in the context of other systems
+  can be found in many of the Bitcoin repositories tracked by Optech.
```
real-or-random, line 37: This could say that EdDSA influenced the design of BIP340 heavily. Without this, it's also not entirely clear to me what the purpose of the paragraph is.

_posts/en/newsletters/2021-10-20-newsletter.md
```diff
@@ -0,0 +1,205 @@
...
+    probe a payment path.  For example, Alice generates an [HTLC][topic
+    htlc] that pays a preimage nobody knows.  She routes the payment
+    through Bob and Charlie to Zed.  Since Zed doesn't know the payment
+    preimage, he's forced to reject the payment even though it indicates
+    he's supposed to be the ultimate receiver.  If Alice receives the
+    reject message from Zed's node, she knows that Bob and Charlie had
+    enough funds in their channels to allow paying Zed, and so she can
+    immediately react to Zed's rejection by sending an actual payment
+    that has a high probability of succeeding (it would only fail if Bob's
+    or Charlie's balances had changed in the meantime).  The advantage
```
jnewbery, line 43: This isn't the _only_ reason that the payment could fail. For example, Bob or Charlie could go offline, or change their forwarding policy, or ...

_posts/en/newsletters/2021-10-20-newsletter.md
```diff
@@ -0,0 +1,205 @@
...
+  Zeus, a mobile Bitcoin and Lightning wallet application, in its
+  [v0.6.0-alpha3][zeus v0.6.0-alpha3] release, provides additional support for [atomic
+  multipath payments (AMPs)][topic amp], [Lightning Addresses][news167 lightning
+  addresses], and [coin control][topic coin selection] features.
+
+- **Sparrow adds coinjoin support:**
+  [Sparrow 1.5.0][] adds [coinjoin][topic coinjoin] features by integrating with
+  Samourai's [Whirlpool][whirlpool].
+
+- **JoinMarket 0.9.0 adds RBF support:**
```
jnewbery, line 96:
```suggestion
- **JoinMarket 0.9.2 adds RBF support:**
```

_posts/en/newsletters/2021-10-20-newsletter.md
```diff
@@ -0,0 +1,205 @@
...
+<!-- we wouldn't normally cover a small code comment like this, but it
+seems worth publicizing the decision to use this value -->
+- [Bitcoin Core #22863][] documents a decision to use the same minimum
+  output amount ("dust amount") of 294 sat for P2TR outputs as for
+  P2WPKH outputs.  This is done even though it costs less to spend a
+  P2TR output than a P2WPKH output because some developers were
+  [opposed][bitcoin core #22779] to lowering the dust limit at this
+  time.
+
+- [Bitcoin Core #23093][] adds a new `newkeypool` RPC that will mark all
```
jnewbery, line 154: Perhaps:

```suggestion
- [Bitcoin Core #23093][] adds the `newkeypool` RPC method that will mark all
```

to remove the slightly awkward new new

_posts/en/newsletters/2021-10-20-newsletter.md
```diff
@@ -0,0 +1,205 @@
...
+  cve] for more information.
+
+- [Rust-Lightning #1124][] extends the `get_route` API with the ability
+  to pass an additional parameter that can be used to avoid reusing
+  previously failed routes.  Additional planned changes will make it
+  easier to use previous routing successes and failures to improve the
+  quality of later results.
+
+- [BOLTs #894][] updates the specification with checks implementations
+  should implement to prevent donating excess bitcoins to miner fees
```
jnewbery, line 180: "with checks implementations should" is a bit difficult to parse. How about splitting the sentence into something like "BOLTs 894 adds new suggested checks on peers' dust limits. Implementors are encouraged to use these checks to prevent..." or similar?

_posts/en/newsletters/2021-10-20-newsletter.md
bitschmidty: fixup'd

dongcarl:
Removed my stub, I looked into the PRs and Issues and didn't think it was that interesting of a change.

_includes/specials/taproot/en/18-trivia.md
gmaxwell: Just commenting on the history here.

It's just showing that eddsa (which is similar to bip340) is widely used.  I don't agree that the design of eddsa was particularly influential.  Or rather, it was a influence but only in the sense that the criteria for what signatures will be accepted/rejected is severely under-specified, leading to actual attacks in cryptocurrencies that use it... and BIP340 took care to not make the same error.  Otherwise, except for the tagged hashes (which eddsa doesn't use either) I believe BIP340 is compatible with the schnorr implementation that we created before we were really looking at eddsa at all.

sipa: The old schnorr signature module in libsecp256k1 didn't have key prefixing (while eddsa does). I think we were aware of eddsa and key prefixing at the time, but it wasn't considered as essential. Only with MuSig's security proof relying on key prefixing, and the key malleability the paper highlighted, did we decide to standardize on key prefixing.

I'd say that BIP340 isn't exactly based on EdDSA, but the end result is rather similar. The fact that another reasonably similar Schnorr-based design existed probably helped being confident about it.

_includes/specials/taproot/en/18-trivia.md
gmaxwell: Schnorr wasn't trying to "do good", he was just trying to make money. Though there isn't anything wrong with making moneyit's not like the idea wouldn't have existed for a decade without him (very similar constructions existed previously-- like the ECDSA-ish ElGamal signature scheme).

Schnorr's unwillingness to open up the patent was controversial even back in the 90s.  Schnorr actually went around [claiming anyone using any kind of discrete log signature violated his patent](https://web.archive.org/web/19991117143502/http://grouper.ieee.org/groups/1363/letters/SchnorrMar98.html) (but due to the aforementioned ElG scheme his threats against DSA weren't credible and only stopped a smaller number of parties).

_includes/specials/taproot/en/18-trivia.md
gmaxwell: So PAIR appears to be down, so I can't log in to check... but US patents receive an extension to their normal term due to delays in granting caused by the patent office, not too infrequently a year or two extra.  So that might be the origin of the 2010/2008 confusion. If so, then it wouldn't really be correct to say it expired in 2008.  Maybe best to just hedge and say 2008-2010. :)  I'd say it's irrelevant except 2010 would actually make it post Bitcoin's deployment (also might explain why EDDSA wasn't published until 2011), so it would be a point that was relevant to the story.

The matter is complicated by the fact that due to the drafting Schnorr's *US* patent wasn't *that* relevant:  Prior to  State Street Bank v. Signature (1998) (1) it was fairly unambiguous in the US that you couldn't patent an abstract algorithm, you had to patent a 'specific machine', and Schnorr's patent attorneys wrote the US patent in a way that would likely only apply to smartcards to avoid that problem (there were other hacks to evade that requirement without being very specific, e.g. saying the patent applied to a "computer readable medium", but apparently they weren't known to the drafter or they decided to be conservative and not use them).  In Europe where patenting abstract algorithms was less of an issue at the time, his patents were written more broadly and so it's actually his European patents that were considered more disruptive and presumably they had their own expiration times.

[(1) Subsequently, In Re Bilski overturned State Street in 2008, resulting in a whole bunch of US patents drafted between 1998 and 2008 having validity problems because they lacked any pretext of tying them to a machine... which has been its own adventure...]

_includes/specials/taproot/en/18-trivia.md
jonasnick: Speaking of EdDSA the story comes to mind when we were specifying the nonce derivation function and quickly realized that for a key-prefixed scheme the public key should be input to the function to protect a user against accidental misuse. @gmaxwell [noticed that this is not the case](https://moderncrypto.org/mail-archive/curves/2020/001012.html) in EdDSA.

_includes/specials/taproot/en/18-trivia.md
real-or-random: Sure he was trying to make money and indeed, there's nothing wrong with. My thought was just that people nowadays understand that patenting your new crypto algorithm is the best way to ensure you won't make any money with it...

_includes/specials/taproot/en/18-trivia.md
gmaxwell: Generally in English when we say that X was influenced by Y we don't *usually* mean by Y's errors and omissions. :P  Key prefixing is a true similarity but inside bitcoin before sighash no input (and likewise, e.g. sighash bug) were considered every transaction signature was inherently implicitly key-prefixed.

PR #668, opened 2021-10-18
170: Update 2021-10-13-newsletter.md

maflcko:
* Normalize all "broadcast(ed)" to the same spelling in this newsletter
* whitelist -> allowlist to mirror the naming use in the source code
* Clarify that `-blocksonly` *does* have a mempool (even though
   it only used for outgoing txs, thus likely empty)

### DIFF
--- a/_posts/en/newsletters/2021-10-13-newsletter.md
+++ b/_posts/en/newsletters/2021-10-13-newsletter.md
@@ -215,8 +215,8 @@ repo], [Hardware Wallet Interface (HWI)][hwi repo],
   there have been two methods to relay blocks: legacy relay and [BIP152][]-style
   [compact block relay][topic compact block relay].
 
-  A node started with the `-blocksonly` setting does not participate in incoming transaction relay to reduce its
-  bandwidth usage, and therefore, does have an empty mempool. Therefore, compact blocks are
+  A node started with the `-blocksonly` setting does not accept relayed transactions from public nodes to reduce its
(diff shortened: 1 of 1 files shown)


_posts/en/newsletters/2021-10-13-newsletter.md
```diff
@@ -210,20 +210,20 @@ repo], [Hardware Wallet Interface (HWI)][hwi repo],
...
-- [Bitcoin Core #22340][] After a block is mined, it is broadcasted to the p2p network,
+- [Bitcoin Core #22340][] After a block is mined, it is broadcast to the p2p network,
   where it will eventually be relayed to all nodes on the network. Traditionally,
   there have been two methods to relay blocks: legacy relay and [BIP152][]-style
   [compact block relay][topic compact block relay].
 
-  A blocks-only node does not participate in transaction relay to reduce its
-  bandwidth usage, and therefore, does not have a mempool. Therefore, compact blocks are
+  A node started with the `-blocksonly` setting does not participate in incoming transaction relay to reduce its
+  bandwidth usage, and therefore, does have an empty mempool. Therefore, compact blocks are
```
jnewbery, line 219: Perhaps even better:

```suggestion
  A node started with the `-blocksonly` setting does not accept relayed transactions from public nodes to reduce its
  bandwidth usage, and so usually has an empty mempool. Therefore, compact blocks are
``````

jnewbery:
@harding any opinion on the style points?

PR #677, opened 2021-10-28
Update cash app compatibility matrix

andozw:
The default receive address is P2SH-P2WPKH. Old accounts that have never
received a deposit might still show a P2PKH address. As soon as the
old account receives a deposit to the P2PKH address, the address will be
rotated to a P2SH-P2WPKH address.

Cash app creates native segwit p2wpkh (bech32) change addresses.
https://www.blockstream.info/tx/bde47e74bde57938385156fc5b5d5abde1392448b535e94992c5aedef2874f79

### DIFF
--- a/_data/compatibility/cashapp.yaml
+++ b/_data/compatibility/cashapp.yaml
@@ -54,19 +54,19 @@ segwit:
     version: "3.5.1"
   features:
     receive:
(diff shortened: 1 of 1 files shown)


bitschmidty (approved):
ACK

Aawesome, thanks @andozw !

PR #679, opened 2021-11-04
Newsletter-173:Translate into Japanese

azuchi:


bitschmidty (approved):
ACK, thanks @azuchi !

PR #686, opened 2021-11-17
Links: update to use new lightning/bolts repo

harding:
Can be tested by:

- In any newsletter from the past two years, click the "Lightning BOLTs" link in the *Notable changes* section
- In the newsletters returned by the following grep, scrolling to the *Notable changes* section and clicking the link for "BOLTs #xxx" `git grep -l -- '- \[BOLTs' | xargs grep '^title: '`

You could also diff the built site from before and after the change; I was too lazy to do that.  (Maybe if I'm ever not lazy, I'll add generating a diff to the CI script.  Don't hold y'alls breath though :smile:)


jnewbery:
ACK. I built the site before and after and diff'ed. I only looked at a subset of the changed generated files, but they all looked good to me. Thanks Dave.

PR #690, opened 2021-11-22
Spell Stack Exchange as two words

Xekyo:
Changed via:
`git grep -l StackExchange . | xargs sed -i s/StackExchange/Stack\ Exchange/`


Xekyo:
After the above replacement `git grep -i stackexchange . ` only found occurrences in links, and 
`git grep -i " stackexchange" . `, i.e. with a preceding space, did not find me any hits.

jnewbery:
ACK. Looks good. Thanks @Xekyo!

PR #691, opened 2021-11-24
Newsletter-176:Translate into Japanese

azuchi:


bitschmidty (approved):
Translations getting so fast that soon you will be authoring the newsletter in Japanese and we will be translating into English after! Thanks @azuchi

PR #692, opened 2021-11-26
Sponsors: add Nayuta

bitschmidty:


jnewbery:
Concept ACK. I'd prefer to have a transparent background to the images, so that if we ever change the background color of the website, the logos still look ok:

image

Here's the png with a transparent background:

nayuta

bitschmidty:
For whatever reason that image you provided lost transparency for me. But I was able to transparent-ize on my side.

jnewbery:
Looks good to me. Thanks Mike!

PR #697, opened 2021-12-08
Newsletter-178:Translate into Japanese

azuchi:


bitschmidty (approved):
ACK thank you @azuchi !

PR #699, opened 2021-12-09
Linker: update links for GitHub API change

### LINKED ISSUE #698: BOLT issue links are broken
In newsletter 177, the link to BOLTs #880 doesn't work. In newsletter 178, the links to BOLTs #906 and #911 are broken.

harding:
Closes #698 

`github.com/org/project/issue/number` used to redirect to either an issue or a PR appropriately.  Now it returns a 404 if the number doesn't correspond to an issue.

`github.com/org/project/pull/number` still has the same redirect behavior, so we update to use that instead.  Tested on both PRs (most of our links) and also an issue, see `/en/newsletters/2020/01/29/#bolts-705` on the preview.

### DIFF
--- a/_includes/linkers/issues.md
+++ b/_includes/linkers/issues.md
@@ -22,18 +22,18 @@ requests, and other templated URLs.
 -->
 {% assign _issues = include.issues | split: "," %}
 {% endcapture %}{% for _issue in _issues %}
-[bitcoin core #{{_issue}}]: https://github.com/bitcoin/bitcoin/issues/{{_issue}}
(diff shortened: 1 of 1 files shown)


bitschmidty:
It could be github changed this functionality back as I dont see problems with the examples provided in https://github.com/bitcoinops/bitcoinops.github.io/issues/698

harding:
@bitschmidty you're correct, the issue appears to be resolved.  I'll close this PR but leave the branch up in case we need it again.  Thanks for testing!

PR #704, opened 2021-12-24
Newsletter-180:Translate into Japanese

azuchi:


bitschmidty:
@azuchi Thank you for the year in review translation and for your help all year!

PR #711, opened 2022-01-19
Newsletter-183:Translate into Japanese

azuchi:


bitschmidty (approved):
Looks good thanks @azuchi

PR #717, opened 2022-02-02
Newsletter-185:Translate into Japanese

azuchi:


bitschmidty (approved):
ACK. Thanks @azuchi !

PR #734, opened 2022-03-16
Typo fix: s/btc-lisp/btc-script/

harding:


bitschmidty:
Added the typo fix to ja version as well.

PR #739, opened 2022-03-30
Reverse osuntokun gossip links

lucasdcf:
The link for Laolu's response to Rusty and the link for his minor update proposal are inverted.


Xekyo (approved):
Looks correct to me.

PR #741, opened 2022-03-31
Update "about" page

jonatack:
Have been inactive here since a bit more than nine months, so it seems best to remove my profile.


bitschmidty:
@moneyball @jonatack Take a look at this alternative based on Steves Alum idea
https://github.com/bitcoinops/bitcoinops.github.io/pull/761

Xekyo:
I'm good with either, but would prefer the alumni section. What do you think, @jonatack?

jonatack:
Sorry for not responding, will look/test soon!

jonatack:
> I'm good with either, but would prefer the alumni section. What do you think, @jonatack?

@Xekyo SGTM.

PR #743, opened 2022-04-06
Newsletter 194 cjdns entry fixups

jonatack:
- fix link to the doc (from jonatack/bitcoin to bitcoin/bitcoin, and to a commit that includes the update merged yesterday)
- add a second pull merged yesterday that covers the most frequent issue users have been running into
- maybe remove link to anonymity networks section (cjdns is an `fc00::/8` reserved IPv6 range overlay network with e2e encryption but does not hide the sender and the recipient from intermediate routers)

Happy to update or drop some of these, per feedback.


bitschmidty:
Thakns for the PR, @jonatack, Added a small change to fix build error.

PR #751, opened 2022-04-27
News197: clarify darosior's request

harding:
Re https://twitter.com/darosior/status/1519281847359975424

PR #752, opened 2022-04-29
Newsletter-197:Apply #751 to japanese translation

azuchi:


bitschmidty (approved):
@azuchi Thanks for always noticing these follow up changes!

PR #757, opened 2022-05-05
Newsletter-198:Translate into Japanese

azuchi:


bitschmidty (approved):
Thanks @azuchi sorry for the delay merging!

PR #763, opened 2022-05-11
Newsletter-199: Add Czech translation

nechteme:


Xekyo (approved):
LGTM, looked over the content via just structure and what google translate gave back to me, built it locally and tested all links.

PR #768, opened 2022-05-18
Newsletter-200: Add Czech translation

nechteme:
Here is Czech translation of the 200th newsletter.

As becomes usual, I added footnotes explaining some terms since there are no topic translations. Additionally, I dared to add a link to a detailed Slovakian article (everybody who reads Czech can read Slovakian) explaining RGB. I hope it's OK, if not, please let me know.


bitschmidty (approved):
@nechteme I built and tested links. Also footnotes / rgb link look good to me

Thank you!

PR #776, opened 2022-06-02
Newsletter-202: Add Czech translation

nechteme:


bitschmidty (approved):
ACK, thanks @nechteme !

PR #779, opened 2022-06-03
Rebuild indices for all languages on preview

Xekyo:
This updates the Makefile to rebuild indices of all languages so that translated articles and links to other languages show up in previews.

Note that I have never worked with Makefiles before, so I'm just taking a stab in the dark here—although it did seem to achieve the desired effect.


harding:
Untested ACK.  Thanks!

PR #786, opened 2022-06-15
Newsletter-203: Add Chinese translation

hulatown:


bitschmidty (approved):
ACK, thank you @hulatown, this is great to have a Chinese translation!

PR #792, opened 2022-06-22
Newsletter-205:Translate into Japanese

azuchi:


bitschmidty (commented):
ACK, thank you @azuchi !

PR #797, opened 2022-06-27
Newsletter-205: Add Chinese translation

hulatown:


bitschmidty (approved):
ACK, thank you @hulatown !

PR #805, opened 2022-07-07
Newsletter-206: Add Chinese translation

hulatown:


bitschmidty (approved):
good, thanks @hulatown !

PR #806, opened 2022-07-08
WIP: Add covenant-related information

harding:
WIP PR to add a collection of information about covenant proposals and covenant-enabled applications to the site.  Seeking concept-level feedback at this time, particularly opinions about the per-entry data and presentation, since changing the templates later will be much harder than changing them now.

### Previews

- Entrypoint: /en/lore/covenants/
- Example covenant-enabled application: /en/lore/covenants/apps/channel-factories/
- Example covenant proposal: /en/lore/covenants/proposals/cat+csfs/
- Example table aggregating covenant info: /en/lore/covenants/analysis/enable-table/

### TODO

- [ ] Fill in remaining information, mainly based on [this sketch](https://gist.github.com/harding/dae085605fa962339358d05e54468c53)
- [ ] Organize commits for review
- [ ] Integrate proposal/application pages with corresponding topic pages
- [ ] Write announcement blog post

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,6 +88,9 @@ test-before-build: $(covprop_validation) $(covapps_validation) $(compatibility_v
 	! git --no-pager grep -i '[d]iscrete log contract'
 
 test-after-build: build
+	## Check for rendering errors in the built site
+	! grep -r ERROR_ _site/
+
 	## Check for broken Markdown reference-style links that are displayed in text unchanged, e.g. [broken][broken link]
 	! find _site/ -name '*.html' | xargs grep ']\[' | grep -v skip-test | grep .
 	! find _site/ -name '*.html' | xargs grep '\[^' | grep .
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,10 @@ compatibility_validation = $(wildcard _data/compatibility/*.yaml)
 compatibility_validation := $(patsubst _data/compatibility/%.yaml,_site/en/compatibility/%/index.html,$(compatibility_validation))
 topic_validation = $(wildcard _topics/en/*.md)
 topic_validation := $(patsubst _topics/en/%.md,_site/en/topics/%/index.html,$(topic_validation))
+covprop_validation = $(wildcard _covprops/en/*.md)
+covprop_validation := $(patsubst _covprops/en/%.md,_site/en/covenants/proposals/%/index.html,$(covprop_validation))
+covapps_validation = $(wildcard _covapps/en/*.md)
+covapps_validation := $(patsubst _covapps/en/%.md,_site/en/covenants/applications/%/index.html,$(covapps_validation))
 
 clean:
 	bundle exec jekyll clean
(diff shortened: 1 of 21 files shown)


harding:
I just noticed that the BIP119 site uses the term "uses" for what I've been calling "applications".  I think "uses" has the advantage of being simpler and clearer, so I plan to switch to that in the next overhaul.

ariard:
See TAPLEAF_UPDATE_VERIFY entry available here : https://github.com/ariard/bitcoinops.github.io/commit/c8d46e5002fbc7a819a18f20390838633b3e88a4

For formatting issues or linking, I think that can be solved a posteriori. Note, the description is a first shot and any statement in the should be taken as lightweighted. I think TLUV is still one of the most "whiteboardy". If you think the entry is mature, feel free to take the commit or if you have more insights to add, let me know.

Among the list of covenants in the referenced sketch, I think there is missing the old [PUSHTXDATA](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2017-September/014963.html) idea. AFAICT, the more recent OP_TX / OP_TXHASH proposals are roughly equivalent and enable the same level of fine-grained transaction introspection, a really useful feature for a number of applications. Anyway, I think one of them deserves an entry.

instagibbs:
concept ACK

Reminder that pretty much any of the most basic CTV use cases can be done(less efficiently!) via APO. Probably would want to analyze how much less efficient, especially in the case of something like congestion trees of course. For other cases like DLC it doesn't matter quite as much, since cooperative resolutions bypass the bloat entirely.

JeremyRubin:
@instagibbs claim about it not mattering for e.g. DLCs is false, one of the benefits of CTV for that purpose is that it can make it like 300x faster to construct, which you can backfill in with cooperative closes.

So in that case it's the opposite sorta; less space efficient with CTV than with signatures, but much faster time wise, and then backfillable with cooperative resolution (or lazy pre-signing backfill AOT of cooperation). 

If you used APO -- unless there is some cool properties of signatures using G as a key or something -- it'd probably neutralize the time efficiency such that the technique might not be as worth it, altho still possibly an improvement.

instagibbs:
> unless there is some cool properties of signatures using G as a key or something 

Whoops, you're right, and yes there is. My memory of the concept was wrong anyways. Thank you for fact-checking me:

> public key P = G.
> When using nonce R = G, signature creation has negligible computational cost (s
> = 1 + H(R, P, m))

https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2022-January/019812.html

ariard:
> @instagibbs claim about it not mattering for e.g. DLCs is false, one of the benefits of CTV for that purpose is that it can make it like 300x faster to construct, which you can backfill in with cooperative closes.

I think here we have a time-vs-space trade-offs, in the sense that if you embed a tree of CTV for any element of your stateful contracts (e.g DLCs) states set, you have a logarithmic witness size. This size would have to be paid by L2 node operators (whatever on-chain, payment-channel level or routed-over-LN) in case of non-cooperative closures. This risk of overhead cost downside might be far above node operators operational risk tolerance.

I think it would be interesting to have quantitative measures of hash-chain-based vs script-in-the-sig based design for DLCs for different sizes of `m` the number of contract states and `n` the number of counterparties committing to the state, where script-in-the-sig would be based on a multi-party signature scheme like MuSig2 we're likely to use in the Bitcoin ecosystem. The performance downside might be reasonable enough for 99% of the DLC payout curves. At the same time, it is expected latency to be taken into accounts by LN routing algorithms in the future, and routing hops-level performance optimizations to be chased out one after the other.

jamesob:
Bummer, this is a great idea and worth doing. Hopefully it can be part of @ariard's upcoming efforts; I'll be happy to pitch in.

PR #807, opened 2022-07-09
Newsletter-207: Translate into Hindi

shaavan:


bitschmidty:
Thank you @shaavan for the translation!

bitschmidty:
> Thanks, @bitschmidty, for the corrections!
> 
> > Pushed two small changes.
> 
> You do have to tell me what kind of sorcery you use to find these corrections! 😄

Just manual. I click each link and check if the link goes to a hi page if available.

I just happened to notice a misformatting of the summary text when reviewing the page (we may have missed this in previous hi translations)

PR #822, opened 2022-07-29
Newsletter209: add Chinese translation

hulatown:


bitschmidty (approved):
ACK, thanks @hulatown and team!

PR #823, opened 2022-08-02
Newsletter210: add Chinese translation

hulatown:


bitschmidty (approved):
ACK, thank you @hulatown !

PR #824, opened 2022-08-03
Set bech32 receive support for Cash App

andozw:
Cash App now defaults to assigning p2wpkh receive addresses with bech32
encoding. Previously issued addresses will remain active until funds are
sent to them. After funds are sent, the next address with be bech32
style


bitschmidty (approved):
Excellent, thank you for the PR and update @andozw !

PR #825, opened 2022-08-03
Newsletter-211:Translate into Japanese

azuchi:


bitschmidty (approved):
Ack, thanks @azuchi !

PR #828, opened 2022-08-04
Newsletter-211: Translate into Hindi

shaavan:


bitschmidty (approved):
I pushed a couple spacing changes but looks good otherwise, thank you @shaavan !

PR #830, opened 2022-08-10
News212 follow-up: amend description of 25610

Xekyo:
There were a few outstanding review comments on yesterday's newsletter, but this one seems especially important to me because the currently published version seems confusing to me.


bitschmidty (commented):
ACK from my side. @adamjonas might have feedback as he was original author

adamjonas:
No problems from me

PR #831, opened 2022-08-11
Newsletter211: Translate into Chinese

hulatown:


bitschmidty (approved):
ACK, thank you @hulatown !

PR #846, opened 2022-08-31
Newsletter-215:Fix lnd release candidate to release

azuchi:


bitschmidty (approved):
ACK, thanks for catching that @azuchi

PR #853, opened 2022-09-07
Newsletter-216:Fix Bitcoin Core version to v0.10.0

azuchi:


glozow:
Oops you're right! Thanks

PR #854, opened 2022-09-07
Newsletter-216:Translate into Japanese

azuchi:
This translation contains fix https://github.com/bitcoinops/bitcoinops.github.io/pull/853. Please point out if this fix is not necessary.


bitschmidty (approved):
ACK, thanks @azuchi !

PR #860, opened 2022-09-10
Newsletter-216: add Chinese translation

hulatown:


bitschmidty (approved):
ACK, thank you @hulatown and @freeyao!

PR #866, opened 2022-09-21
Newsletter-218:Translate into Japanese

azuchi:


bitschmidty (approved):
Looks good, thank you for the translation @azuchi !

PR #868, opened 2022-09-23
Newsletter-217: add Chinese translation

hulatown:


bitschmidty (approved):
ACK, thank you for the translation @hulatown!

PR #877, opened 2022-10-06
Newsletters: s/0.24/24/ in Bitcoin Core RC versions

harding:
My mistake identified by Twitter user VandelayBTC (thanks!) https://twitter.com/VandelayBTC/status/1578052156187353090


Xekyo (approved):
Oops, I think that's on all of us for missing it!

I notice that we sometimes write “RC1” and sometimes “rc1”, perhaps we could also homogenize that? I'd be fine with merging as is, though.

bitschmidty (approved):
ACK

PR #889, opened 2022-10-13
fix(newsletter-219): corrects some errors in the text

acidbunny21:


bitschmidty (approved):
Taking @Copinmalin 👍  as an ACK.

Thanks @acidbunny21 !

PR #894, opened 2022-10-15
(fix)newsletter 221: Highlighting fix

tschweg:
Fixes a missing term highlight in the english newsletter 221


LarryRuane (approved):
ACK
Thanks, good catch!

PR #899, opened 2022-10-20
Newsletter-222:Translate into Japanese

azuchi:


bitschmidty (approved):
Tested links, looks good, thank you @azuchi !

PR #908, opened 2022-10-27
News223: Q&A item fixup

vostrnad:
Hope my correction is what was originally intended.


Xekyo (approved):
Good catch, thank you

PR #911, opened 2022-10-30
Newsletter-223:apply fixup #908 to japanese translation

azuchi:


bitschmidty (approved):
ACK, thanks @azuchi !

PR #914, opened 2022-11-02
Newsletter-224:Translate into Japanese

azuchi:


bitschmidty (approved):
ACK, thanks for the translation @azuchi !

PR #919, opened 2022-11-07
Add an Anchor for Newsletter-220 (zh)

hulatown:
Add an anchor for Newsletter-220 (zh) in order to generate correct link for Newsletter-223


bitschmidty (approved):
Looks good, @hulatown

PR #920, opened 2022-11-08
Newsletter-223: add Chinese translation

hulatown:


bitschmidty (approved):
Ack, thanks @hulatown !

PR #921, opened 2022-11-08
Newsletter-222: Chinese translation format fix

bitschmidty:
@hulatown could you confirm?


hulatown:
Oh! It should be corrected as in the commit. Sorry that I used GitLocalize to translate but forgot to replace this tag.

PR #922, opened 2022-11-08
add zero-conf to compound words list

LarryRuane:
Check this PR -- I'm unsure if "zero-conf" is the convention we actually want. LMK if not and I can update this PR. The command
```
$ grep -i zero.*conf _posts/en/*/*md
```
shows several forms, `zero-conf`, `zeroconf`, and `zero conf`, so we haven't been very consistent.


harding:
ACK on "zero-conf".  I'm also fine with the other spellings except 0conf.

bitschmidty:
ACK on preferring "zero-conf"

Xekyo:
Post-merge ACK

PR #928, opened 2022-11-10
Fix duplicate frontmatter key

aureleoules:
Replaces incorrect `text` key with `title`. This broke my parser :neutral_face:.


bitschmidty (approved):
Good catch, thanks @aureleoules!

PS - curious what are you using this data for?

aureleoules:
A search engine :smile:

PR #932, opened 2022-11-15
Newsletter-205: Chinese translation link fix

hulatown:


bitschmidty (commented):
utACK, thanks @hulatown

PR #938, opened 2022-11-20
Newsletter-225: translate into Chinese

hulatown:


bitschmidty (approved):
ACK, thanks @hulatown !

PR #941, opened 2022-11-23
News227 fixup: Link to right RC

Xekyo:
The link name says rc2, but the link goes to rc1


harding:
Whoops, I also made this exact same substitution error with a different link.  Pushed an updated commit, will merge as soon as tests pass.  Thanks @Xekyo !

Xekyo:
Ah good catch, thanks

PR #942, opened 2022-11-23
Newsletter-227: Translate into Czech

jirijakes:
Translation of #227 into Czech language.

I used to do Czech translations under account @nechteme but lost access to that account. So from now on, I will be submitting translations under this one.


bitschmidty (approved):
ACK, checked links and formatting. Thank you @jirijakes !

PR #946, opened 2022-11-30
Newsletter-228: Translate into Czech

jirijakes:
Translation of 228 into Czech.

I dared to add two additional links (`news20{1,4} package relay`) pointing to information about package relays in Czech translations of the newsletter. Some links (Wiki, semver) point to Czech versions.


bitschmidty (approved):
ACK, thank you @jirijakes I think the different links are ok.

PR #959, opened 2022-12-14
Newsletter-230: Translate into Czech

jirijakes:
Translation of no. 230 into Czech language.


bitschmidty (approved):
ACK, thank you @jirijakes !

PR #965, opened 2022-12-23
Newsletter-230: translate into Chinese

hulatown:


bitschmidty (approved):
Ack, thank you @hulatown and team!

PR #975, opened 2023-01-05
Newsletter-232: Translate into Czech

jirijakes:


bitschmidty (approved):
ACK, thank you @jirijakes !

PR #981, opened 2023-01-11
Newsletter-233: Translate into Czech

jirijakes:


bitschmidty (approved):
ACK, thank you for the translation @jirijakes !

PR #983, opened 2023-01-12
Newsletter-232: Translate into Chinese

hulatown:


bitschmidty (approved):
Ack, thanks for the translation @hulatown! Sorry for the delayed review this week

PR #994, opened 2023-01-27
Newsletter-235: Translate into Czech

jirijakes:


bitschmidty (commented):
@jirijakes looks like there are some "Trailing spaces" that the build caught, can you take a look?

jirijakes:
>  there are some "Trailing spaces" that the build caught, can you take a look?

Ah, I missed them. Fixed now!

bitschmidty (approved):
ACK, thank toy @jirijakes !

PR #1030, opened 2023-02-25
News237: s/Sqlite/SQLite/

harding:
Closes #1005 

Per https://github.com/bitcoinops/bitcoinops.github.io/pull/999#discussion_r1094862609


LarryRuane (commented):
There are occurrences in two older newsletters (2021-10-20 and 2021-12-22); I assume you don't want to change those (are they considered immutable?), but just want to make sure you're aware. Otherwise lgtm.

harding:
@LarryRuane yeah, I'm aware.  Thanks.   :-)  I don't like changing newsletters long after we publish them; 237 is an exception because we discussed this during its writing.

PR #1037, opened 2023-03-02
Newsletter-240-translate-in-french

Copinmalin:


bitschmidty (commented):
Rebased on master to pull in the 239 translation which should fix the broken links

bitschmidty (approved):
ACK, thanks @Copinmalin !

PR #1041, opened 2023-03-08
Newsletter-241:Translate into Japanese

azuchi:


bitschmidty (approved):
The newsletter and blog translations look good, thanks @azuchi !

PR #1050, opened 2023-03-17
Switch ephemeral anchors to OP_TRUE

instagibbs:
BIP/implementation was changed after enough people got confused: https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2023-March/021536.html


Xekyo (approved):
ACK

PR #1055, opened 2023-03-22
Podcast: add 242, update icon logic

bitschmidty:


stickies-v (approved):
Light ACK bd5fb705b0a62e67380f99ff42be579a32fd0899

Everything looks good to me but I'm not very familiar yet with this codebase. I verified that:
- the new include anchor if clause makes sense
- the podcast-callout includes link correctly
- the new references in the podcast and newsletter match with sanitized titles
- the podcast timestamps match when the conversation on that topic starts
- did some cross-checks with the previous week's newsletter

Most notably, the anchors in the podcast are still missing but that seems expected given that the transcript still needs to be added. I think doing this in 2 steps makes sense.

PR #1064, opened 2023-03-29
Podcast: add 243 recap

bitschmidty:


Xekyo (commented):
LGTM

Xekyo (approved):
Unread ACK:

I built a preview locally and tested all links. Haven’t gotten around to reading the whole transcript, but looks reasonable when scrolling over it.

bitschmidty:
Pushed a change to replace BDK core with bdk-core in the transcript. Squashed.

PR #1066, opened 2023-03-30
News244: fix rust bitcoin release notes link

bitschmidty:
I missed this link in review so we should link to the official repo


harding:
ACK.  Sorry!

PR #1069, opened 2023-04-05
Newsletter-245: Add Czech translation

jirijakes:


bitschmidty (approved):
ACK, thank you @jirijakes !

PR #1080, opened 2023-04-11
Update 2023-03-22-newsletter.md

eval-exec:
Fix typo, it should be PR instead of RP


Xekyo (approved):
That looks right to me. Thank you @eval-exec!

PR #1083, opened 2023-04-12
Podcast: simplify newsletter references

kouloumos:
This PR removes the need for boilerplate logic when adding newsletter references. The new logic requires just the podcast timestamps to be added in the related newsletter sections.

**Current logic**:
- In the podcast, recreate the newsletter structure:
  - add title and anchor for each discussed item
  - add timestamp
  - add anchors for each segment in the transcript 
- In the referenced newsletter, for each item, add a link to the podcast page

**New logic**:
- In the podcast:
  - point to the newsletter it references using the `reference` variable in the frontmatter
- In the referenced newsletter:
  - add timestamp (podcast reference mark) for each discussed item

For each podcast, the `recap_references_generator` will parse the `reference` newsletter (declared in frontmatter) looking for podcast reference marks (timestamps) in order to create the newsletter references used by `{% include newsletter-references.md %}` in the podcast page.

### DIFF
--- a/_posts/en/podcast/2023-04-06-newsletter-recap.md
+++ b/_posts/en/podcast/2023-04-06-newsletter-recap.md
@@ -1,63 +1,23 @@
 ---
 title: 'Bitcoin Optech Newsletter #245 Recap Podcast'
 permalink: /en/podcast/2023/04/06/
+reference: /en/newsletters/2023/04/05/
 name: 2023-04-06-recap
 slug: 2023-04-06-recap
 type: podcast
 layout: podcast-episode
 lang: en
 ---
-Mark "Murch" Erhardt and Mike Schmidt are joined by Sergi Delgado Segura to discuss [Newsletter #245][news245].
+Mark "Murch" Erhardt and Mike Schmidt are joined by Sergi Delgado Segura to discuss [Newsletter #245]({{page.reference}}).
 
 {% include functions/podcast-links.md %}
 
 {% include functions/podcast-player.md url="https://anchor.fm/s/d9918154/podcast/play/68345636/https%3A%2F%2Fd3ctxlq1ktw2nl.cloudfront.net%2Fstaging%2F2023-3-10%2F3bdcea63-7ec8-0854-25fe-913e04492aeb.mp3" %}
 
-## News
-
-{% include functions/podcast-note.md title="Watchtower accountability proofs"
-  url="news245 watchtowers" anchor="" timestamp="0:36" %}
-
-## Releases and release candidates
-
-{% include functions/podcast-note.md title="LND v0.16.0-beta"
-  url="news245 lnd" anchor="" timestamp="21:37" %}
-{% include functions/podcast-note.md title="BDK 1.0.0-alpha.0"
-  url="news245 bdk" anchor="" timestamp="22:06" %}
-
-## Notable code and documentation changes
-
-{% include functions/podcast-note.md title="Core Lightning #5967"
-  url="news245 cln5967" anchor="" timestamp="22:59" %}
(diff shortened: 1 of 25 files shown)


harding:
Nice!  This looks like it doesn't require any changes to the newsletter writing/editing flow, so concept ACK from me.

@kouloumos would you mind separating out the change that creates `_plugins/common_utils.rb` and calls it from `_plugins/auto-anchor.rb` into a separate commit?  I'd like to verify that the change doesn't break any existing anchors and having that in a separate commit makes before/after testing easier.

This change will break existing anchors related to the podcast.  I think we could probably significantly reduce the breakage by keeping the existing anchors by dropping their visual identifiers, or we could just bite the bullet and make this one-time change to a relatively new part of the site.  I leave that decision to @bitschmidty and @Xekyo.

The only other thing I plan to test is the impact on site build time.

kouloumos:
PR updated with @harding's commit separation request.

> This change will break existing anchors related to the podcast. [...] or we could just bite the bullet and make this one-time change to a relatively new part of the site.

It's indeed a breaking change, but I saw it as acceptable because those anchors are mostly used/useful internally (linking/navigation between different sections of the website) and because it's a relatively new part of the site the external references to those anchors should be close to zero. Also, note that the affected anchors are only those at the transcript's segment titles as the ones at the beginning (of the page) were already based on the title of the list item and therefore are unaffected (with a couple of exceptions).

> The only other thing I plan to test is the impact on site build time.

Build time on my machine is ~72s for this branch vs ~66s for master

bitschmidty (commented):
@kouloumos thank you for this PR! It will be nice to cutout the boilerplate pieces (and associated time!). Im glad you have the jekyll/rb chops to put this together.

ACK

Im pushing 245/246 pods to this thread and closing associated PRs then we can have everything in the new approach.

bitschmidty:
Retooled pod246 in the new format, which was very smooth, thanks @kouloumos and pushed a commit for it here

Rebased on master to get the latest newsletter editions and had to force push (sorry!)

If @harding is ok regarding any performance concerns, I think we would be ok to merge.

_posts/en/podcast/2023-04-06-newsletter-recap.md
```diff
@@ -18,6 +18,618 @@ Mark "Murch" Erhardt and Mike Schmidt are joined by Sergi Delgado Segura to disc
...
+have a bunch of Lightning-related PRs later in the newsletter if you care to
+comment on any of those.  But otherwise, if you're busy with other things,
+you're free to drop off as well.
+
+**Sergi Delgado Segura**: I'll stay around just in case.
+
+**Mike Schmidt**: Awesome.  Next section of the newsletter is releases and
+release candidates.
+
+{:#lnd}
```
kouloumos, line 405: These ids for the titles of the transcript's sections are not used based on the new logic.

bitschmidty: Absolutely, my oversight. Fixed.

harding:
ACK 9724a7a07f2679e3f153dede090dc8cd0cf16fdf

Verified that a2cb30912673db9785d31b81b7c2d68eb452f340 doesn't break any existing anchors (thanks @kouloumos for splitting that commit out!  For reference in case anyone else verifies this, I rearranged the order of commits for verification so that the function was only called by the legacy code)

Checked the build time; there was no meaningful difference here either way.

Previewed the site locally and clicked some random links to ensure that they worked as expected after the update.

I only quickly skimmed the added podcast content.

This is great, thanks!

bitschmidty:
Thank you again to @kouloumos for this simplifying on the podcast references, will be a nice time saver (for me!)! ❤️

PR #1092, opened 2023-04-23
Podcasts: add 247 recap

bitschmidty:


bitschmidty:
Going to merge this and if there is feedback we can address in a follow up

PR #1122, opened 2023-05-22
Podcasts: add 250 recap

bitschmidty:

PR #1123, opened 2023-05-23
Podcasts: fix automatic reference generation for special sections

kouloumos:
The current logic in `recap_references_generator.rb` doesn't work when a section does not have nested list items, which is the case with the recent special sections in newsletters [250](https://bitcoinops.org/en/newsletters/2023/05/10/#celebrating-optech-newsletter-250) and [251](https://bitcoinops.org/en/newsletters/2023/05/17/#waiting-for-confirmation-1-why-do-we-have-a-mempool). This first became visible with #1122 and that's why this PR is on top of that one, in order to be able to see the change in action. @bitschmidty feel free to either cherry-pick my commits in your PR, or you can merge this one, whatever you prefer.

This is how those special sections will be rendered in the recap podcast page:
preview


bitschmidty (approved):
ACK. Tested:

- build process/production build with no errors
- tested 250 w/ special section and cross linking
- tested previous 6 newsletters linking

Will drop the 250 pod commit from this PR.

Thanks for this update @kouloumos! Sorry I did not realize this might come up during the original podcasts automatic reference generation PR!

Issue #251, opened 2019-10-30
Localization Improvements


bitschmidty:
- [ ] add flags/graphics/drop-down menu to make this a bit more pleasing on the eye
- [x] add a localized `https://bitcoinops.org/<code>/newsletters/` page so international readers have a place to find all newsletters translated to their language
- [x] add documentation/contrib scripts to make it easier for people to contribute translations.
- [ ] add language specific xml feeds

jnewbery:
>  add a localized https://bitcoinops.org/<code>/newsletters/ page so international readers have a place to find all newsletters translated to their language

Those pages are initially going to be sparsely populated (eg we only have newsletter translations going back to 66 for ja). Perhaps we should add an announcement to that page pointing people to where they can go to help translate.

PR #1127, opened 2023-05-25
Indexes: add People index

kouloumos:
### Overview

Topics' mentions are currently manually updated with every newsletter. But the current writing flow and structure of the newsletter allows for automated calculation of mentions.
This PR introduces two main features:
- Automated optech mentions from newsletters
- A new People index that leverages  and enhances automated mentions

### Topics automated mentions

The goal is to automate topics' mentions without degrading the experience. The **1st commit** introduces the bidirectional linking logic by adding support for double-bracket link syntax `[[]]`. When ``[[]]`` is used, the parser in `bidirectional_links_generator.rb` creates a new mention from the newsletter to the `[[topic]]`, using the title of the paragraph as the `title` of the mention.

The **2nd commit** showcases this new logic by making a few changes to a recent newsletter in order for automated mentions to start working. Automated mentions are backwards-compatible with the manual mentions.

#### Limitations

When `optech_mentions` are manually added they often include a more descriptive title than just the title of the paragraph. 
For example, news item [LDK #1794](https://bitcoinops.org/en/newsletters/2023/05/17/#ldk-1794) from [newsletter #251](https://bitcoinops.org/en/newsletters/2023/05/17/):
https://github.com/bitcoinops/bitcoinops.github.io/blob/a051a454766125c957d703458fddf11271d7fd00/_posts/en/newsletters/2023-05-17-newsletter.md?plain=1#L206-L208

when it was mentioned in the dual funding topic it was added as 
https://github.com/bitcoinops/bitcoinops.github.io/blob/a051a454766125c957d703458fddf11271d7fd00/_topics/en/dual-funding.md?plain=1#L110-L111
but with the new automated mentions logic it will be displayed as just `LDK #1794` because the rest of the text cannot be derived from the writing style. You can see this in action in the newsletter podcast recaps ([example](https://bitcoinops.org/en/podcast/2023/05/11/#notable-code-and-documentation-changes)) as the newsletter-reference logic introduced with https://github.com/bitcoinops/bitcoinops.github.io/pull/1083 shares some of the same implementation details.

#### Dealing with Limitations

If this limitation is a deal-breaker, we could introduce some minor changes in the writing style of the newsletter that would allow for the parser to detect a more descriptive title to add to the mention. The exact solution is still an open question.

### New index for People

The **3rd commit** adds a page `/en/people` that provides another way to navigate the website and gather context based on what people are working on. Because it makes use of the automated mentions logic it doesn't need any manual work except of using the new linking syntax as shown in the **4th commit**.

The automated mentions logic can allow for more verbose mentions and easier creation of them.
That's how I originally started working on this; I imagined an enhancement of the topics page mentions that would allow for more efficient context gathering, kind of like I did in https://github.com/bitcoin-core-review-club/website/pull/639. That original idea is not part of this PR as I later realized that it would be a breaking change for a non-validated idea.

The **5th commit** is that same enhancement but instead applied on the newly introduced people index:

The above screenshot also includes the podcast mentions functionality that is introduced with the **6th commit** and showcased with the **7th commit**.

This PR contains only a limited number of People as a way to showcase the feature. If there is positive sentiment for this feature, I could add the rest of the people in a subsequent PR(s).

**Build time change**: I see a 6s increase in my local builds, from ~66s to ~72s.

### Open Questions

- Is automated mentions for Topics a feature worth having? 
  - If yes, what do you think about the described limitation?
  - Would it make sense for Topics mentions to be also verbose?


bitschmidty:
On the podcast side of things, it would be very easy for me to add the `[[]]` tagging for participants on our weekly show, since we already list them on each podcast page. I would commit to doing that portion, if the team agreed it would be interesting for the site. 

I think a mentions outside of explicit podcast guests could be a bit hairy, resulting in some folks being discouraged due to their type of work not being acknowledged) and others attempting to game the mentions (or am I thinking too highly of Optech that someone would try to game Optech mentions?).

Would it be possible to auto-generate the people/ pages when someone new is referenced?

Dave is both the primary topic author as well as the one that provides the weekly manual entries, so on those items I'd defer to him. Additionally, Dave's idea of compiling topics that is affected potentially here.

Dave's idea of the project index could be interesting as well, not sure if that could be a separate PR or consideration or bundled in here.

@kouloumos thanks for opening this and waiting so long for feedback.

PR #1138, opened 2023-06-02
Add well-known nostr

bitschmidty:


harding:
Untested ACK f3ac0d362d5166f6b10e880fff417c3fdd3800da .  This matches the key I received out of band.  Thanks!

PR #1149, opened 2023-06-12
Newsletter-252: translate into Chinese

bitschmidty:
@hulatown


hulatown:
Thanks Mike! LGTM. I will try to avoid conflict next time

PR #1163, opened 2023-06-21
News256: add topic entries

harding:
I forgot to add these during the edit of the newsletter before publication, so add them now.


bitschmidty (approved):
ACK, thanks @harding !

PR #1176, opened 2023-07-04
News-fr 254 : edit to correct wrong link 

Copinmalin:
correct wrong link : 
{% include specials/policy/fr/04-estimation-taux-de-frais.md %}


bitschmidty (approved):
ACK, thanks @Copinmalin I missed that in reviewing previously

PR #1161, opened 2023-06-21
Podcast: add 255 recap

bitschmidty:

PR #1187, opened 2023-07-07
[Tests] Suppress duplicate-words-test on podcast transcripts

harding:
Addresses https://github.com/bitcoinops/bitcoinops.github.io/pull/1161#discussion_r1240212915

Apparently some some people often say the same word twice in a row row.


bitschmidty (approved):
Tested ACK, thanks @harding !

- tested podcast transcriptions with duplicate words and it performed as expected
- tested duplicate words on a newsletter to verify they are still caught

PR #1188, opened 2023-07-07
Topic index: bold topics mentioned within the past year

harding:
Direct preview link: https://deploy-preview-1188--bitcoinops.netlify.app/en/topics/

This bolds the name of each topic on our index page if it has been mentioned within the past year, indicating which topics someone looking for the latest information might want to concentrate on reading.

This is personally useful to me as I begin to update these entries.  I think it might also be generally useful, but it does add some visual clutter, so I'm definitely not pushing hard for this.

### DIFF
--- a/en/topics.md
+++ b/en/topics.md
@@ -4,6 +4,7 @@ permalink: /en/topics/
 layout: page
 ---
 {% include linkers/topic-pages.md %}
+{% assign year_ago_date = 'now' | date: "%s" | minus: '31536000' %}
 
 {% comment %}<!-- Build an "ENDTOPIC"-separated string with
 Markdown-style links for each topic or topic alias.  We use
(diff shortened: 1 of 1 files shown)


bitschmidty (commented):
ACK, works as intended.

I dont have another usability suggestion here, but I could see visitors on this page thinking the bold topics as having some other importance/prioritization meaning outside of the recent mentions.

I wouldn't

harding:
Yeah.  I agree; if we want to convey this information, we can find a better way.  Closing, but I'll leave the branch around for potential future use.

murchandamus:
Perhaps we could add a function to sort by recency next to the alphabetical sort?

harding:
@murchandamus certainly, although an open question to me was how many visitors would find that useful.

I think maybe an alternative approach to the one taken in this PR is to invert the results, e.g. instead of emphasizing the topics mentioned within the past year, de-emphasize those that haven't been mentioned in >2 years (or whatever arbitrary number).  Again, though, I'm not sure how many visitors would find that useful.

PR #1189, opened 2023-07-08
Newsletter-257: translate into Chinese (#64)

hulatown:
Newsletter-257: translate into Chinese


bitschmidty (approved):
ACK, thank you @hulatown !

PR #1193, opened 2023-07-11
Topics: update primary references and see also sections

harding:
Also includes a few other small changes, plus a separate commit that starts the process of changing our usage of the terms "multisigature" and "multisig" to, respectively, "scriptless multisignature" and "scripted multisig(nature)", which came about as I was writing about them in a different context and had some discussions with a few Bitcoin wizards.  It's easy for me to drop that separate commit if that change is controversial.

### DIFF
--- a/_topics/en/channel-commitment-upgrades.md
+++ b/_topics/en/channel-commitment-upgrades.md
@@ -36,9 +36,15 @@ optech_mentions:
     url: /en/newsletters/2022/04/06/#updating-ln-commitments
 
 ## Optional.  Same format as "primary_sources" above
-# see_also:
-#   - title:
-#     link:
+see_also:
+  - title: Anchor outputs
+    link: topic anchor outputs
+
+  - title: "LN-Symmetry (Eltoo)"
+    link: topic eltoo
+
+  - title: PTLCs
+    link: topic ptlc
 
 ## Optional.  Force the display (true) or non-display (false) of stub
 ## topic notice.  Default is to display if the page.content is below a
--- a/_topics/en/rendez-vous-routing.md
+++ b/_topics/en/rendez-vous-routing.md
@@ -67,6 +67,12 @@ see_also:
   - title: Unannounced channels
     link: topic unannounced channels
 
+  - title: Onion messages
+    link: topic onion messages
+
+  - title: Offers
+    link: topic offers
+
 ## Required.  Use Markdown formatting.  Only one paragraph.  No links allowed.
 ## Should be less than 500 characters
 excerpt: >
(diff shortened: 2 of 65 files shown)


_topics/en/channel-commitment-upgrades.md
```diff
@@ -36,9 +36,9 @@ optech_mentions:
     url: /en/newsletters/2022/04/06/#updating-ln-commitments
 
 ## Optional.  Same format as "primary_sources" above
-# see_also:
-#   - title:
-#     link:
+see_also:
+  - title: Anchor outputs
+    link: topic anchor outputs
```
murchandamus, line 42: Perhaps LN Symmetry should be linked as a related topic here as well, since it essentially is a replacement of the commitment scheme?

```suggestion

  - title: "LN-Symmetry (Eltoo)"
    link: topic eltoo
```

_topics/en/rendez-vous-routing.md
```diff
@@ -67,6 +67,12 @@ see_also:
   - title: Unannounced channels
     link: topic unannounced channels
```
murchandamus, line 69: I think “Blinded Paths” may be the main name for this meanwhile, so perhaps the title and alias should be switched out?

murchandamus (approved):
ACK with a couple suggestions

Strong concept ACK on the terminology update regarding multisig vs multisignature.

_topics/en/rendez-vous-routing.md
harding: I agree blinded paths is now the main name for the thing we describe on this topic page.

For minor renames (e.g. "bech32" -> "bech32(m)"), I feel comfortable changing the YAML `title` field and creating an alias as the URL still makes sense (e.g. `/en/topics/bech32`).  For major renames (e.g. "SIGHASH_NOINPUT" -> "SIGHASH_ANYPREVOUT"), I want to also change the URL, which requires renaming the file, adding a redirect field, and potentially making other mildly invasive changes to the site.  I think that's best done in a separate PR, either an independent PR just for the rename or a PR which significantly updates the topic description.

This page is one of several that I plan to rename and significantly update in the near future, so I'm going to punt this until then.

_topics/en/channel-commitment-upgrades.md
harding: Indeed!

murchandamus:
ACK https://github.com/bitcoinops/bitcoinops.github.io/commit/cb0dd9395c21bdbce1d93002687edd66c0ab3378

bitschmidty (commented):
ACK 

Incredibly thorough, thank you @harding !

PR #1208, opened 2023-07-19
Podcast: add 230 recap

bitschmidty:


murchandamus (commented):
I haven’t read this completely, but it looks like it’s ready to go?

bitschmidty:
Im ok to merge it (even if we are waiting on 230), if you are @murchandamus

murchandamus:
Ah, so the transcription is there, but we don’t have the audio track?

bitschmidty:
Sorry, I meant Im good to push 230 even if we are waiting on 231.

murchandamus:
Ah yeah, it’s over half a year anyway, I don’t think it matters if one is missing in the order.

PR #1235, opened 2023-08-09
Newsletter-263: Translate into Czech

jirijakes:


bitschmidty (commented):
ACK, thank you @jirijakes !

PR #1260, opened 2023-08-28
Policy Series and Bitgo-musig2 (zh): Correction and Typo-Fix 

hulatown:


murchandamus:
Looks good to me. Thanks for following up with corrections, @hulatown, @editor-Ajian!

PR #1263, opened 2023-08-30
Newsletter-266: Translate into Czech

jirijakes:
Additionally, I found better translation of ‘side channel attacks’, changed in the past newsletters.


bitschmidty (approved):
ACK on links and formatting.

Thank you @jirijakes !

PR #1297, opened 2023-09-09
Newsletter-267: Translate into Czech

jirijakes:


bitschmidty (approved):
ACK for both updates, thanks @jirijakes !

PR #1302, opened 2023-09-13
Newsletter-268:Translate into Japanese

azuchi:
About lnd release, there seems to be a tag for [v0.17.0-beta.rc3](https://github.com/lightningnetwork/lnd/releases/tag/v0.17.0-beta.rc3), can we keep rc2?


bitschmidty (approved):
@azuchi Thanks for the translation.

I think keeping rc2 is fine as rc3 was just released and we will cover next week likely

PR #1305, opened 2023-09-14
Newsletter-268: Translate in French

Copinmalin:


_topics/fr/client-side-validation.md
bitschmidty, line 1: @Copinmalin Thanks for translating the topic pages!

Unfortunately, the site doesnt support translations in the topic section (yet). Can you drop these from this commit and save for a future PR once we can support that?

bitschmidty (changes requested):

_topics/fr/client-side-validation.md
Copinmalin: Done !

bitschmidty (approved):
ACK, thank you @Copinmalin !

PR #1312, opened 2023-09-20
Fix mistypes in B'SST description in 2023-09-20-newsletter.md

dgpv:
It is "Tracer", not "Trace", and also "Elements", not "Element"


bitschmidty (approved):
@dgpv Thanks for the correction, my apologies

PR #1342, opened 2023-10-04
Newsletter-271: Translate into Czech

jirijakes:
Additionally fixes an imprecise translation in an older newsletter.


bitschmidty (approved):
ACK, thank you @jirijakes !

PR #1373, opened 2023-10-29
Newsletter-273: add Chinese translation

hulatown:
* newsletter273zh

* Apply suggestions from code review

* fix style

---------


bitschmidty (approved):
ACK, thank you @hulatown !

PR #1381, opened 2023-11-04
fix amounts in overflow bug description

MBaum21:
**easy changes:**
`INT64_MAX` is `9223372036854775807` (`2**63-1`)
`INT64_MIN` is `-9223372036854775808` (`-2**63`)

**other changes:**
For whatever reason, post # 1 in [bitcointalk topic 822](https://bitcointalk.org/index.php?topic=822.0) contains a different output amount than post # 3  
`post #1: 92233720368.54277039 (0x7ffffffffff863af)`  
`post #3: 92233720368.54275808 (0x7ffffffffff85ee0)`  
I'm assuming post # 3 ist right, because it's `2**63-500000`, and the sum of outputs is a round number: -0.01 BTC (minus 1 million sats). (calculation: `9223372036854275808*2 - 2**64 = -1000000`)

### DIFF
--- a/_topics/en/soft-fork-activation.md
+++ b/_topics/en/soft-fork-activation.md
@@ -490,13 +490,13 @@ occasions included:
 
 - **Fix value overflow bug (August 2010):** someone
   created a transaction that spent 0.5 BTC to two outputs worth
-  92,233,720,368.54277039 BTC.  Bitcoin requires that the output
+  92,233,720,368.54275808 BTC.  Bitcoin requires that the output
   amount be less than or equal to the input amount, but this was
   checked by adding the output values into a single 64-bit
   integer which can only hold a maximum value of
-  9,223,372,036,854,776 sats (about 92 billion BTC) before rolling over to negative numbers,
-  starting with -9,223,372,036,854,776 sats.  This meant it looked like the
-  transaction spent a total of -0.1 BTC (negative 0.1 BTC).  This
+  9,223,372,036,854,775,807 sats (about 92 billion BTC) before rolling over to negative numbers,
+  starting with -9,223,372,036,854,775,808 sats.  This meant it looked like the
+  transaction spent a total of -0.01 BTC (negative 0.01 BTC).  This
   bypassed another rule that forbid individual negative outputs---but
   not negative aggregate amounts---since it assumed the sum of any set
   of positive values would also be positive.  This gave the person who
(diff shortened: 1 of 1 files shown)


murchandamus:
I spent some time looking into this. The raw transaction from the value overflow incident is posted here:

The input used in the overflow transaction was the 0th output of https://mempool.space/tx/237fe8348fc77ace11049931058abb034c99698c7fe99b1cc022b1365a705d39

_topics/en/soft-fork-activation.md
```diff
@@ -490,13 +490,13 @@ occasions included:
...
+  92,233,720,368.54275808 BTC.  Bitcoin requires that the output
   amount be less than or equal to the input amount, but this was
   checked by adding the output values into a single 64-bit
   integer which can only hold a maximum value of
-  9,223,372,036,854,776 sats (about 92 billion BTC) before rolling over to negative numbers,
-  starting with -9,223,372,036,854,776 sats.  This meant it looked like the
-  transaction spent a total of -0.1 BTC (negative 0.1 BTC).  This
+  9,223,372,036,854,775,807 sats (about 92 billion BTC) before rolling over to negative numbers,
+  starting with -9,223,372,036,854,775,808 sats.  This meant it looked like the
+  transaction spent a total of -0.01 BTC (negative 0.01 BTC).  This
```
murchandamus, line 499:
```suggestion
  starting with -9,223,372,036,854,775,808 sats.  This meant it looked like the
  transaction outputs summed to a total of -0.01 BTC (negative 0.01 BTC).  This
```

murchandamus:
Okay, yeah, you are right. The two outputs sum to –0.01 ₿ which then explains why the coinbase collects 50 ₿ new coins and 0.5 ₿ –(–0.01 ₿) = 50.51 ₿ from the transaction.

murchandamus:
Oh, I just realized that thread is from last month. Is that you digging up the actual raw transaction? :)

MBaum21:
Thanks for your response @murchandamus :)

Nice investigation by ercewubam in that thread from march. No, that's not me.
Good to have the amount confirmed, i.e. 92233720368.54275808 ₿.

> I will look over the text, I think we might need to amend the text a little further.

Is there still something to amend? :)

murchandamus (approved):
I think I have rediscovered the minutiae of this situation.

• There were two 92 billion BTC outputs, each individually 500,000 sats below INT64_MAX
• In sum, these two outputs add up to -1,000,000 sats or -0.01 BTC
• This resulted in a transaction fee of 0.51 BTC

I’ve added half a sentence that would have helped me catch on faster, and otherwise concur with the proposed changes by @MBaum21.

bitschmidty (approved):
ACK, thank you @MBaum21  and @murchandamus

PR #1397, opened 2023-11-15
French translation fixups

bitschmidty:
@Copinmalin can you take a look?

It looks like the timestamps for the podcast were copied over during the translation and I missed it in review.


Copinmalin:
Oh OK ! So now I stop to insert this information in french transalation, thanks

PR #1435, opened 2023-12-17
Newsletter-280: translate into Chinese

hulatown:


hulatown:
Hi! I can pass the check by `make production` on my local environment. But it seems failed with Travis CI. Do you what the reason might be and how could I fix it? Thanks

bitschmidty:
> Hi! I can pass the check by `make production` on my local environment. But it seems failed with Travis CI. Do you know what the reason might be and how I could fix it? Thanks

Ive opened an issue for this and began work on a fix which is more complicated since it involves upgrading to ruby 3.x.

bitschmidty (approved):
ACK, checked links and formatting.

Ive also run travis-related checks locally to ensure that they pass.

Thank you @hulatown !

PR #1440, opened 2023-12-19
Build: remove Travis CI bundler update directive

bitschmidty:
One does not simply remove the command that is required in order to not break the builds but which is also currently breaking the builds

image

But it works(?)


harding:
I can't see any harm here.  Let's do it.  Let's look into upgrading all our deps next year; there's probably some performance improvements and other stuff we're missing out on.

PR #1448, opened 2023-12-26
Newsletter-282: translate into Chinese 

hulatown:
and fix some links in related newsletters


bitschmidty (approved):
ACK, thank you @hulatown !

PR #1458, opened 2024-01-09
Fix link to Rendez-vous routing spec proposal

shaavan:


bitschmidty (approved):
ACK, thanks for noticing this and opening a PR @shaavan

PR #1529, opened 2024-02-26
Newsletter-1: translate into Chinese

bittomhan:


bittomhan:
I've noticed that when I use "make" locally, the tests pass, but when using "make production" remotely, it fails due to errors in the title's name and slug. Upon trying "make production" locally, I encountered the same errors. I've corrected the title and resubmitted.

A small observation, and perhaps it was my oversight, but moving forward, I will strictly use "make production" for testing before I make any submissions.

bitschmidty:
> I've noticed that when I use "make" locally, the tests pass, but when using "make production" remotely, it fails due to errors in the title's name and slug. Upon trying "make production" locally, I encountered the same errors. I've corrected the title and resubmitted.
> 
> A small observation, and perhaps it was my oversight, but moving forward, I will strictly use "make production" for testing before I make any submissions.

FYI, Ill usually do `make preview` to initially draft things for quicker iteration and then run `make production` before opening a PR or adding a commit to a remote. This sounds similar to your workflow.

bitschmidty (approved):
Tested links, page formatting, and a rough translation of the content and all look good!

ACK, thank you @bittomhan

PR #1541, opened 2024-03-05
Prefer blockchain over block chain

LarryRuane:
The term blockchain has become accepted since the style guide was written in 2019. There are already significantly more instances of "blockchain" than "block chain" in this repository.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -93,7 +93,7 @@ beginning of a sentence).
 | Use | Don't use | Notes |
 |-|-|-|
 | anti fee sniping | anti-fee sniping, anti-fee-sniping | |
-| block chain | blockchain | |
+| blockchain | block chain | "block chain" is acceptable in unmodified older documents |
 | chain split | chainsplit | |
 | CPFP or Child Pays For Parent | Child-Pays-For-Parent | |
 | coinjoin | Coinjoin, coinJoin or coin-join | |


harding:
Eh, almost all uses of "blockchain" in the repo are in translations, transcripts, URLs, proper names, or code (like RPC names).  I see only about 10 actual violations of the policy in `_posts/en/newsletters` and none in `_topics/en`

That said, I agree most other technical sources have moved to "blockchain", so a concept ACK from me with the proviso that I think we should apply standard code style policy change rules: we only update an older document that says "block chain" to say "blockchain" if we're making a substantial edit for some other purpose.

@LarryRuane if you agree, could you please add a note to the blockchain entry; something like, '"block chain" is acceptable in unmodified older documents'

LarryRuane:
Yes, I agree, I just force pushed your suggested update and also removed the part of the commit message about "significantly more" instances of blockchain (good catch, sorry, I didn't know that).

harding:
ACK da85a2d67d4fdf7f3f606596958ec7fa1b941a9a

If there are no objections, I plan to merge this on Sunday, March 10th, before opening the PR for next week's newsletter.

bitschmidty:
I dont have a preference here, so ACK from me.

murchandamus:
Sounds good, I think I’ve been writing blockchain for many years already

Issue #1551, opened 2024-03-12
Build site using Hugo instead of Jekyll (upgrade tracking issue)


harding:
Since its inception, this site has always been built with [Jekyll][] but Jekyll appears to be largely unmaintained; it's [changelog](http://jekyllrb.com/docs/history/) indicates the last time a major feature was added was "4.0.0 / 2019-08-19".  I don't we particularly need new features from the static site compiler, but as we consider adding new features to the site, I want to make sure we're building on a foundation that will continue to be supported for years to come.

[Hugo][] appears to be actively maintained.  In my tests, it also completes a whole site compile at least 10x faster than Jekyll.  I've locally completed some preliminary conversion of our site to Hugo and I don't see anything that will prevent us from a complete conversion.  This issue for for discussing any objections to a conversion as well as tracking its progress.  The plan is:

- [ ] Work on converting the site in a local branch until near parity has been reached between the Jekyll and Hugo versions
- [ ] (In parallel with the above) Open PRs for changes to Jekyll content to simplify conversion
- [ ] When near-parity is reached:
  - [ ] update our CI infrastructure to begin building and testing both Jekyll and Hugo versions 
  - [ ] Host a preview of the Hugo version for reviewers
- [ ] After a few weeks of parallel building and collecting reviews, switch live site to Hugo version (with an easy revert option in case of problems); continue building and testing the Jekyll version
- [ ] After a few weeks with Hugo running in production, drop support for the Jekyll version

Please let me know your thoughts on converting.  Until we complete conversion, I suggest a moratorium on adding major new features to the site, with the exception of an update to the Compatibility Matrix (as work on that has already begun).

[jekyll]: http://jekyllrb.com
[hugo]: https://gohugo.io/

murchandamus:
I like this approach. While it might be a bit more work to maintain the dual infrastructure until the switch-over, it should make the switch significantly smoother than some sort of one-time switch over. I don’t have much information on either Hugo or Jekyll, but generally running on an unmaintained stack seems like a source of future pain, so what you write sounds sensible to me. Also very excited not to wait a few minutes every time I want to preview a branch. :D

kouloumos:
This is not an immediate action, but when/if you need to re-create the Topics index in Hugo native way, I might be able to help as we essentially [mirrored the index for bitcointranscripts](https://github.com/bitcointranscripts/bitcointranscripts.github.io/pull/56).

harding:
[...]

PR #1552, opened 2024-03-13
Get rid of some old template helpers

harding:
Part of #1551

- `linkers/github-log.md` used to display the diff of commits that I reviewed for the newsletter each week.  We only used it for a couple dozen newsletters before switching to the modern format that does not contain that information.  This PR drops it and resolves all previous diff references to the root project repositories (like we do in modern newsletters).

- `linkers/newsletters.md` contained links to each newsletter for our first hundred newsletters so we could writing things like `see [Newsletter #1]`.  Modern newsletters link directly to the anchor for particular items.  This PR drops the old links and resolves them to the particular item being referenced.

  - `_contrib/update-newsletter-index-variables` was a script run annually to update the above `newsletters.md` file.  This PR drops it as no longer being necessary.


_posts/en/newsletters/2019-04-23-newsletter.md
```diff
@@ -155,7 +155,7 @@ backported to its pending release.*
   not be useful to a high-level API consumer.  Although there are
   differences in implementation and motivation, this seems conceptually
   related to [C-Lightning #2382][] as described in the *notable code
-  changes* section of [Newsletter #36][] in two separate bullet points.
+  changes* section of [Newsletter #36][].
```
harding, line 158: This should be the only visible content change in this PR.  During a previous change to the site, we combined two bullet points into one to use with the auto-anchor plugin; however, this text that references that page still says that they're separate; I've updated it to reflect what readers will see today.

murchandamus (commented):
Concept ACK

I did a light review, this looks good to me, but I have not tested all links or something as in-depth as that.

PR #1559, opened 2024-03-15
Newsletter-291: translate into Chinese 

hulatown:


bitschmidty (approved):
CK, thank you @hulatown !

PR #1565, opened 2024-03-20
Newsletter-293: translate into Chinese

hulatown:


bitschmidty (approved):
ACK, thank you @hulatown !

bitschmidty (changes requested):
@hulatown actually the translation looked good, but I see there are other commits and file changes here, can you take a look?

I think you can just rebase on master and be fine here.

hulatown:
> @hulatown actually the translation looked good, but I see there are other commits and file changes here, can you take a look?
> 
> I think you can just rebase on master and be fine here.

Rebased! It seems that it was affected by some forced push before, but it looks fine now. Can you please check again?

bitschmidty (approved):
ACK, thank you @hulatown !

PR #1572, opened 2024-03-25
Newsletter-5: translate into Chinese

bittomhan:


bitschmidty (approved):
ACK, thank you @bittomhan!

Ill merge and if @hulatown has further suggestions we can address separately.

PR #1581, opened 2024-03-29
Newsletter-294: translate into Chinese (#103)

hulatown:


bitschmidty (approved):
ACK, thank you @hulatown !

PR #1604, opened 2024-04-07
Newsletter-18: translate into Chinese

bittomhan:
Newsletter 16 in zh reuired


bitschmidty:
> Newsletter 16 in zh reuired

Ready to be rebase on master for 16

PR #1619, opened 2024-04-18
Fix use of now-removed github-log.md

harding:
#1552 was opened about a month ago to remove github-log.md

Several translations of old newsletters were made after that.

I then merged #1552 without checking that it built on master, which led to a build failure.  Sorry!  This fixes it.


harding:
All tests passed, merging!

PR #1626, opened 2024-04-24
Newsletter-299:Translate into Japanese

azuchi:


bitschmidty (approved):
ACK, thanks @azuchi !

bitschmidty:
Having build/GH mergability issues, so rebased on master and force pushed to trigger a new build.

PR #1628, opened 2024-04-25
Topics: rename two fields to be compatible with Hugo

harding:
Part of #1551 project to switch to the Hugo static site generator.

This changes the name of two fields in the topic files, `aliases` and `categories`.  Hugo uses both of those field names for special purposes, so this PR gives them alternative names.  This should not have any effect on the site output.

- "aliases" to "title-aliases".  In Hugo, "aliases" creates copies of the current page at the indicated alias urls.

- "categories" to "topic-categories".  In Hugo, "categories" results in the creation of category pages.

Both of those are cool and we'll probably use them in the future, but we want the initial switchover to produce two nearly identical sites.

### DIFF
--- a/_contrib/new-topic
+++ b/_contrib/new-topic
@@ -30,12 +30,12 @@ title: ${topic}
 # shortname: foo
 
 ## Optional.  An entry will be added to the topics index for each alias
-#aliases:
+#title-aliases:
 #  - Foo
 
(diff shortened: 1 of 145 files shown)


bitschmidty (approved):
ACK

Thanks for moving this along, @harding !

I've:

- checked for any missed instances of these variables
- built locally
- smoke tested some topics pages

PR #290, opened 2019-12-08
Newsletters: add 76 (2019-12-11)

harding:
- [x] Add link for LND release

PR #1629, opened 2024-04-25
Posts: check YAML formatting

harding:
Part of #1551

Some of our posts contain an extra space in the YAML header between a key and value, e.g.:

    title : 'Bulletin Hebdomadaire Bitcoin Optech #290'

Jekyll strips the extra space, so the key becomes `title`.  Hugo preserves the extra space, so the field doesn't work as expected.

This PR eliminates the extra space in the source file and adds a test to help avoid future extra spaces.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,8 +43,11 @@ test-before-build: $(compatibility_validation) $(topic_validation)
 	@ ## - MD009: trailing spaces (can lead to extraneous <br> tags
 	bundle exec mdl -g -r MD009 .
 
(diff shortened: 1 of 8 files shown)


bitschmidty (commented):
Pushed a small comment wording change.

- Pulled locally and built
- Rebased on master which included a new fr translation that violated the check
- Confirmed that the violation was caught by the new check
- Fixed the new fr translation format to be compliant

@Copinmalin be aware that this PR makes the yaml formatting (title/permalink/name fields) more strict

murchandamus:
Post merge ACK

PR #1633, opened 2024-04-27
Compat: use category pages instead of pages + data files

harding:
Part of #1551 upgrade to Hugo

Before this PR, the source files for the compatibility matrix information are `_data/compatibility/<wallet>.yaml` and `en/compatibility/<wallet>.md`.

After this PR, the only per-wallet source files are `_compat/en/<wallet>.md`.

All pages have the same URLs and essentially identical content as verified using `diff -ruNw _oldsite _site/ | colordiff | less -R`

This merging of pages leaves us with some redundancies that could be eliminated, but optimization seems premature given that we'll have to redo things a bit when we move to Hugo anyway.


bitschmidty (commented):
ACK,

- built locally
- smoke tested compatibility pages
- reviewed documentation/contributing update, makefile check updates, front end rendering code updates

PR #1649, opened 2024-05-08
News301 correction: OP_CHECKSIG -> OP_CHECKSIGVERIFY

harding:
Reported by @darosior: https://github.com/bitcoinops/bitcoinops.github.io/pull/1645/files/e5875a622cf59a86b389c78a2c6959fcf62438af#r1593946375

I'll merge as soon as tests pass.

### DIFF
--- a/_posts/en/newsletters/2024-05-08-newsletter.md
+++ b/_posts/en/newsletters/2024-05-08-newsletter.md
@@ -82,7 +82,7 @@ and descriptions of changes to popular Bitcoin infrastructure software.
   commits to the actual size of the ECDSA signature.  For example:
 
   ```text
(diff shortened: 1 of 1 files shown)


darosior:
Sorry i should have commented earlier, and maybe it's not that important, but now the script terminates with an empty stack. I think the EQUALVERIFY should be EQUALs in each of the if branch to terminate with 1 on the stack.

-------- Original Message --------

harding:
@darosior ah, indeed.  I think it's ok to leave as-is.  It's an understandable code fragment with the change to OP_CHECKSIGVERIFY and we explicitly say in the opening paragraph that we've elided important security details.  Thanks again!

PR #1661, opened 2024-05-15
Newsletter-21: translate into Chinese

bittomhan:
Actions:

1. Updated the link comment in issue #6, changing "C-lightning" to "Core Lightning" instead of "c-lightning."
2. Fixed a link error in issue #19 that affected issue #21, and modified the workflow to avoid such situations in the future.
3. Added the translation for issue #21.


bitschmidty (approved):
ACK

Checked news21 zh formatting and links

PR #1663, opened 2024-05-15
Newsletter-24: translate into Chinese

bittomhan:
Based on # 23


bitschmidty (approved):
ACK, thank you @bittomhan

PR #1670, opened 2024-05-16
Newsletter-30: translate into Chinese

bittomhan:


bittomhan:
The original text here contains a "[and]" but there is no link as a comment afterwards. I removed the space in the translated Chinese.

`make preview` checked

bitschmidty (approved):
ACK thank you @bittomhan

Issue #39, opened 2018-07-20
Tidy up Workshop page


jnewbery:
I think we can tidy up the workshops page https://bitcoinops.org/workshops/.

Suggestions:
- start with a more general introduction on our workshops, and then have summary details of the SF workshop below
- emphasize that the workshops are Chatham House rules to encourage open and candid conversations, but that general findings may be published and fed back to the open source community (we've already had two requests to info@ asking for recordings/streamings, which is a reasonable request. It'd be helpful to explain our reasoning for not offering those).
- add a call to action.
    - companies to be involved
    - companies to host
    - open source developers to attend (or perhaps speak if we change the format)
    - suggestions for locations/topics

PR #1675, opened 2024-05-16
Newsletter-39: translate into Chinese

bittomhan:


bittomhan:
I noticed that there is an inserted file. Currently, under "bech32," there is only English. Can I directly create the path and file "bech32/zh/02-stats.md"?

bittomhan:
> > I noticed that there is an inserted file. Currently, under "bech32," there is only English. Can I directly create the path and file "bech32/zh/02-stats.md"?
> 
> @bittomhan
> 
> Yes, this is part of the series on bech32 (see english version here: https://bitcoinops.org/en/bech32-sending-support/)
> 
> You can see a similar setup with the _includes/specials/policy/zh/ we did more recently. However, I am seeing that when we did the bech32 special, we did not internationalize it in the /en/ folder. That said you may be able to put the zh series in a _includes/specials/bech32/zh/ folder and reference it from the newsletters (and blog) accordingly. Feel free to give that a try as I am on the road and wont be able to assist for the next week or so.

I attempted to create the Chinese translation for the Bech32 related content in the `_includes/specials/bech32/zh/` path and it has passed the tests.

I translated `02-stats.md` into Chinese and placed it in `Newsletter #39`, which links correctly. The local make production also passed.

However, `02-stats.md` needs to link to the Chinese version of `Newsletter #38`, which has not been translated yet, so the link cannot be established. Furthermore, `Newsletter #38` itself needs `Newsletter #37` translated to link correctly.

I will merge this issue and the Bech32 content into this branch after the translations of other issues are completed and the PRs are approved, to ensure the translation links work properly.

Thank you!

bitschmidty (approved):
ACK, looks good, thank you @bittomhan

PR #1682, opened 2024-05-20
fix format of title links in zh

bittomhan:
**Purpose:** Update Link Format in Simplified Chinese Version to Seamlessly Align with the English Version

**Explanation:** Through `make preview` checks and trials, I have updated the formatting issues in the Simplified Chinese translation version.

In the original English paragraph such as `- **Opportunity available for providing utility functions outside of Bitcoin Core:**`, the link format generated is:
`/en/newsletters/2018/11/13/#opportunity-available-for-providing-utility-functions-outside-of-bitcoin-core`

However, in the Simplified Chinese translation, if the format is `- **<!--opportunity-available-for-providing-utility-functions-outside-of-bitcoin-core-->提供 Bitcoin Core 之外的实用功能的机会：**`, the generated link format would be:
`/zh/newsletters/2018/11/13/#opportunity-available-for-providing-utility-functions-outside-of-bitcoin-core-bitcoin-core`

As you can see, compared to the English link, the Chinese link, despite commenting the same content, adds the English and numeric content from the beginning of the Chinese paragraph to the end of the link during page generation. The reason is that the content between `- **` and `**` is treated as a whole when generating the link. This difference in link generation leads to a poor experience for readers who wish to read the original text by changing zh to en in the link, as they cannot locate the same paragraph. To improve this, we need to modify the existing format to generate the corresponding link.

In this update, I have changed the format of the translated content by adding `****` between `-->` and `the Chinese title` to form a separation, which alters the title generation logic in the original code. After checking with `make preview`, the generated link format is:
`/zh/newsletters/2018/11/13/#opportunity-available-for-providing-utility-functions-outside-of-bitcoin-core`
Thus, the generated Chinese link format can remain consistent with the original English link format.

I have thoroughly checked the link generation method for all the first 21 issues and made corresponding modifications. I will update the correct link format in the manual and revise future links accordingly. The previous link generation method referred to the Simplified Chinese translation method after issue 203, without attempts to modify it for optimal results. This is now corrected.

---

Additional Explanation: If the `**` is not added around the initial link comment, i.e., `<!--xyz-->`, forming the format `**<!--xyz-->**`, it cannot be recognized as a comment to generate a link.

Whether the translated Chinese paragraph's initial content also uses `**` to bold the format, forming `**一二三：**`, should only be consistent with the English content.


bitschmidty:
@hulatown your feedback might be valuable here at a conceptual level

hulatown:
I also noticed the repeat of any English word in the translated title.
I think this trick can solve the problem 👍

bitschmidty:
@harding I wanted to check if this was any concern regarding the migration to Hugo?

PR #1684, opened 2024-05-21
Topics: anchor-outputs fix wrong statement

Lagrang3:
An unconfirmed commitment transaction does not allow funds to be stolen. The timelock in the `to_local` output (see BOLT03) is constructed with an OP_CSV which makes it relative to the height of the tx confirmation. The situation actually leads to funds stuck from the point of view of the party that started the unilateral close. He cannot CPFP this transaction due to the relative timelock.

### DIFF
--- a/_topics/en/anchor-outputs.md
+++ b/_topics/en/anchor-outputs.md
@@ -186,8 +186,11 @@ become unresponsive).  Because the broadcast of the commitment
 transaction may occur a long time after it was created, the commitment
 transaction may pay too much or too little in transaction fees.
 Paying a too-low feerate may prevent the commitment transaction from
-confirming before any timelocks contained within it expire, allowing
-funds to be stolen.
+confirming making funds stuck indefinitely for the closing party---the
+output *to us* in the commitment trasaction has a time lock relative
+to it's confirmation height `to_self_delay OP_CHECKSEQUENCEVERIFY`
+which is part of the protection mechanism against cheating by broadcasting
+an old commitment state.
 
 The solution to this is for the commitment transaction to pay a
 minimal amount of fees and then to allow either channel participant to fee


harding:
> An unconfirmed commitment transaction does not allow funds to be stolen. 

An unconfirmed commitment transaction that has pending HTLCs does allow funds to be stolen if it fails to confirm before some (or all) of those HTLC timelocks expire.

murchandamus:
I know too little about this to be assertive, but I happened to have access to two lightning developers and they agreed that HTLCs are vulnerable to be claimed on a separate channel if the commitment transaction remains unconfirmed. It seems to me that this contradicts the original claim, but perhaps we should still elaborate our description?

harding:
Sure, an update to this PR that notes the additional problem of stuck funds would be welcome.

Lagrang3:
HTLCs do have a absolute time lock, but I didn't mention HTLCs.
I am referring to the outputs that pay out the channel parties.

Just considering a state of channel without HTLCs. The payout to the remote peer is unencumbered, while the payout
to oneself is delayed with a time lock relative to the confirmation height. That's why if I broadcast the transaction (force close)
I cannot CPFP it, because I am unable to spend my own output until the transaction is confirmed plus some delay.

harding:
@Lagrang3 I think we're in agreement that a commitment transaction with no HTLC outputs does not create a theft risk.  However, if a channel has pending HTLCs, they must be settled onchain before a deadline if one of the parties is not following the protocol (e.g. they've become unresponsive).  A prerequisite for settling pending HTLCs onchain is getting the commitment transaction confirmed since it creates the outputs paying those HTLCs.  Therefore, a commitment transaction with HTLC outputs must be possible to confirm onchain before the deadline.

I'm happy to see an update to this topic that adds nuance about timely confirmation not always being necessary, but I think the existing text that focuses on the (sometimes) need for timely confirmation is accurate in the worst case (which is what security-focused developers design for).

PR #1695, opened 2024-05-28
fix YAML for `x-only-public-keys`

kouloumos:
there is a minor issue with the `see_also` key
resulting in broken YAML and "Schnorr signatures"
to not be displayed in the ["See also" section](https://bitcoinops.org/en/topics/x-only-public-keys/).


bitschmidty (approved):
ACK, thank you @kouloumos !

PR #1700, opened 2024-05-30
Newsletter-40: translate into Chinese

bittomhan:


bitschmidty:
@bittomhan I think you can rebase on master that has zh 39 now

bittomhan:
> @bittomhan I think you can rebase on master that has zh 39 now

I have completed the modifications. If I only modify the links at the end of the `2019-03-19-bech32-sending-support.md` file each time, I may need to `rebase to master` each time and wait for your secondary review. 

Do you think such a double review process is too cumbersome? If so, I will not modify the end links in the future and will make a unified modification after all registrations are complete. Or should I continue to make thorough modifications for each issue to ensure the work is as accurate as possible, even though this requires an additional review?

bitschmidty (approved):
ACK, thank you @bittomhan

PR #1702, opened 2024-05-30
Newsletter-46: translate into Chinese

bittomhan:
Hello,

I noticed that there are references to "bech32"-related articles here. However, the articles ultimately point to the "specials/bech32/" path, where there are only English articles. Should I translate them? Or how should I handle this?

I found that if they are not translated, there will be some misalignment in the footnotes: the Chinese footnotes and the English footnotes cannot be unified in a single numerical sequence according to the format.

Screenshot 2024-05-30 at 10 25 08 AM


bittomhan:
> Hello,
> 
> I noticed that there are references to "bech32"-related articles here. However, the articles ultimately point to the "specials/bech32/" path, where there are only English articles. Should I translate them? Or how should I handle this?
> 
> I found that if they are not translated, there will be some misalignment in the footnotes: the Chinese footnotes and the English footnotes cannot be unified in a single numerical sequence according to the format.
> 
> Screenshot 2024-05-30 at 10 25 08 AM

This problem still exist.
Screenshot 2024-06-23 at 7 00 31 AM

bitschmidty (approved):
ACK

PR #1710, opened 2024-05-31
Newsletter-305:Translate into Japanese

azuchi:


bitschmidty (approved):
ACk, thank you @azuchi !

PR #1750, opened 2024-06-25
Newsletter-307: Apply #1739 to translations(ja, fr)

azuchi:


bitschmidty (approved):
Thank you @azuchi ! utACK

PR #1781, opened 2024-07-12
Newsletter-311:Translate into Japanese

azuchi:


bitschmidty (commented):
ACK, thank you @azuchi !

Issue #1773, opened 2024-07-05
Support autolinking to markdown BIPs


bitschmidty:
The `[BIP379][]` style linking doesnt work correctly with BIPs that use .md instead of .mediawiki extensions

harding:
Fixed in #1804

PR #1804, opened 2024-07-30
Includes: manually add BIP reference links

harding:
Some merged BIPS now use non-`.mediawiki` extensions, so our code for automatically generating BIP links is broken.  Also, as BIP assignments become more sparse, that code is slowing down the side build for both Jekyll and Hugo.

Replace automatic code with a manual list.  This should not be too much of a problem for authors and it allows any BIP extension (or URL change) to be accommodated.

Our existing tests catch any case where we fail to provide a link definition, so this should be safe to merge as soon as it passes tests.


bitschmidty:
Concept ACK. This will address https://github.com/bitcoinops/bitcoinops.github.io/issues/1773

bitschmidty (approved):
ACK

murchandamus:
Postmerge ACK

PR #1809, opened 2024-08-02
Refactor: replace deprecated global link defs with local link defs

harding:
Part of #1551 

Back in 2020, when we first added the topic index, we added a Jekyll function to deprecate some link definitions that we had been using.  Rather than port that function to Hugo and carry over that technical debt, I've integrated each old link into the source files that called it.  This produces identical HTML output to HEAD, which can be verified using something like:

1. Checkout this branch
2. Build the site: `make clean all` (all tests should pass)
3. Backup the HTML output: `cp -a _site _new_site`
4. Checkout the commit before this PR: `git reset --hard HEAD^`
5. Build the site again: `make clean all`
6. Compare the HTML output: `diff -ruN _site _new_site | colordiff | less -R`  (should be identical except for the dynamic date in the RSS file)

### DIFF
--- a/_includes/references.md
+++ b/_includes/references.md
@@ -52,24 +52,6 @@
 [rust bitcoin repo]: https://github.com/rust-bitcoin/rust-bitcoin
 [rust-lightning repo]: https://github.com/rust-bitcoin/rust-lightning
 
-{% comment %}<!-- deprecated links; don't use these any more -->{% endcomment %}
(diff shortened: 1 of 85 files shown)


bitschmidty (approved):
tACK, thank you @harding !

- scanned commit diff
- diff'd before/after site output and confirmed just feed.xml changed

Feel free to merge when you see fit @harding

PR #1810, opened 2024-08-03
Fix links for hugo

harding:
Part of #1551 

This cleans up a bunch of links to make them work correctly with Hugo.  Although code review on this PR might be kinda annoying, it's easier to review as an HTML diff as a lot of these changes have no effect on the HTML output:

1. Checkout this branch
2. Build the tip: `make clean all` (all tests should pass)
3. Backup the generated site: `cp -a _site _new_site`
4. Checkout the merge root, something like `git reset --hard HEAD^10`
5. Build the site again: `make clean build`
6. Diff the resultant HTML: `diff -ruN _site _new_site | colordiff | less -R`

I've separate each type of change into its own commit, with the commit message describing the motivation for the change.


harding:
Converted to draft.  Found an edge case with the Hugo include code that breaks footnotes; fixing that may allow dropping some of these commits.

harding:
> Converted to draft. Found an edge case with the Hugo include code that breaks footnotes; fixing that may allow dropping some of these commits.

Converting back to Ready For Review.  Fixed the edge case and updated this PR to drop about half the commits, making it a much smaller PR that should be easier to review.  The same review suggestions from the OP can be used, but it's now also probably practical to simply review the commit diffs alone.

bitschmidty (approved):
tACK

- reviewed each commit (thanks for the descriptions @harding )
- built and diff'd locally
- reviewed diff logs

lgtm, thanks for all your effort on this transition again, @harding 

Ok to merge when you see fit from your perspective.

PR #1813, opened 2024-08-06
Podcast: add 314 recap

bitschmidty:
also fixes up a new314 news title

### DIFF
--- a/_posts/en/newsletters/2024-08-02-newsletter.md
+++ b/_posts/en/newsletters/2024-08-02-newsletter.md
@@ -16,7 +16,7 @@ infrastructure software.
 
 ## News
 
-- **Disclosure of vulnerabilities affecting Bitcoin Core versions before 0.21.0:**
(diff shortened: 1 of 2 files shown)


_posts/en/newsletters/2024-08-02-newsletter.md
```diff
@@ -16,7 +16,7 @@ infrastructure software.
 
 ## News
 
-- **Disclosure of vulnerabilities affecting Bitcoin Core versions before 0.21.0:**
+- **Disclosure of vulnerabilities affecting Bitcoin Core versions before 22.0:**
   Niklas Gögge [posted][goegge disclosure] to the Bitcoin-Dev mailing
```
harding, line 20: Whoops.  I copied and pasted from a previous newsletter so that we would use a consistent format for these disclosures, but it appears I forgot to update the title.  ACK on the change and thanks for catching!

PR #1824, opened 2024-08-15
Newsletter-314: translate into Chinese

hulatown:


bitschmidty (commented):
ACK, thank you @hulatown

PR #1830, opened 2024-08-16
Remove never-released scaling book content

harding:
Part of #1551 

We have a few pages at [/en/scaling](https://bitcoinops.org/en/scaling/) that we intended to release as a longer version of [Techniques to reduce tranaction fees](https://en.bitcoin.it/wiki/Techniques_to_reduce_transaction_fees), but we never finished the work and @bitschmidty kindly converted the useful bits to blog posts almost 4 years ago (e.g. #461).

I think we should drop this content now.

I haven't provided any link directs or other compatibility.  AFAIK, this material was never official published and its full of missing content and TODOs.  We can always `git revert` if someone appears with complaints.

This is motivated by the Hugo upgrade, as I don't think it's useful to port this material to the new site.


bitschmidty (commented):
ACK

Searched for any links to /en/scaling*

PR #1832, opened 2024-08-16
Localize a few pages: about, workshops, internal/sources

harding:
Part of #1551 

Moves the following pages:

- /about -> /en/about
- /workshops -> /en/workshops
- /internal/sources -> /en/internal/sources

All internal links to those pages are updated (as verified by the first commit passing tests) and redirects are added from the old URLs to the new pages for anyone linking externally (second commit).

This move is good on its own, but it's motivated by my Hugo WIP using multilingual mode which wants all pages in a language hierarchy (e.g. /en/).

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -20,10 +20,10 @@
         </label>
 
         <div class="trigger">
-            <a class="page-link" href="/about/">About</a>
+            <a class="page-link" href="/en/about/">About</a>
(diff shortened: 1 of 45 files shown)


bitschmidty (approved):
ACK

Rebased on master and resolved the index.md conflict from the scaling book removal changeset

Issue #306, opened 2019-12-23
Travis build error with bundler


bitschmidty:
Example of failed build:
https://travis-ci.org/bitcoinops/bitcoinops.github.io/builds/628560320?utm_source=github_status&utm_medium=notification

Affecting 

https://github.com/bitcoinops/bitcoinops.github.io/pull/305
and
https://github.com/bitcoinops/bitcoinops.github.io/pull/282

Appears to have started after https://github.com/bitcoinops/bitcoinops.github.io/pull/304 was merged.

harding:
Weird.  #304 said it passed its Travis check.

PR #1843, opened 2024-08-23
Newsletter-315: Fix references to past newsletters

jirijakes:
The original English text contained swapped references to #306 and them.


bitschmidty (approved):
Thanks for correcting this @jirijakes !

PR #1860, opened 2024-08-31
Fix duration of second epoch in time warp description

### LINKED ISSUE #1859: Time Warp attack description
Hi.
In the topic about the Time Warp Attack it reads: 
```
... For the second epoch, miners can make it appear that the first block was mined two weeks
in the past and the last block was mined at present, giving an average of about 20 minutes per block ...
```
If the time stamp difference between the first and the last block is 2 weeks that gives an approximate
time of 10 minutes per block since `2016 = 2 weeks * 7 days * 144 daily blocks`.
If the attackers wants to cut in half the difficulty they should make it look like if 4 weeks have passed between
the first and the last block.

A note aside: I think by this reasoning the difficulty could be reduced at most by a factor of 4 on each difficulty period
because of the limit on the adjustment step defined in `pow.cpp` `CalculateNextWorkRequired`.

murchandamus:
Fixes #1859 

Also amends the maximum block production rate to 6 blocks per second.

### DIFF
--- a/_topics/en/time-warp.md
+++ b/_topics/en/time-warp.md
@@ -93,7 +93,7 @@ Bitcoins DAA only adjusts difficulty once every 2016 blocks.  Miners who
 begin a time warp attack cannot significantly affect difficulty for the
 first 2016-block epoch because both the first and last blocks in that
 epoch will need to have approximately accurate times.  For the second
-epoch, miners can make it appear that the first block was mined two
+epoch, miners can make it appear that the first block was mined four
 weeks in the past and the last block was mined at present, giving an
 average of about 20 minutes per block and causing the DAA to halve the
 difficulty.  Miners will then by able to complete the next epoch twice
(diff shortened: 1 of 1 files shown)


_topics/en/time-warp.md
```diff
@@ -93,13 +93,13 @@ Bitcoins DAA only adjusts difficulty once every 2016 blocks.  Miners who
...
 epoch will need to have approximately accurate times.  For the second
-epoch, miners can make it appear that the first block was mined two
+epoch, miners can make it appear that the first block was mined four
 weeks in the past and the last block was mined at present, giving an
 average of about 20 minutes per block and causing the DAA to halve the
 difficulty.  Miners will then by able to complete the next epoch twice
 as fast and yet make it seem like it took about 25 minutes per block,
 lowering difficulty further.  They can repeat the attack indefinitely
-until they're producing one block per second, the lower bound allowed by
+until they're producing [six blocks per second]({{bse}}123698), the lower bound allowed by
```
bitschmidty, line 102: I dont think bse is is a global reference. Should probably just reference https://bitcoin.stackexchange.com/questions/123698/how-many-blocks-per-second-can-sustainably-be-created-using-a-time-warp-attack in the foot of the doc like a normal link

murchandamus: Thanks, I amended the link.

bitschmidty (approved):
ACK, thanks @murchandamus

PR #1873, opened 2024-09-06
block-withholding: remove excerpt double negation.

epysqyli:
I believe there's a typo in the first paragraph of the [block-withholding](https://bitcoinops.org/en/topics/block-withholding) entry, i.e. it seems to me that the paragraph should actually be like below:

> ... where a miner submits and receives payment for shares that are not eligible to become full blocks but doesn’t submit shares that are eligible to become full blocks. ...

Is this the case?


bitschmidty (approved):
ACK, thank you @epysqyli for catching this and submitting a correction

PR #1899, opened 2024-09-17
Newsletter-70: translate into Chinese

bittomhan:

PR #1902, opened 2024-09-19
Blog: taproot workshop url fixup

bitschmidty:
Historical URL fixup that is easy enough. Seen via https://github.com/bitcoinops/bitcoinops.github.io/pull/1899#pullrequestreview-2315999484

cc: @harding in case these declarations will be problematic for the hugo conversion


harding:
Tests are failing, but concept ACK

PR #1927, opened 2024-09-29
Newsletter-81: translate into Chinese

bittomhan:


bittomhan:
@bitschmidty I noticed an error and this is really confusing to me since it has no clue on what happened to the `wasabi` path. Should I translate this file or just keep it in `en``? What should I do?

Screenshot 2024-09-29 at 8 37 47 AM

bitschmidty:
@bittomhan I dont think you should translate the individual matrix pages at this time. You can leave as /en/ versions for this PR. 

@TorokNRoll is working on a new version of the compatibility pages that will be coming soon and is also working on any URL redirects

bitschmidty (approved):
ACK

PR #1936, opened 2024-10-07
Newsletter-83: translate into Chinese

bittomhan:


bitschmidty (approved):
Note: news 82 was also included here, which was just merged.

ACK

PR #1946, opened 2024-10-11
Bitcoin Core 28.0 Wallet Integration Guide :Translate into Japanese

azuchi:


bitschmidty (approved):
ACK, thank you @azuchi !

PR #1947, opened 2024-10-11
Newsletter-324:Translate into Japanese

azuchi:
After PR #1946 is merged the tests will pass.


bitschmidty (approved):
ACK, thank you @azuchi !

PR #1961, opened 2024-10-15
Fix typo in wallet integration guide

murchandamus:
:s/feereate/feerate/


bitschmidty (approved):
ACK, thanks @murchandamus

PR #1997, opened 2024-10-30
Podcast: add SuperScalar Deep Dive

bitschmidty:
Two commits here

First one is a refactor of the podcast linking includes so I can reuse component in an non-newsletter podcast.

I tried to keep the diff small including spacing, but you can check using the commands: (provided in another PR by @harding)

1. Checkout this branch
2. Build the tip: make clean all (all tests should pass except the fixme)
3. Backup the generated site: cp -a _site _new_site
4. Checkout the merge root, something like git reset --hard HEAD^2
5. Build the site again: make clean build
6. Diff the resultant HTML: diff -ruN _site _new_site | colordiff | less -R

Second adds the SuperScalar podcast. I am using dropbox to host the audio file before publication, so the timestamp seeking might be slow.

Current publication targeting for 10/31.

If there are better ways to jekyll this, open to suggestions. @harding if this ruins hugo compatibility, let me know.

### DIFF
--- a/_includes/functions/podcast-bullet.md
+++ b/_includes/functions/podcast-bullet.md
@@ -1,0 +1,7 @@
+<li id="{{ include.podcast_slug | slice: 1, include.podcast_slug.size }}" class="anchor-list">
+        <p>
+          <a href="{{ include.podcast_slug }}" class="anchor-list-link">●</a>
+          {{ include.title }}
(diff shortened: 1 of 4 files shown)


harding:
Tested ACK.  The code changes look fine to me.  Viewed the preview page for this podcast and was able to play from beginning; seeking didn't work for me but @bitschmidty notes that's probably an issue with the temporary staging on dropbox.  Viewed the preview page for several previous podcasts and everything LGTM.

Thanks @bitschmidty!

PR #2001, opened 2024-11-03
Newsletter-111: translate into Chinese

bittomhan:


bitschmidty (changes requested):
This is now a conflicting PR because of the changes required in the last newsletter recently merged. I suspect all of the other open PRs are conflicting for the same reason.

@bittomhan perhaps you can just include the files for each newsletter's PR and not the other newsletter files? Although there would be broken links for the newsletter that have not been merged yet.

Or perhaps we just have one open zh translation PR at a time? Or someone else has a suggestion.

bittomhan:
> This is now a conflicting PR because of the changes required in the last newsletter recently merged. I suspect all of the other open PRs are conflicting for the same reason.
> 
> @bittomhan perhaps you can just include the files for each newsletter's PR and not the other newsletter files? Although there would be broken links for the newsletter that have not been merged yet.
> 
> Or perhaps we just have one open zh translation PR at a time? Or someone else has a suggestion.

I double checked this is one PR issue which affact all other PR.

I have already checked and fixed it. I will keep the zh translation PR in 5 newsletters.

Thank you for your help!

bitschmidty:
@bittomhan I used this approach to squash all the commits from this branch into one and also dropped the commit to change 110: https://stackoverflow.com/a/14885097

bittomhan:
> @bittomhan I used this approach to squash all the commits from this branch into one and also dropped the commit to change 110: https://stackoverflow.com/a/14885097

Thank you and I used the command on local pc. It can be used to check the diff and I have already see that even I commit it there will be no change on the PR. So I know that it is OK.

Thank you!

bitschmidty:
> > @bittomhan I used this approach to squash all the commits from this branch into one and also dropped the commit to change 110: https://stackoverflow.com/a/14885097
> 
> Thank you and I used the command on local pc. It can be used to check the diff and I have already see that even I commit it there will be no change on the PR. So I know that it is OK.
> 
> Thank you!

Yes, but the idea is to have only commits related to the newsletter in the PR for: 1. simplicity and 2. not having to rebase if an earlier newsletter changes

bitschmidty (approved):
ACK

PR #2020, opened 2024-11-07
Newsletter 327 translate in French

Copinmalin:


bitschmidty (approved):
ACK, thank you @Copinmalin 

(unsure why there is a build happening for this PR at the moment, but Ill let it finish)

PR #2023, opened 2024-11-08
calc-size: Clarify conservative signature size

murchandamus:
Adds an explanation for the signature size and mentions that low-r grinding wallets will produce smaller signatures consistently.


bitschmidty (approved):
ACK, thank you @murchandamus

PR #2029, opened 2024-11-15
Newsletter-329:Translate into Japanese

azuchi:


bitschmidty (approved):
ACK, thank you @azuchi and apologies for my delay in reviewing

PR #2113, opened 2025-01-03
Update topics to reflect taproot activation

vostrnad:
This hopefully removes all remaining mentions of taproot, schnorr signatures and MAST that describe them as _proposed_ additions to Bitcoin.

Better late than never.


bitschmidty (approved):
ACK, thanks for doing this @vostrnad

PR #2161, opened 2025-02-06
tools: Add reorg calculator

### LINKED ISSUE #2156: Feature suggestion: reorg calculator
Greg Maxwell used to host an online calculator based on Nakamoto's code from the original Bitcoin paper that let you plug in what percentage of hashrate an attacker controlled and how many blocks they wanted to reorg.  It then returned the probability of attacker success.

I think it would be useful for us to host something like that, as it's a calculation I occasionally need to perform.

Note: IIRC, there's a more accurate version of Nakamoto's math, which we should probably use.  See https://bitcoinmagazine.com/technical/how-satoshi-messed-his-math-and-how-these-academics-just-fixed-it

jrakibi:
This PR adds a new tool that shows the probability that an attacker with a given percentage of the total network hashrate could successfully execute a block reorganization of a given depth. 

The current implementation matches the mathematical model presented in the Bitcoin whitepaper (Chapter 11) and is similar to Greg Maxwell's [online calculator](https://web.archive.org/web/20190201183912/https://people.xiph.org/~greg/attack_success.html) 

The tool also:
- Uses pure JS (no-dependencies)
- Follows similar approach/styling to the existing Transaction Size Calculator

Note: the PR is marked as draft because it currently implements the original whitepaper formula. 
This will be updated to match the calcultion described in [Double Spend Races](https://diyhpl.us/~bryan/papers2/bitcoin/Double%20spend%20races%20-%202017.pdf) paper that takes into account the `time deviation factor`

Closes #2156

### DIFF
--- a/en/tools/reorg-calculator.md
+++ b/en/tools/reorg-calculator.md
@@ -1,0 +1,144 @@
+---
+permalink: /en/tools/reorg-calculator/
+title: "Reorg Calculator"
+layout: page
+breadcrumbs: false
+---
+
+
+<form action="" id="hashcalc_kappa" onsubmit="return false;">
+  <fieldset>
+    Attacker's hash power (%): <input type="number" id="hashpower" min="0" max="100" value="24" onchange="updateProbability()"/><br>
+    Blocks to catch up: <input type="number" id="blocks" min="0" value="6" onchange="updateProbability()"/><br>
+    Time ratio (κ): <input type="number" step="0.0001" id="kappa" min="0.0" value="1.0" onchange="updateProbability()"/>
+
+    <br><br><hr>
+    <div id="result"></div>
+  </fieldset>
+</form>
+
+<script>
+/**
+ * Q(z, x) = Upper Regularized Incomplete Gamma (integer z>0):
+ *           Q(z,x) = e^(-x) * sum_{n=0}^{z-1} [x^n / n!].
+ */
+function upperGammaQ(z, x) {
+  if (z <= 0) return 1.0;
+  if (x < 0)  return 1.0;
+
+  let sum = 0.0;
+  let term = 1.0; // x^0 / 0!
+  for (let n = 0; n < z; n++) {
+    if (n > 0) term *= x / n;
+    sum += term;
+  }
+  return Math.exp(-x) * sum;
+}
+
+/**
+ * attackerSuccessProbabilityKappa(q, z, kappa)
+ *
+ * Conditional double-spend probability from "Double Spend Races".
+ *
+ * P(z, kappa) = 1
+ *   - Q(z, kappa*z*(q/p))
+ *   + (q/p)^z * exp[kappa*z*((p-q)/p)] * Q(z, kappa*z)
+ */
+function attackerSuccessProbabilityKappa(q, z, kappa) {
+  if (q >= 0.5) {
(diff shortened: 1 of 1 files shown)


kanzure:
needs tests. I don't know how bitcoinops prefers to do javascript packaging for the website. A separate javascript repo for this functionality might make sense?

jrakibi:
> needs tests. I don't know how bitcoinops prefers to do javascript packaging for the website. A separate javascript repo for this functionality might make sense?

Thanks for the review

- Added tests
- For JavaScript packaging, I followed the same approach used for the Transaction Size Calculator (as requested by @bitschmidty here https://github.com/bitcoinops/bitcoinops.github.io/issues/2156#issuecomment-2640144561), where javascript  is included inline in the Markdown file

Question: Should we keep the calculation as it is in the whitepaper and Greg Maxwell’s tool, or should we update it to align with the new calculation in the [Double Spend Races](https://diyhpl.us/~bryan/papers2/bitcoin/Double%20spend%20races%20-%202017.pdf) paper, which accounts for the `time deviation factor`?

bitschmidty:
> Question: Should we keep the calculation as it is in the whitepaper and Greg Maxwell’s tool, or should we update it to align with the new calculation in the [Double Spend Races](https://diyhpl.us/~bryan/papers2/bitcoin/Double%20spend%20races%20-%202017.pdf) paper, which accounts for the `time deviation factor`?

I like @harding's idea:

> Note: IIRC, there's a more accurate version of Nakamoto's math, which we should probably use. See https://bitcoinmagazine.com/technical/how-satoshi-messed-his-math-and-how-these-academics-just-fixed-it

jrakibi:
updated the PR to account for the parameter **κ** (Time Ratio) following the *Double Spend Races* paper.

Previously, the reorg calculator used **Satoshi’s approximate formula**:

$$
P_{\text{Satoshi}}(z) =
1 - \sum_{k=0}^{z-1}
\mathrm{Poisson}(\lambda,k) \bigl[ 1 - (q/p)^{z-k} \bigr],
\quad \lambda = \frac{z q}{p}
$$

which does **not** account for whether the honest \(z\) blocks arrived faster or slower than usual.

Now we implement the **conditional formula**:

$$
P(z,\kappa) =
1 - Q\bigl(z, \kappa z \tfrac{q}{p} \bigr) + \Bigl(\tfrac{q}{p}\Bigr)^{z}
\exp\Bigl[ \kappa z \tfrac{p-q}{p} \Bigr]
Q\bigl(z, \kappa z \bigr)
$$

where **κ** measures the **time ratio**: means how long the honest \(z\) blocks actually took. This parameter lets us see a higher or lower attacker success probability depending on whether the honest chain was “lucky” ($\kappa < 1$) or “unlucky” ($\kappa > 1$).

---

### Validation Against Paper’s Data
Tested the implementation using values from the tables on pages 30 and 31 of *Double Spend Races*. The results from our tool matched the expected values.

image

bitschmidty (approved):
tACK, This is great, thanks for creating this @jrakibi !

I tested the functions on desktop (Chrome and Safari) and mobile and reviewed content.

I also ran the tests ("runTests()") and they passed.

I checked a few different sample values from both the z=3 and z=6 tables from the research and the values match.

PR #2169, opened 2025-02-08
Update html-proofer gem for better performance

azuchi:
The following command to check for broken links:

```
bundle exec htmlproofer --check-html --disable-external --url-ignore '/^\/bin/.*/' ./_site
```

Currently (version 3.15.0) takes 271.221 seconds in my local environment. 
If upgrade to 3.19.4, it will take 128.295 seconds, which is about half the time.

One of the factors that contributed to the performance improvement is https://github.com/gjtorikian/html-proofer/pull/590, and the rest is likely due to external dependent libraries.

Upgrading html-proofer to 4.0 or later will break the current build, so I'm sticking to the 3.0 series.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :development do
 end
 
 group :testing do
-  gem 'html-proofer'
+  gem 'html-proofer', "~> 3.0"
   gem 'mdl'
   gem 'json-schema'
   gem 'toml'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -13,11 +13,11 @@ GEM
     eventmachine (1.2.7)
     ffi (1.11.3)
     forwardable-extended (2.6.0)
-    html-proofer (3.15.0)
+    html-proofer (3.19.4)
       addressable (~> 2.3)
       mercenary (~> 0.3)
-      nokogumbo (~> 2.0)
-      parallel (~> 1.3)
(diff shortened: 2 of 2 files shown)


jirijakes:
It works! Although I measured only 40 second speedup from 190 s to 150 s, it is still a great improvement.

bitschmidty:
Ive had issues running the proofer since updating to the latest macOS versions (and hardware) in the last ~year. Ive just ignore those issues, partially since we will eventually move to Hugo and I can still do build previews which is most of my workflow. And also since troubleshooting was unfruitful.

So I cannot directly test this performance improvement change.

Note: Installing the latest dependencies for this PR on macOS did require a workaround so Ill leave this breadcrumb for others with the issue: https://danielabaron.me/blog/fix-gem-install-errors-m3-mac-ruby-278/#nokogiri

Given the builds can be sensitive, Ill hold off on merging this until I get feedback from @harding that this doesn't interrupt (and in fact, should help!) his workflow.

harding:
Sounds and looks great to me!  I'll try to test it today if I get a chance, but if I haven't replied within 36 hours, it has my ACK.

bitschmidty (commented):
utACK, per my comment

With @jirijakes and @harding acks Im inclined to merge

azuchi:
With the introduction of #2577, this PR is no longer necessary and has been closed.

PR #2191, opened 2025-02-22
Newsletter-342: Translate into German

Walpurga03:


murchandamus (approved):
Looks good to me. I left a couple nits, and a couple term suggestions.

I’m fine with these being addressed either in this pull request or in a follow-up. I will have limited availability in the next few days, so please do not wait for another review from me for this merge.

bitschmidty (approved):
ACK, thank you @Walpurga03 !

PR #2213, opened 2025-03-05
fix threshold sig typo: any 2 of 3 participants

glozow:


bitschmidty (approved):
ACK, thanks @glozow !

PR #2217, opened 2025-03-09
translate newsletter 344 german

Walpurga03:


bitschmidty (approved):
ACK, thank you @Walpurga03 !

Checked links and skimmed a google translation.

Can address any additional feedback in a follow up PR if there is any

PR #1540, opened 2024-03-05
New dockerfile and docker-compose files

arshbot:
Had a ton of issues running this repo locally (macOS 14, apple silicon). Eventually abandoned all effort to run things locally, found the dockerfile.md but still had a bunch of issues related to locales. Whipped up my own dockerfiles, pushing upstream. If the contribution is welcome, I can modify the dockerfile.md file to support this

PR #2222, opened 2025-03-13
docker: speed up building, improve documentation

stickies-v:
A couple of improvements to avoid unnecessary operations and speed up building the container when previewing the website with Docker Compose. Rationale for each change is documented in the commit message.

cc @arshbot - thank you for your earlier work on adding docker support in #1540 !

Note: you'll have to rebuild your docker image for this PR:

```sh
docker compose down && docker compose up --build
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,14 +26,25 @@ site an unlimited number of times with no extra work.
 
 **Docker**
 
-Ensure docker and docker-compose are installed. Run the command in the project
-directory.
+Using Docker eliminates the need for local Ruby installation and
+simplifies the setup process. Ensure Docker and Docker Compose are
+installed, then run the following command in the project directory:
 
-    docker-compose up --build
+    docker compose up
 
-To restart (in the event of code change).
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -7,5 +7,4 @@ services:
       - "4000:4000"
     volumes:
       - .:/usr/src/app
-    working_dir: /usr/src/app
```
bitschmidty, line 10: I wasnt able to get your branch to work. After playing with it more, it seems to, in part, be this line. When I add it back in, make preview seems to work fine and the website renders fine at :4000. When this line is removed, I get an immediate:

`make: *** No rule to make target 'preview'.  Stop.`

As I dont use Docker often, its entirely possible that Im doing something wrong.

bitschmidty:
> I'm guessing you might not have rebuilt your Docker image? ... - I think that should address your issue?

Yes, that was it indeed! I thought if I nuked the containers in the Docker GUI that would have forced a rebuild on future uses, but apparently not. Thanks!

PR #2274, opened 2025-04-18
Newsletter-350:Translate into Japanese

azuchi:
Source page as well.


bitschmidty (approved):
ACK, thank you @azuchi !

PR #1871, opened 2024-09-05
Replace compatibility with new matrix

bitschmidty:

PR #2285, opened 2025-04-23
matrix: Record Nunchuk’s MuSig2 support

murchandamus:
Source: https://bitcoinops.org/en/newsletters/2025/01/24/#nunchuk-adds-taproot-musig2-features

### DIFF
--- a/_data/matrix/Nunchuk.yaml
+++ b/_data/matrix/Nunchuk.yaml
@@ -21,7 +21,7 @@ feature:
   psbt_version: "V0"
   psbt_export: "Yes"
   psbt_update_finalize: "Yes"
-  musig2: "No Support"
+  musig2: "Yes"
   coinjoin: "No Support"
   payjoin_BIP78: "No Support"
   payjoin_alternate: "No Support"


bitschmidty (approved):
Awesome, thanks for updating this @murchandamus !

cc: @TorokNRoll

bitschmidty:
@TorokNRoll after having merged this I realized perhaps the update should have been made in the backing data and not the end yaml code file itself?

murchandamus:
Looking at #1871, it looks like the .yaml files are the data files, I don’t see another backing data type in the pull request.

TorokNRoll:
@murchandamus - as you know, the initial intake to create the .yaml files is a google form.  Thus the "backing data" is in a google sheets.  A script transforms that "backing data" into the .yaml files.

as a background reminder:  this approach was made so that we could more easily have product owners submit their own product data....and also proactively manage it going forward.

@bitschmidty - no sweat, I will update the Nunchuk entry in the google sheets. easy peasy for such situations.

TorokNRoll:
@murchandamus the idea is that the submissions can be maintained by the person who made the initial submission.  Once a submission is made, a unique link can be used by the submitter to update their data as required.

The general idea is:  we will have a record of all the submitters, and their unique inks, then periodically ask them to review and update as needed (either their product has evolved, or the matrix has evolved to capture additional features).

Regarding the current flow for updating the .yaml files:  for the moment, I will ensure that the process is managed as needed.  Once it is "stable", I can give further people in the Optech team access to the google form, google sheet and the g script which creates the .yaml files (of course with a step by step instruction).

make sense?

murchandamus:
That does sounds like a nice workflow for the submitters. It seems potentially a bit error-prone to have two different sources of truth, but let’s try it out.

TorokNRoll:
@murchandamus - speaking of the low barrier for entry for new submitters, a few months back you mentioned that someone from BitGo was willing to participate....should we ask them to submit as the next step of our "ramp up"?

murchandamus:
@TorokNRoll: I reached out to BitGo and will probably have an update by the end of the week.

PR #355, opened 2020-02-24
Newsletters: add 86 (2020-02-26)

harding:
- [x] Bitcoin StackExchange section @bitschmidty 
- [x] `Bitcoin Core #13339` @bitschmidty 
- [x] Stanford Blockchain Conference talk summaries x3 by @jnewbery 
- [x] Maybe: Chaincode Residency announcement by @jnewbery (dependent on the result of a discussion about an event listing policy)
- [ ] Update releases and RCs section for latest info on Tuesday evening @harding

PR #2328, opened 2025-05-23
Newsletter 355 german

Walpurga03:
This PR adds the German translation of Bitcoin Optech Newsletter #355 (May 23, 2025).

The translation maintains all technical terminology while providing a natural reading experience in German. All links have been preserved with appropriate targets to either English or German versions where available.


bitschmidty (approved):
ACK, thank you @Walpurga03 !

PR #2334, opened 2025-05-26
Newsletter-187: translate into Chinese

bittomhan:


bitschmidty (commented):
@bittomhan can you standardize the spacing on all of these open PRs for the notable code segments to have the spacing in there?

PR #2340, opened 2025-05-27
Improve workshop page UX by separating current status from history

Walpurga03:
- Update main workshops page to prominently display current status
- Move historical workshop information (2018-2020) to separate page
- Improve user perception by highlighting that site is actively maintained
- Preserve all historical workshop content in dedicated history page

### DIFF
--- a/en/workshops.md
+++ b/en/workshops.md
@@ -13,76 +13,10 @@ and the specific scaling challenges that they are facing.
 If you have any requests or suggestions for future workshop events, please
 [contact us][optech email].
 
-## Workshops #3, #4 and #5 - Schnorr and Taproot Seminars {#taproot-workshop}
+## Current Status
 
-- San Francisco, September 24 2019
-- New York, September 27 2019
-- London, February 5 2020
+**Currently, no workshops are scheduled.** We are evaluating potential future workshop topics and formats based on community feedback and technical developments in the Bitcoin ecosystem.
 
-*Schnorr signatures* and *Taproot* are proposed changes to the Bitcoin
-protocol that promise greatly improved privacy, fungibility, scalability and
-functionality.
-
-Bitcoin Optech hosted two seminar format workshops which included a mixture of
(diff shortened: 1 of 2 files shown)


bitschmidty (commented):
@Walpurga03 I dont personally like the idea of hiding and javascript showing the items on the page.

That said I think adding this text before the workshop list is helpful:

Currently, no workshops are scheduled. We are evaluating potential future workshop topics and formats based on community feedback and technical developments in the Bitcoin ecosystem.

And then having a Previous Workshops header before listing the old workshops would be helpful.

Also curious what others think.

en/workshops.md
```diff
@@ -13,76 +13,53 @@ and the specific scaling challenges that they are facing.
...
-- Fee estimation, RBF, CPFP best practices
-- Optech community and communication
-
-#### Thanks
-
-Thanks to Square for hosting the workshop and Coinbase for helping with
-organization.
+## Current Status
+
+**Currently, no workshops are scheduled.** We are evaluating potential future workshop topics and formats based on community feedback and technical developments in the Bitcoin ecosystem.
```
murchandamus, line 18: I think the addition of the Current Status saying that we currently don’t plan any workshops is a great improvement. I don’t think the Previous Workshops section need to be hidden, it seems alright to me to just list them under a second section at the same level as Current Status, i.e.:

## Current Status

Not planning anything at this time.

## Previous Workshops

### Workshop 3–5

Bla

### Workshop 2
…


---

Could you explain why you think it would be better to hide the old workshops, @Walpurga03?

Walpurga03: In my view, displaying old workshops that happened long ago gives your website an outdated or neglected impression. A streamlined first impression with less but more current content often looks more professional and inviting.

bitschmidty: Im still share Murch's opinion of listening them under the current status as is without hiding

PR #2357, opened 2025-06-07
feat(pt): add scaffolding for Portuguese and `#357` translation

luisschwab:

PR #2362, opened 2025-06-11
Fix redirects on multiple pages

murchandamus:
I noticed via #2357 that the Chinese translation introduced redirects for top level pages to the Chinese translation that overwrite the English redirects, so I’m proposing to remove them, since the English version is the main site.

While looking these changes over, I also noticed a typo that propagated through several translations and added a redirect for the new matrix page.

### DIFF
--- a/zh/about.md
+++ b/zh/about.md
@@ -2,7 +2,6 @@
 layout: page
 title: About
 permalink: /zh/about/
-redirect_from: /about
 ---
 
 Bitcoin Operations Technology Group（简称 Optech）致力于将最好的开源技术和技术手段引入使用比特币的企业，以降低成本并改善客户体验。
--- a/zh/workshops.md
+++ b/zh/workshops.md
@@ -2,7 +2,6 @@
 layout: page
 title: Workshops
 permalink: /zh/workshops/
-redirect_from: /workshops
 ---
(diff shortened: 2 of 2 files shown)


en/matrix.md
```diff
@@ -4,6 +4,7 @@ title: Bitcoin Feature Matrix
 permalink: /en/matrix/
 nowrap: true
 redirect_from:
+  - /matrix/
```
bitschmidty: Are there links going here currently?

_posts/en/2019-10-29-schnorr-taproot-workshop.md
```diff
@@ -8,6 +8,7 @@ lang: en
 slug: 2019-10-21-schnorr-taproot-workshop
 redirect_from:
   - /en/schorr-taproot-workshop/
```
bitschmidty: @murchandamus I think we actually had a page that was /en/schorr-taproot-workshop/ for a while publicly. So when we renamed it we put this in to redirect from that old url

I believe translations included that url in their versions so included similar redirects when we fixed it.

The above is from my memory. But either way, these revised redirects dont make any sense with this corrected spelling since it would just redirect to itself (although it seems smart enough to not do that on the preview site, perhaps as it would be an infinite redirect).

murchandamus:
I see, I completely misunderstood the situation with the redirects. Fixed it now.

bitschmidty (approved):
ACK, thank you @murchandamus for seeing this

PR #2406, opened 2025-07-04
Newsletter-361:Translate into Japanese

azuchi:
I also fixed the issue where the recap release date was still on Thursday.


bitschmidty (approved):
ACK, thank you @azuchi !

PR #2439, opened 2025-07-31
Newletter-362: translate into Chinese

hulatown:


bitschmidty (commented):
ACK, thank you @hulatown

Issue #365, opened 2020-03-07
Writeup for Eclair 1323


jnewbery:
@harding. For inclusion in Newsletter 88

```
 - [Eclair #1323][] allows nodes to advertise that they'll accept channel
  opens with a value higher than the previous limit of about 0.168 BTC. They
  do this by using the new `option_support_large_channel` feature in the
  `init` message, which was recently [added to BOLT 9](https://bitcoinops.org/en/newsletters/2020/02/26/#bolts-596).
  Supporting channel capacities larger than 0.168 BTC is part of the feature
  set known informally as "wumbo".  See [Newsletter #22][news22 wumbo] for details.
```

jnewbery:
Included in #366

PR #2447, opened 2025-08-05
Fix unintended anchor creation in Newsletter #365

Walpurga03:
Having a link in the first line turns the bullet point into an anchor which is not expected here. Shifting the link to BIP360 to the next line removes the anchor.


murchandamus (commented):
ACK

harding:
LGTM

PR #2743, opened 2026-05-01
Contributor, social updates

bitschmidty:


murchandamus (commented):
Regarding updates of existing people, LGTM.
There are probably some people that we could add meanwhile.

PR #2816, opened 2026-07-15
matrix: Add BIP352 send support for Wasabi

Kruwed:
https://github.com/WalletWasabi/WalletWasabi/pull/13621

### DIFF
--- a/_data/matrix/Wasabi_Wallet_.yaml
+++ b/_data/matrix/Wasabi_Wallet_.yaml
@@ -27,7 +27,7 @@ feature:
   payjoin_alternate: ""
   payjoin_alternate_name: ""
   payment_code: ""
-  silent_payments_BIP352: "No Support"
+  silent_payments_BIP352: "Send Support"
   silent_payments_dns: "No Support"
   lightning_11: "No Support"
   lightning_12: "No Support"
--- a/_data/matrix/Wasabi_Wallet_.yaml
+++ b/_data/matrix/Wasabi_Wallet_.yaml
@@ -26,7 +26,7 @@ feature:
   payjoin_BIP78: "Send Support"
   payjoin_alternate: ""
   payjoin_alternate_name: ""
-  payment_code: ""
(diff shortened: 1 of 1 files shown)


murchandamus (approved):
I checked out the linked PR, and took a look at the preview of the updated matrix. This LGTM.

@Kruwed: Would you happen to know whether Wasabi supports either BIP77 or BIP78 Payjoin? The BIP47 Payment Code line is so far also left empty—I assume that Wasabi does not support Payment codes, but if you know for sure and could backfill those lines that would be great!

Kruwed:
>Would you happen to know whether Wasabi supports either BIP77 or BIP78 Payjoin? The BIP47 Payment Code line is so far also left empty—I assume that Wasabi does not support Payment codes, but if you know for sure and could backfill those lines that would be great!

Wasabi was a pioneer supporter of BIP78 payjoin, but the spec has apparently changed: https://github.com/WalletWasabi/WalletWasabi/issues/13652

BIP77 is not yet on Wasabi's roadmap, but there is a tracking issue: https://github.com/WalletWasabi/WalletWasabi/issues/13628 | (I have a plan to propose an edit to BIP77 that will make it more secure)

BIP47 is a NACK, Wasabi will never integrate it.

_data/matrix/Wasabi_Wallet_.yaml
```diff
@@ -27,7 +27,7 @@ feature:
   payjoin_alternate: ""
   payjoin_alternate_name: ""
   payment_code: ""
```
murchandamus: Oh, the Payjoin support was already tracked. I had missed line 26 which already stated BIP78 Send Support. Outside of this PR, we should probably add a Payjoin_BIP77 line.

Could you please add "No Support" for the `payment_code` line?

```suggestion
  payment_code: "No Support"
```

murchandamus (approved):
Thanks, LGTM.

PR #2822, opened 2026-07-19
fix(matrix): remove trailing backslash from Phoenix topic URL

Bartok9:
## Summary
In `_data/matrix/Phoenix.yaml`, a Bitcoin Optech topic URL ended with a stray backslash, which 404s.

- Broken: https://bitcoinops.org/en/topics/swap-in-potentiam/\
- Fixed: https://bitcoinops.org/en/topics/swap-in-potentiam/

## Why
Keeps the compatibility matrix pointing at a live topic page.

## AI disclosure
Assisted by an AI coding agent (Hermes/Sera). Human-reviewed and URL-checked before opening.

## Testing
- Confirmed URL without trailing backslash resolves; with backslash returned 404.

---

**Agent-Owner:** sera · **Platform:** hermes · **Claim:** sera

### DIFF
--- a/_data/matrix/Phoenix.yaml
+++ b/_data/matrix/Phoenix.yaml
@@ -9,7 +9,7 @@ feature:
   os: "Android, iOS"
   web: "No"
   hwi: "No"
-  default_receive: "P2TR via <a href=\"https://bitcoinops.org/en/topics/swap-in-potentiam/\">swap-in-potentiam</a>"
+  default_receive: "P2TR via <a href=\"https://bitcoinops.org/en/topics/swap-in-potentiam/">swap-in-potentiam</a>"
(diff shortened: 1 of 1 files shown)


Bartok9:
Closing — this change was based on a misread of YAML escaping.

In the double-quoted YAML string, `href=\"…/\"` is the correct way to embed HTML attribute quotes. The backslash before the closing `"` is **not** a trailing backslash on the URL path; Psych/Jekyll already resolve the href to `https://bitcoinops.org/en/topics/swap-in-potentiam/`.

Removing that escape breaks the YAML parser (`Psych::SyntaxError` at line 12), which is why Travis/Netlify failed on this PR. Master is already correct — no code change needed.

Sorry for the noise.

PR #282, opened 2019-11-30
Add Wasabi RBF notification support

nopara73:
PR implements it: https://github.com/zkSNACKs/WalletWasabi/pull/2670

Messages:

image

PR #2836, opened 2026-07-29
docs: fix double-slash FR blog URL in newsletter #282

Bartok9:
FR year-in-review (#282) linked the mempool policy series as `/fr/blog//waiting-for-confirmation/`. Drop the extra slash so it matches the live FR permalink and the EN sibling. Destination returns 200; the double-slash form is a Netlify 301.

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 24h


bitschmidty (approved):
ACK.

@Bartok9 perhaps if there are more of these trailing slash items, you can batch them together?

Bartok9:
Thanks for the merge and the tip, @bitschmidty! Good call — I'll scan the newsletter archive for any other trailing/double-slash URL issues and batch them into a single PR rather than one-offs. Appreciated. 🙏

PR #2823, opened 2026-07-20
News346: correct waitNext() fee_threshold / MAX_MONEY description

Bartok9:
News #346 said waitNext() returns when mempool fees rise "above MAX_MONEY". That is wrong: fee_threshold defaults to MAX_MONEY and that default skips the fee check so the wait is tip-change (or timeout) only. Callers who want fee-triggered updates set a lower threshold.

Closes #2591.

English only; translations of #346 can follow.

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 24h

PR #2857, opened 2026-08-21
Fix waitNext() description in News346

murchandamus:
This supersedes https://github.com/bitcoinops/bitcoinops.github.io/pull/2823. It clarifies that the call only returns once a new template will be returned, fixes up the description regarding the `fee_threshold` parameter, and drops the incorrect bit about template generation being avoided. (Templates have to be generated to compare the fees.) I have decided to not add a description of the `timeout` parameter, as I felt that it was less relevant.


bitschmidty:
@murchandamus do you think this warrants a "correction" (see example from this item: https://bitcoinops.org/en/newsletters/2020/10/07/#bitcoin-core-19723)

bitschmidty (approved):
ACK, Im not strong a correction note so lgtm

PR #2878, opened 2026-09-11
Topics: clarify OP_CSFS introspection requires OP_CAT for stack placement

### LINKED ISSUE #2267: topics: `OP_CHECKSIGFROMSTACK`: CSFS does not enable introspection
The [topic page about `OP_CSFS`](https://github.com/bitcoinops/bitcoinops.github.io/blob/5b09bb7180d6d593af07a221eb16de89cb63d3ee/_topics/en/op_checksigfromstack.md) states:
>  The ability to use `OP_CSFS` to verify an arbitrary message can enable several new features for Bitcoin users:
>  - [...]
>  - **Transaction introspection:** If the same pubkey and signature pair are valid both with `OP_CSFS` and `OP_CHECKSIG`, then the contents of the arbitrary message passed to `OP_CSFS` is identical to the serialized spending transaction (and other data) implicitly used with `OP_CHECKSIG`.  This makes it possible to put a validated copy of the spending transaction on the script evaluation stack where other opcodes can run tests on it in order to enforce restrictions on the spending transaction.  For example, if `OP_CSFS` had been available in 2015 and 2016, it would've been possible to implement the features of [BIP65][] `OP_CHECKLOCKTIMEVERIFY` (CLTV) and [BIP112][] `OP_CHECKSEQUENCEVERIFY` (CSV) using without any consensus changes just by writing a verification script. Looking forward, `OP_CSFS` could also allow scripts to [implement the features][oconnor generic] of the proposed [SIGHASH_ANYPREVOUT][topic sighash_anyprevout] signature hash, as well as other opcode proposals such as [OP_CHECKTEMPLATEVERIFY][topic op_checktemplateverify].  Additionally, `OP_CSFS` would allow the creation of [covenants][topic covenants] that restrict the way in which a set of bitcoins may be spent---for example, a [vault][topic vaults] may restrict its spending transaction to a small set of acceptable scriptPubKeys to limit the risk of theft. [...]

`OP_CSFS` alone does not enable arbitrary introspection, as far as i can tell. I think this text assumes `OP_CAT` but then i think this is misleading to mention transaction introspection as a feature enabled by `OP_CSFS` alone.

Bartok9:
The OP_CSFS topic page described transaction introspection as a feature of `OP_CSFS` alone, without noting that placing the serialized transaction on the stack requires `OP_CAT` or a similar serialization mechanism.

Closes #2267.


murchandamus:
While this clarification is helpful, a full fix would also rework the previous paragraphs which give the wrong impression in the first place.

